PR #6, opened 2018-07-27
header of every file in llvm runtime: defines types

dwightguth:
There isn't much actual code here, but I've laid out how I propose that we represent each type of K term as well as how I propose we allocate them, deallocate them, and trace them for garbage collection. It does also include the representation of the types in terms of llvm types.

I would appreciate thorough feedback prior to moving on because a lot of the code we can write after this point will be dependent on this representation.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "runtime/immer"]
+	path = runtime/immer
+	url = http://github.com/arximboldi/immer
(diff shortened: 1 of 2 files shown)


dwightguth:
I mean, there will be statically allocated instances of %string corresponding to string and identifier constants referenced in the semantics, but not all strings or identifiers will be statically knowable, so what you propose won't work.

PR #7, opened 2018-07-31
add map, list, set hooks in rust

dwightguth:
@bmmoore your review of the rust code would be appreciated. @theo25 feel free to review the function signatures and changes to header.ll as well as the test coverage and any rust you feel you can understand.

### DIFF
--- a/runtime/datastructures/src/decls.rs
+++ b/runtime/datastructures/src/decls.rs
@@ -24,6 +24,8 @@ pub type List = Vector<K>;
 #[link(name="gmp")]
 extern "C" {
   pub fn __gmpz_set_ui(rop: *mut Int, op: usize);
+  pub fn __gmpz_fits_ulong_p(op: *const Int) -> i32;
+  pub fn __gmpz_get_ui(op: *const Int) -> u64;
 }
 
 #[cfg(test)]
--- a/runtime/datastructures/src/decls.rs
+++ b/runtime/datastructures/src/decls.rs
@@ -1,0 +1,69 @@
+extern crate im;
+use decls::im::hashmap::HashMap;
+use decls::im::hashset::HashSet;
+use decls::im::vector::Vector;
+
+pub enum Block {}
+#[allow(non_camel_case_types)]
+pub enum mp_limb_t {}
+
+#[repr(C)]
+#[allow(non_camel_case_types)]
+pub struct Int(
+  i32, // _mp_alloc
+  i32, // _mp_size
+  *const mp_limb_t, // _mp_d
+);
+
+pub type K = *const Block;
+pub type Map = HashMap<K, K>;
+pub type Set = HashSet<K>;
+pub type List = Vector<K>;
+
+#[link(name="gmp")]
+extern "C" {
+  pub fn __gmpz_set_ui(rop: *mut Int, op: usize);
+}
+
+#[cfg(test)]
+pub mod testing {
+  use super::{K,Int};
+  use std::ptr;
+
+  #[link(name="gmp")]
+  extern "C" {
+    pub fn __gmpz_init(rop: *mut Int);
+    pub fn __gmpz_clear(rop: *mut Int);
+    pub fn __gmpz_cmp_ui(op1: *const Int, op2: i64) -> i32;
+  }
+
+  #[repr(C)]
+  pub struct DummyBlock {
+    header: i64,
+  }
+
+  pub const DUMMY0: K = &DummyBlock{header: 0} as *const DummyBlock as K;
+  pub const DUMMY1: K = &DummyBlock{header: 1} as *const DummyBlock as K;
+  pub const DUMMY2: K = &DummyBlock{header: 2} as *const DummyBlock as K;
+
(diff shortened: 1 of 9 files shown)


runtime/datastructures/src/hook_map.rs
```diff
@@ -0,0 +1,394 @@
...
+
+#[no_mangle]
+pub unsafe extern "C" fn hook_MAP_remove(result: *mut Map, m: *const Map, key: K) -> bool {
+  *result = (*m).without(&key);
+  true
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn hook_MAP_difference(result: *mut Map, m1: *const Map, m2: *const Map) -> bool {
+  *result = (*m1).clone().difference_with((*m2).clone(), |v1, v2| if v1 == v2 { None } else { Some(v1) });
```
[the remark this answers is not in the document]
dwightguth: So it removes any keys in m1 which are bound to the same value in m2. That is the same behavior as the ocaml backend, and as you can see below, it's being tested. Is there a particular case you think is incorrect?

runtime/datastructures/src/hook_map.rs
```diff
@@ -0,0 +1,394 @@
...
+  Box::into_raw(b)
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn free_map(ptr: *mut Map) {
+  Box::from_raw(ptr);
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn hook_MAP_element(result: *mut Map, key: K, value: K) -> bool {
```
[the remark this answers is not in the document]
dwightguth: The intent of the signature of the hooks is that the K value of the arguments in the llvm backend are passed to the hook as-is, the result is allocated outside the hook itself, and the hook writes to the previously allocated result.

dwightguth:
@bmmoore this is ready for you to look at again.

runtime/datastructures/src/hook_map.rs
```diff
@@ -0,0 +1,404 @@
...
+  if (*m).is_empty() {
+    return false;
+  }
+  ptr::write(result, *(*m).keys().next().unwrap());
+  true
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn hook_MAP_size(result: *mut Int, m: *const Map) -> bool {
+  __gmpz_set_ui(result, (*m).len());
```
bmmoore: Does this require the int to already be initialized? Many of the other methods work with uninitialized targets, and would leak memory if called with result pointing to an initialized value.

runtime/datastructures/src/hook_map.rs
```diff
@@ -0,0 +1,404 @@
...
+  ptr::write(result, (*m2).clone().union((*m1).clone()));
+  true
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn hook_MAP_removeAll(result: *mut Map, map: *const Map, set: *const Set) -> bool {
+  let mut tmp = (*map).clone();
+  for key in (*set).iter() {
+    tmp.remove(key);
+  }
```
bmmoore: I'm surprised there wasn't already a method, but this seems to be a reasonable implementation (the code for `difference` also involves a loop calling `remove`).

runtime/datastructures/src/hook_list.rs
```diff
@@ -0,0 +1,12 @@
+use super::decls::List;
+
+#[no_mangle]
+pub extern "C" fn alloc_list() -> *mut List {
```
bmmoore: List and Set should probably have the allocation style changed as well.

runtime/datastructures/src/hook_set.rs
```diff
@@ -0,0 +1,12 @@
+use super::decls::Set;
+
+#[no_mangle]
+pub extern "C" fn alloc_set() -> *mut Set {
```
bmmoore: List and Set should probably have the allocation style changed as well.

bmmoore:
`hook_MAP_lookupOrDefault` should do the `match` in the same style as the function before it.

runtime/datastructures/src/hook_map.rs
dwightguth: the assumption is that some initialization will have already occurred when the heap block is created, and I'm including the call to mpz_init in that category.

runtime/datastructures/src/decls.rs
```diff
@@ -0,0 +1,70 @@
...
+pub struct Int(
+  i32, // _mp_alloc
+  i32, // _mp_size
+  *const mp_limb_t, // _mp_d
+);
+
+pub type K = *const Block;
+pub type Map = HashMap<K, K>;
+pub type Set = HashSet<K>;
+pub type List = Vector<K>;
```
bmmoore: Let's see how it goes, but based on the documentation the minimum size of this Vector is a bit high:
https://docs.rs/im/11.0.1/im/vector/index.html
Perhaps 2 or 3 of those arrays will often be shared for small vectors, and we could add further small-vector optimizations if needed.

PR #12, opened 2018-08-07
Code to generate terms from substitutions

dwightguth:
Currently tested: 
* constructing constant terms

Not tested:
* tokens
* functions
* variables
* polymorphism

See the code generated in test.ll. The code is for the generation of the right hand side of the axiom tagged with the `theRule` attribute. I have tested that llc compiles the module and that it looks correct on inspection but have not tested running the generated code yet.


include/kllvm/ast/AST.h
```diff
@@ -102,11 +104,19 @@ class KORESymbol {
...
 };
 
+struct HashSymbol;
+
+class KOREDefinition;
+class KOREObjectSymbolDeclaration;
+
 class KOREObjectSymbol : public KORESymbol {
 private:
   std::string name;
```
pmeredit: Please add comments for member variables. What to firstTag and lastTag do? What is layout?

include/kllvm/ast/AST.h
```diff
@@ -110,13 +135,39 @@ class KOREObjectSymbol : public KORESymbol {
...
 };
 
+struct HashSymbol {
+  size_t operator()(const kllvm::KOREObjectSymbol &s) const noexcept {
+    size_t hash = 0;
+    boost::hash_combine(hash, s.name);
+    boost::hash_combine(hash, s.arguments);
+    return hash;
+  }
+};
```
pmeredit: I'm curious why this is a struct. It doesn't have any members, and is being used as a function object, and it's not static, so you must instantiate a stateless object to use it. Why not just use a namespace?

include/kllvm/ast/AST.h
```diff
@@ -183,6 +236,7 @@ class KOREMetaVariable : public KOREVariable {
 class KOREPattern {
 public:
   virtual void print(std::ostream &Out, unsigned indent = 0) const =0;
```
pmeredit: A comment for markSymbols is in order, too.

include/kllvm/ast/AST.h
```diff
@@ -508,6 +582,10 @@ class KOREDefinition {
 
   using KOREMetaVariableMapType = llvm::StringMap<KOREMetaVariable *>;
 
+  using KOREObjectCompositeSortDeclarationMapType = llvm::StringMap<KOREObjectCompositeSortDeclaration *>;
+
+  using KOREObjectSymbolDeclarationMapType = llvm::StringMap<KOREObjectSymbolDeclaration *>;
+
```
pmeredit: Nice.

include/kllvm/ast/AST.h
```diff
@@ -519,16 +597,25 @@ class KOREDefinition {
   KOREObjectVariableMapType objectVariables;
   KOREMetaVariableMapType metaVariables;
   KOREModuleMapType moduleNames;
+  KOREObjectCompositeSortDeclarationMapType sortDeclarations;
+  KOREObjectSymbolDeclarationMapType symbolDeclarations;
 
   std::vector<KOREModule *> modules;
-  std::vector<KOREPattern *> attributes;
+  llvm::StringMap<KOREObjectCompositePattern *> attributes;
```
pmeredit: I'm interested in what heuristic you use to decide what should have its own type name and what shouldn't. It seems to be related to type length.

lib/ast/AST.cpp
```diff
@@ -1,11 +1,33 @@
...
 void KOREObjectCompositeSort::addArgument(KOREObjectSort *Argument) {
   arguments.push_back(Argument);
 }
 
+bool KOREObjectCompositeSort::operator==(const KOREObjectSort &other) const {
+  if (auto sort = dynamic_cast<const KOREObjectCompositeSort *>(&other)) {
+    if (sort->name != name || sort->arguments.size() != arguments.size()) {
+      return false;
+    }
+    for (int i = 0; i < arguments.size(); i++) {
```
pmeredit: super nitpicky, but prefer ++i to i++.

lib/ast/AST.cpp
```diff
@@ -1,11 +1,33 @@
...
+      if (!(sort->arguments[i] == arguments[i])) return false;
```
pmeredit: Why negating `==` instead of `!=`?

lib/ast/AST.cpp
```diff
@@ -14,6 +36,121 @@ void KOREObjectSymbol::addSort(KOREObjectSort *Sort) {
...
+    if (!(*arguments[i] == *other.arguments[i])) return false;
+  }
+  return true;
+}
+
+std::string KOREObjectSymbol::layoutString() const {
+  std::string result;
+  for (auto arg : arguments) {
+    auto sort = dynamic_cast<KOREObjectCompositeSort *>(arg);
+    if (sort->getName() == "Map") {
```
pmeredit: Not a big deal for a PoC, but we should be using an enum here instead of strings comparisons, it also allows for using a switch.

lib/ast/AST.cpp
```diff
@@ -43,11 +210,48 @@ void KOREDeclaration::addMetaSortVariable(KOREMetaSortVariable *SortVariable) {
...
+static const std::string UNIT = "unit";
+static const std::string FUNCTIONAL = "functional";
+static const std::string CONSTRUCTOR = "constructor";
+
+bool KOREAxiomDeclaration::isRequired() {
+  return !attributes.count(ASSOC) && !attributes.count(COMM) && !attributes.count(IDEM) && !attributes.count(UNIT) && !attributes.count(FUNCTIONAL) && !attributes.count(CONSTRUCTOR);
+}
+
+KOREPattern *KOREAxiomDeclaration::getRightHandSide() const {
+  if (auto top = dynamic_cast<KOREObjectCompositePattern *>(pattern)) {
```
pmeredit: This seems really fragile, but I haven't thought of a better way to handle this, yet. Definitely replacing getName() with unique pointers to each builtin will clean this up some, but the static indices will still be here.

lib/ast/AST.cpp
```diff
@@ -14,6 +36,121 @@ void KOREObjectSymbol::addSort(KOREObjectSort *Sort) {
...
+bool KOREObjectSymbol::isPolymorphic() const {
+  for (auto sort : arguments) {
+    if (auto var = dynamic_cast<KOREObjectCompositeSort *>(sort)) {
+      return false;
+    }
+  }
+  return true;
+}
+
+bool KOREObjectSymbol::isBuiltin() const {
```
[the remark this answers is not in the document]
pmeredit: Additionally there will be a static named instance of each type, e.g.,

`static KoreObjectSymbol *DVSym = ...`  Then it can be referred to simply as `KoreObjectSymbol::DVSym`

lib/ast/AST.cpp
```diff
@@ -75,10 +283,83 @@ void KOREModule::addDeclaration(KOREDeclaration *Declaration) {
...
+  }
+  uint32_t nextSymbol = 0;
+  uint16_t nextLayout = 1;
+  auto instantiations = std::unordered_map<KOREObjectSymbol, uint32_t, HashSymbol>{};
+  auto layouts = std::unordered_map<std::string, uint16_t>{};
+  auto variables = std::unordered_map<std::string, std::pair<uint32_t, uint32_t>>{};
+  for (auto iter = symbols.begin(); iter != symbols.end(); iter++) {
+    auto entry = *iter;
+    uint32_t firstTag = nextSymbol;
+    for (auto iter = entry.second.begin(); iter != entry.second.end(); iter++) {
```
pmeredit: Definitely prefer ++i when using iterators. The compiler should optimize i++ to ++i, but it's still semantically nicer.

lib/codegen/CreateTerm.cpp
```diff
@@ -0,0 +1,190 @@
...
+; %layoutN = type { %blockheader, [0 x i64 *], %map, %mpz *, %block * }
+)LLVM";
+
+std::unique_ptr<llvm::Module> newModule(std::string name, llvm::LLVMContext &Context) {
+  llvm::SMDiagnostic Err;
+  return llvm::parseIR(*llvm::MemoryBuffer::getMemBuffer(LLVM_HEADER), Err, Context);
+}
+
+llvm::Type *getValueType(std::string sort, llvm::Module *Module) {
+  if (sort == "Map") {
```
pmeredit: enums here also.

lib/codegen/CreateTerm.cpp
```diff
@@ -0,0 +1,190 @@
...
+  } else if (sort == "MInt") {
+    assert(false && "not implemented yet: MInt");
+  } else {
+    return llvm::PointerType::get(Module->getTypeByName("block"), 0);
+  }
+}
+
+
+llvm::StructType *getBlockType(llvm::Module *Module, const KOREObjectSymbol *symbol, KOREObjectSymbolDeclaration *symbolDecl) {
+  llvm::StructType *BlockHeaderType = Module->getTypeByName("blockheader");
```
pmeredit: Anywhere where we must absolutely use strings constants like here, they should be declared as globals to avoid mistyping.

include/kllvm/ast/AST.h
theo25: This is typically used when one wants to use such a struct as a template argument, e.g. for an unordered map. The alternative is to use the C++ lambdas which looks kind of awkward.

lib/ast/AST.cpp
theo25: yes I tend to agree. Typically it only matters when ++ is overloaded for some iterator, where the prefix does an in place update and hence is a bit more efficient, but it's nice to be consistent.

lib/ast/AST.cpp
theo25: agreed!

lib/ast/AST.cpp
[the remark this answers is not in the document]
dwightguth: Unfortunately these symbols have parameters and therefore there will not be a single unique instance of them. There is in fact a potentially unbounded number of possible instantiations of these symbols. However, we can make an unordered_set of strings if you'd like?

include/kllvm/ast/AST.h
```diff
@@ -306,7 +369,7 @@ class KOREMetaCharPattern : public KOREMetaPattern {
 // KOREDeclaration
 class KOREDeclaration {
 protected:
-  std::vector<KOREPattern *> attributes;
```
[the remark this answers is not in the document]
dwightguth: This is correct, yes. It's possible the semantics of k pdf wasn't updated yet, but this is based on the discussions we had about attributes in the haskell backend, so it should be correct.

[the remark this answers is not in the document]
theo25: OK in this case maybe a parser update is also in order, but I can do that.

lib/ast/AST.cpp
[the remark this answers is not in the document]
pmeredit: Well, I still see two more options:
1. Create a class just for the head of symbols
2. Use enums instead of strings.

2. Probably works better.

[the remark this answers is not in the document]
pmeredit: Ah, I see the downside of using enums for this now, I think just add the `unordered_set` and call it good.

lib/ast/AST.cpp
```diff
@@ -75,10 +283,83 @@ void KOREModule::addDeclaration(KOREDeclaration *Declaration) {
...
+        symbol->firstTag = symbol->lastTag = instantiations.at(*symbol);
+        symbol->layout = layouts.at(layoutStr);
+      }
+    }
+    uint32_t lastTag = nextSymbol-1;
+    if (!entry.second.empty()) {
+      variables.emplace(entry.first, std::pair<uint32_t, uint32_t>{firstTag, lastTag});
+    }
+  }
+  for (auto iter = symbols.begin(); iter != symbols.end(); iter++) {
```
bmmoore: postincrement here as well, if you are changing to `++iter`

pmeredit:
lgtm, let's just add the set for builtins, everything else looks great.

lib/codegen/CreateTerm.cpp
```diff
@@ -0,0 +1,190 @@
...
+#include "llvm/IR/BasicBlock.h"
+#include "llvm/IR/Constants.h"
+#include "llvm/IR/DerivedTypes.h"
+#include "llvm/IR/Instructions.h"
+#include "llvm/Support/MemoryBuffer.h"
+#include "llvm/Support/SourceMgr.h"
+
+namespace kllvm {
+
+std::string LLVM_HEADER = R"LLVM(
```
bmmoore: This might be the best way to get these declarations into the generated module, but the file also had some value as documentation of our intended memory layout. Perhaps it would be possible to `#include` the text? At least we should have a file of documentation somewhere that covers this information.

lib/codegen/CreateTerm.cpp
```diff
@@ -0,0 +1,190 @@
...
+    const KOREObjectSymbol *symbol = constructor->getConstructor();
+    assert(symbol->isConcrete() && "not supported yet: sort variables");
+    assert(symbol->getName() != "\\dv" && "not supported yet: \\dv");
+    KOREObjectSymbolDeclaration *symbolDecl = definition->getSymbolDeclarations().lookup(symbol->getName());
+    llvm::LLVMContext &Context = block->getContext();
+    if (symbolDecl->getAttributes().count("function")) {
+      assert(false && "not implemented yet: functions");
+    } else if (symbol->getArguments().empty()) {
+      llvm::StructType *BlockType = Module->getTypeByName("block");
+      llvm::IntToPtrInst *Cast = new llvm::IntToPtrInst(llvm::ConstantInt::get(llvm::IntegerType::get(Context, 32), symbol->getTag()), llvm::PointerType::get(BlockType, 0), "", block);
```
bmmoore: I don't see how this is setting the least bit as we agreed on for tagging.

test/test.ll
```diff
@@ -0,0 +1,25 @@
...
+entry:
+  %malloccall = tail call i8* @malloc(i64 ptrtoint ({ %blockheader, [0 x i64], %block* }* getelementptr ({ %blockheader, [0 x i64], %block* }, { %blockheader, [0 x i64], %block* }* null, i32 1) to i64))
+  %0 = bitcast i8* %malloccall to { %blockheader, [0 x i64], %block* }*
+  %1 = getelementptr inbounds { %blockheader, [0 x i64], %block* }, { %blockheader, [0 x i64], %block* }* %0, i64 0, i32 0
+  store %blockheader { i64 131079 }, %blockheader* %1
+  %malloccall1 = tail call i8* @malloc(i64 ptrtoint ({ %blockheader, [0 x i64], %block* }* getelementptr ({ %blockheader, [0 x i64], %block* }, { %blockheader, [0 x i64], %block* }* null, i32 1) to i64))
+  %2 = bitcast i8* %malloccall1 to { %blockheader, [0 x i64], %block* }*
+  %3 = getelementptr inbounds { %blockheader, [0 x i64], %block* }, { %blockheader, [0 x i64], %block* }* %2, i64 0, i32 0
+  store %blockheader { i64 131123 }, %blockheader* %3
+  %4 = inttoptr i32 0 to %block*
```
bmmoore: This tag definitely doesn't have the least bit set.

lib/codegen/CreateTerm.cpp
```diff
@@ -0,0 +1,198 @@
...
+    const KOREObjectSymbol *symbol = constructor->getConstructor();
+    assert(symbol->isConcrete() && "not supported yet: sort variables");
+    assert(symbol->getName() != "\\dv" && "not supported yet: \\dv");
+    KOREObjectSymbolDeclaration *symbolDecl = definition->getSymbolDeclarations().lookup(symbol->getName());
+    llvm::LLVMContext &Context = block->getContext();
+    if (symbolDecl->getAttributes().count("function")) {
+      assert(false && "not implemented yet: functions");
+    } else if (symbol->getArguments().empty()) {
+      llvm::StructType *BlockType = Module->getTypeByName(BLOCK_STRUCT);
+      llvm::IntToPtrInst *Cast = new llvm::IntToPtrInst(llvm::ConstantInt::get(llvm::IntegerType::get(Context, 32), symbol->getTag()), llvm::PointerType::get(BlockType, 0), "", block);
```
bmmoore: This does not ensure the least bit is set. You said the tag would be put in the high 32 bits, and and a 32-bit 1 in the low half.

include/kllvm/ast/AST.h
```diff
@@ -182,6 +182,11 @@ class KOREObjectSymbol : public KORESymbol {
   bool isPolymorphic() const;
   bool isBuiltin() const;
 
+  /* instantiates this symbol (which should be parsed from a pattern in an axiom)
+     with the sorts corresponding to its actual sort parameters after instantiating
+     polymorphic parameters. This happens by replacing the variables in the arguments
+     of the specified declaration with their substitution in the arguments to the pattern
+     that were parsed in braces. */
```
bmmoore: I'm still confused here. Where in KoreSymbol is that information from parsing stored? I guessed `arguments` are the sorts of the expected pattern arguments, but then where is information on the formal and actual sort parameters stored? Maybe this calls for some class-level comments.

lib/ast/AST.cpp
```diff
@@ -150,21 +150,32 @@ bool KOREObjectSymbol::isBuiltin() const {
...
+        instantiated.push_back(arguments[vars.at(*var)]);
       } else {
         instantiated.push_back(sort);
       }
     } else {
       instantiated.push_back(sort);
     }
   }
+  auto returnSort = decl->getSymbol()->sort;
+  if (auto var = dynamic_cast<KOREObjectSortVariable *>(returnSort)) {
```
bmmoore: Doesn't this need to substitute for the sort variables everywhere inside the declared sort?
Consider something like `symbol addMapEntry{K,V}(Map{K,V},K,V):Map{K,V}`. `Map{K,V}` isn't a sort variable, so I think it would be left alone.

test/test.ll
```diff
@@ -7,11 +7,11 @@ entry:
   %malloccall = tail call i8* @malloc(i64 ptrtoint ({ %blockheader, [0 x i64], %block* }* getelementptr ({ %blockheader, [0 x i64], %block* }, { %blockheader, [0 x i64], %block* }* null, i32 1) to i64))
   %0 = bitcast i8* %malloccall to { %blockheader, [0 x i64], %block* }*
   %1 = getelementptr inbounds { %blockheader, [0 x i64], %block* }, { %blockheader, [0 x i64], %block* }* %0, i64 0, i32 0
-  store %blockheader { i64 65538 }, %blockheader* %1
+  store %blockheader { i64 281483566645248 }, %blockheader* %1
```
bmmoore: Why is this showing up in this commit?

include/kllvm/ast/AST.h
```diff
@@ -32,7 +56,11 @@ class KOREObjectSortVariable : public KOREObjectSort {
     return new KOREObjectSortVariable(Name);
   }
 
+  virtual bool isConcrete() const override { return false; }
+  virtual KOREObjectSort *substitute(const std::unordered_map<KOREObjectSortVariable, KOREObjectSort *, HashSort> &subst) override { return subst.at(*this); }
```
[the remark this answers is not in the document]
bmmoore: I'd prefer a bit higher as a hard limit, like 100, but this line is anyway too long being over 150.
80 can be a bit tight - the template instantiation `std::unordered_map<KOREObjectSortVariable, KOREObjectSort *, HashSort>` is already 70 characters by itself.

include/kllvm/ast/AST.h
```diff
@@ -53,21 +81,44 @@ class KOREMetaSortVariable : public KOREMetaSort {
   KOREMetaSortVariable(const std::string &Name) : name(Name) { }
 };
 
+enum class SortCategory {
```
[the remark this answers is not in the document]
dwightguth: I disagree with this. We are already using a scoped identifier, so you will have to prepend `SortCategory::`, which I think is more intuitive than prepending a containing class. It's also shorter, and we don't really have that many uses for SortCategory inside the KOREObjectCompositeSort class, so I don't see that changing the scope makes things much easier.

include/kllvm/ast/AST.h
```diff
@@ -110,13 +187,48 @@ class KOREObjectSymbol : public KORESymbol {
...
+
+  friend HashSymbol;
+
+  friend KOREDefinition;
+
 private:
   KOREObjectSymbol(const std::string &Name) : name(Name), sort(nullptr) { }
 };
 
+struct HashSymbol {
```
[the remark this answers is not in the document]
dwightguth: I would prefer to keep this as is and add an overload to the same class if we start using meta symbols.

include/kllvm/ast/AST.h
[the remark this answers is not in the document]
theo25: OK let's agree to 100 then. I suggested 80 because this is typically the limit but 100 for today's screens should be fine.

include/kllvm/ast/AST.h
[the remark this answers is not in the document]
theo25: OK that sounds good too.

PR #15, opened 2018-08-13
Expand rhs generation to cover many more features

dwightguth:
I have tested each feature I added by manually examining the code it generates in a single case, and I have tested that the generated code compiles, and added coverage of each feature to the ci testing, but I have not manually gone through the entire reference files.

That said, we can now handle 182 of the 183 axioms in imp.kore. The single missing case is injection transitivity which is polymorphic over its sorts and therefore must generate different code depending on the SortCategory of its argument. I will fix this in the next pull request.

### DIFF
--- a/include/kllvm/ast/AST.h
+++ b/include/kllvm/ast/AST.h
@@ -320,6 +320,7 @@ public:
   }
 
   std::string getName() const;
+  KOREObjectSort *getSort() const { return sort; }
 
   virtual void print(std::ostream &Out, unsigned indent = 0) const override;
   virtual void markSymbols(std::map<std::string, std::vector<KOREObjectSymbol *>> &) override {}
@@ -406,6 +407,8 @@ public:
   static KOREMetaStringPattern *Create(const std::string &Contents) {
     return new KOREMetaStringPattern(Contents);
   }
+
+  std::string getContents() { return contents; }
 
   virtual void print(std::ostream &Out, unsigned indent = 0) const override;
   virtual void markSymbols(std::map<std::string, std::vector<KOREObjectSymbol *>> &) override {}
--- a/include/kllvm/ast/AST.h
+++ b/include/kllvm/ast/AST.h
@@ -84,8 +84,10 @@ private:
 };
 
 enum class SortCategory {
-  Map, List, Set, Int, Float, StringBuffer, Bool, MInt, Symbol
+  Uncomputed, Map, List, Set, Int, Float, StringBuffer, Bool, MInt, Symbol
 };
+
+class KOREDefinition;
 
 class KOREObjectCompositeSort : public KOREObjectSort {
 private:
@@ -99,7 +101,7 @@ public:
   }
 
   const std::string getName() const { return name; }
-  const SortCategory getCategory() const { return category; }
+  SortCategory getCategory(KOREDefinition *definition);
 
   virtual bool isConcrete() const override { return true; }
   virtual KOREObjectSort *substitute(const substitution &subst) override;
@@ -109,18 +111,7 @@ public:
   virtual bool operator==(const KOREObjectSort &other) const override;
 
 private:
-  KOREObjectCompositeSort(const std::string &Name) : name(Name) {
-    if (name == "Map") category = SortCategory::Map;
-    else if (name == "List") category = SortCategory::List;
-    else if (name == "Set") category = SortCategory::Set;
-    else if (name == "Array") category = SortCategory::List;
-    else if (name == "Int") category = SortCategory::Int;
-    else if (name == "Float") category = SortCategory::Float;
-    else if (name == "StringBuffer") category = SortCategory::StringBuffer;
-    else if (name == "Bool") category = SortCategory::Bool;
-    else if (name == "MInt") category = SortCategory::MInt;
-    else category = SortCategory::Symbol;
-  }
+  KOREObjectCompositeSort(const std::string &Name) : name(Name), category(SortCategory::Uncomputed) {}
 };
 
 class KOREMetaCompositeSort : public KOREMetaSort {
@@ -146,7 +137,6 @@ public:
 
 struct HashSymbol;
 
-class KOREDefinition;
 class KOREObjectSymbolDeclaration;
 
 class KOREObjectSymbol : public KORESymbol {
@@ -197,8 +187,8 @@ public:
   bool operator==(KOREObjectSymbol other) const;
   bool operator!=(KOREObjectSymbol other) const { return !(*this == other); }
 
-  std::string layoutString() const;
   uint8_t length() const;
+  std::string layoutString(KOREDefinition *) const;
 
   bool isConcrete() const;
   bool isPolymorphic() const;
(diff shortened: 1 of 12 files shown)


include/kllvm/ast/AST.h
```diff
@@ -188,7 +178,7 @@ class KOREObjectSymbol : public KORESymbol {
   const std::vector<KOREObjectSort *> &getArguments() const {
     return arguments;
   }
-  const KOREObjectSort *getSort() const { return sort; }
```
theo25: instead of removing the `const` version, I would suggest to overload with a non-`const` one. So keep
`const KOREObjectSort *getSort() const { return sort; }`
and add the overload
`KOREObjectSort *getSort() const { return sort; }`

dwightguth: Pretty sure you can't have two overloads with the same signature and different return types, but it doesn't matter because KOREObjectSort * is implicitly convertible to const KOREObjectSort *.

lib/codegen/CreateTerm.cpp
```diff
@@ -99,100 +101,335 @@ llvm::Type *getValueType(SortCategory sort, llvm::Module *Module) {
...
+    return Substitution.lookup(variable->getName());
   } else if (auto constructor = dynamic_cast<KOREObjectCompositePattern *>(pattern)) {
     const KOREObjectSymbol *symbol = constructor->getConstructor();
     assert(symbol->isConcrete() && "not supported yet: sort variables");
+    if (symbol->getName() == "\\dv") {
+      auto sort = dynamic_cast<KOREObjectCompositeSort *>(symbol->getArguments()[0]);
+      auto strPattern = dynamic_cast<KOREMetaStringPattern *>(constructor->getArguments()[0]);
+      return createToken(sort->getCategory(Definition), strPattern->getContents());
+    }
     assert(symbol->getName() != "\\dv" && "not supported yet: \\dv");
```
theo25: you should remove this assertion

include/kllvm/ast/AST.h
theo25: Oh yes, sorry I meant to suggest these as the overloaded methods:
`const KOREObjectSort *getSort() const { return sort; }`
`KOREObjectSort *getSort() { return sort; }`

I am positive that you can have overloads like above. This way if one has a `const KOREObjectSort *` or `const KOREObjectSort &`, they cannot get access to a non-const pointer to the member.

pmeredit: Yes, you can overload on the constness of `this`, but not on return type. Since this would overload on this, you can also have a differing return type.

include/kllvm/ast/AST.h
```diff
@@ -84,9 +84,11 @@ class KOREMetaSortVariable : public KOREMetaSort {
 };
 
 enum class SortCategory {
-  Map, List, Set, Int, Float, StringBuffer, Bool, MInt, Symbol
+  Uncomputed, Map, List, Set, Int, Float, StringBuffer, Bool, MInt, Symbol
```
pmeredit: I assume Uncomputed is equivalent to some kind of "sort variable"?

lib/ast/AST.cpp
```diff
@@ -51,6 +51,31 @@ KOREObjectSort *KOREObjectCompositeSort::substitute(const std::unordered_map<KOR
...
+  else if (name == "SET.Set") category = SortCategory::Set;
+  else if (name == "ARRAY.Array") category = SortCategory::List;
+  else if (name == "INT.Int") category = SortCategory::Int;
+  else if (name == "FLOAT.Float") category = SortCategory::Float;
+  else if (name == "BUFFER.StringBuffer") category = SortCategory::StringBuffer;
+  else if (name == "BOOL.Bool") category = SortCategory::Bool;
+  else if (name == "MINT.MInt") category = SortCategory::MInt;
+  else category = SortCategory::Symbol;
+  return category;
+}
```
pmeredit: I like this lazy setting of the category.

include/kllvm/ast/AST.h
dwightguth: It just means we haven't computed the category yet for this term.

dwightguth:
I addressed your comments. @bmmoore do you want to review this?

lib/codegen/CreateTerm.cpp
```diff
@@ -99,100 +101,334 @@ llvm::Type *getValueType(SortCategory sort, llvm::Module *Module) {
...
   }
   return llvm::StructType::get(Module->getContext(), Types);
 }
 
-llvm::Value *getBlockHeader(llvm::Module *Module, const KOREObjectSymbol *symbol) {
+llvm::Value *getBlockHeader(llvm::Module *Module, KOREDefinition *definition, const KOREObjectSymbol *symbol, llvm::Type *BlockType) {
   llvm::StructType *BlockHeaderType = Module->getTypeByName(BLOCKHEADER_STRUCT);
   uint64_t headerVal = symbol->getTag();
-  headerVal |= (uint64_t)symbol->length() << 32;
+  headerVal |= (llvm::DataLayout(Module).getTypeAllocSize(BlockType) / 8) << 32;
```
bmmoore: Truncating division looks a bit suspicious here. Maybe have a method return the block count, or round up here.

lib/codegen/CreateTerm.cpp
```diff
@@ -99,100 +101,334 @@ llvm::Type *getValueType(SortCategory sort, llvm::Module *Module) {
...
+    int i = 0;
+    for (auto val = applyRule->arg_begin(); val != applyRule->arg_end(); ++val, ++i) {
+      val->setName(paramNames[i]);
+      subst.insert({paramNames[i], val});
+    }
+    llvm::BasicBlock *block = llvm::BasicBlock::Create(Module->getContext(), "entry", applyRule);
+    llvm::BasicBlock *stuck = llvm::BasicBlock::Create(Module->getContext(), "stuck");
+    llvm::FunctionType *AbortType = llvm::FunctionType::get(llvm::Type::getVoidTy(Module->getContext()), false);
+    llvm::Function *AbortFunc = llvm::dyn_cast<llvm::Function>(Module->getOrInsertFunction("abort", AbortType));
+    AbortFunc->addFnAttr(llvm::Attribute::NoReturn);
```
bmmoore: is this idempotent? This looks like it might pile on a lot of duplicate attributes.

dwightguth: I mean the resulting llvm code doesn't have any duplicate noreturn attributes so I'm assuming yes.

lib/codegen/CreateTerm.cpp
dwightguth: I'm assuming that getTypeAllocSize returns a multiple of 8 because BlockType is guaranteed to be at least 64-bit aligned, but I will add an explicit assertion

lib/codegen/CreateTerm.cpp
bmmoore: Looks like there's some deduplication in the implementation of attribute lists or AttrBuilder.

PR #16, opened 2018-08-17
hook and function refactoring

dwightguth:
Changes hooks and functions to return their value and abort on error, so that they can be tail called by other functions.

### DIFF
--- a/runtime/datastructures/src/hook_set.rs
+++ b/runtime/datastructures/src/hook_set.rs
@@ -1,4 +1,4 @@
-use super::decls::{Set,List,Int,K,__gmpz_init_set_ui};
+use super::decls::{Set,List,Int,K,__gmpz_init_set_ui,move_int};
 use std::iter::FromIterator;
 use std::ptr;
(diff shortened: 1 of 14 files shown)


runtime/datastructures/src/hook_set.rs
```diff
@@ -14,72 +14,63 @@ pub unsafe extern "C" fn drop_set(ptr: *mut Set) {
...
+  __gmpz_init_set_ui(&mut result, (*s).len());
+  move_int(result)
 }
 
 #[no_mangle]
-pub unsafe extern "C" fn hook_SET_set2list(result: *mut List, s: *const Set) -> bool {
-  ptr::write(result, List::from_iter((*s).iter().map(|k| *k)));
-  true
+pub unsafe extern "C" fn hook_SET_set2list(s: *const Set) -> List {
+  List::from_iter((*s).iter().map(|k| *k))
```
bmmoore: This could use `.cloned()` instead of `map(|k| *k)`.

PR #27, opened 2018-08-26
Merged the haste code to the repository

raduom:


dwightguth:
Jenkins: add to whitelist

bmmoore:
You should rename the package along with the directory.

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "matching/submodules/kore"]
+	path = matching/submodules/kore
+	url = git@github.com:raduom/kore.git
```
dwightguth: shouldn't this point to kframework/kore?

matching/.gitignore
```diff
@@ -0,0 +1,3 @@
+.stack-work/
+haste.cabal
```
dwightguth: I believe this should now be matching.cabal.

matching/ChangeLog.md
```diff
@@ -0,0 +1,3 @@
+# Changelog for haste
```
dwightguth: I think we can delete this file.

matching/LICENSE
```diff
@@ -0,0 +1,30 @@
+Copyright Author name here (c) 2018
```
dwightguth: and this one.

matching/README.md
```diff
@@ -0,0 +1 @@
+# matching
```
dwightguth: and this one.

matching/package.yaml
```diff
@@ -0,0 +1,49 @@
+name:                matching
+version:             0.1.0.0
+github:              "githubuser/matching"
```
dwightguth: I believe this should be kframework/llvm-backend.

matching/package.yaml
```diff
@@ -0,0 +1,49 @@
...
+license:             BSD3
```
dwightguth: pretty sure this should be the UIUC license.

matching/package.yaml
```diff
@@ -0,0 +1,49 @@
...
+author:              "Author name here"
```
dwightguth: maybe we could fill these out?

matching/package.yaml
```diff
@@ -0,0 +1,49 @@
...
+maintainer:          "example@example.com"
+copyright:           "2018 Author name here"
+
+extra-source-files:
```
dwightguth: if we delete these files don't forget to delete this too.

matching/package.yaml
```diff
@@ -0,0 +1,49 @@
...
+- ChangeLog.md
+
+# Metadata used when publishing your package
+# synopsis:            Short description of your package
+# category:            Web
+
+# To avoid duplicated efforts in documentation and dealing with the
+# complications of embedding Haddock markup inside cabal files, it is
+# common to point users to the README.md file.
+description:         Please see the README on GitHub at <https://github.com/githubuser/matching#readme>
```
dwightguth: this should point to kframework/llvm-backend also.

matching/test/imp.kore
```diff
@@ -0,0 +1,2627 @@
+[]
```
dwightguth: can we delete this and point it at the same test/imp.kore in the root?

.gitmodules
raduom: I did think about that, but when my repository gets updated I can check it does not brake anything, so we get no real downtime. I could change to the kframework one though if anyone else wants to manage the possible braking scenarios.

dwightguth: it won't break anything anyway because it will be pointed at a specific commit and we can take the time to update at our leisure.

matching/LICENSE
dwightguth: okay, but in that case they should be updated with actual content.

matching/.gitignore
pmeredit: I see no reason not to make it *.cabal

matching/ChangeLog.md
pmeredit: I agree on removing this for now. We can always add it later if we decide to make this a Haskell module, since it contains no useful information now.

matching/LICENSE
pmeredit: I added our three names here, since we've touched the code in some manner or another (Radu first, of course). We'll just need to remember to update names if anyone else works on it.

matching/test/imp.kore
pmeredit: Deleted, it isn't actually being used right now by the haste test anyway.

PR #28, opened 2018-08-30
Code to deserialize and preprocess decision tree

dwightguth:


include/kllvm/codegen/Decision.h
```diff
@@ -30,8 +30,13 @@ class DecisionCase {
...
 public:
-  DecisionCase(KOREObjectSymbol *constructor, DecisionNode *child) :
-    constructor(constructor), child(child) {}
+  DecisionCase(
+    KOREObjectSymbol *constructor, 
+    std::vector<std::string> bindings,
+    DecisionNode *child) :
+      constructor(constructor),
+      bindings(bindings),
+      child(child) {}
```
pmeredit: Think we will not want to use strings for bindings. We should create some sort of interned strings that can be compared by pointer and call the class `Var` or similar. Interning the strings is some small overhead, but we can assume that the matching will be done many times.

pmeredit: Or, it's not completely clear to me, these bindings may not be used at runtime? If the bindings are not used after codegen, leaving them as strings is probably fine.

lib/codegen/DecisionParser.cpp
```diff
@@ -0,0 +1,154 @@
...
+          result->addBinding(constructors[constructors.size()-1-idx2]);
+        } else {
+          result->addBinding(parents[parents.size()-idx1]);
+        }
+      }
+      return result;
+    }
+    }     
+  }
+};
```
pmeredit: I kind of feel that each of these cases should be put in separate functions due to the length of this, but it's a nit.

include/kllvm/codegen/Decision.h
dwightguth: This is purely an identifier used in the compiler, it compiles at runtime into a local llvm value.

dwightguth:
Jenkins: test this please

lib/codegen/DecisionParser.cpp
```diff
@@ -0,0 +1,161 @@
...
+
+  DecisionNode *switchCase(Kind kind, YAML::Node node) {
+    YAML::Node list = node["specializations"];
+    std::string name = constructors.back();
+    constructors.pop_back();
+    auto result = SwitchNode::Create(name);
+    auto occurrence = node["occurrence"].as<std::vector<int>>();
+    for (auto iter = list.begin(); iter != list.end(); ++iter) {
+      auto _case = *iter;
+      std::vector<std::string> copy = constructors;
```
bmmoore: It looks strange to set up state for the child by modifying the field constructors, and then reset it from a copy in the local variable, rather than setting up the list for the child in the local variable.

PR #31, opened 2018-09-04
Many minor improvements to matching code

dwightguth:
Includes:

* enables `stack test` on CI build
* uses symbol names instead of integers for constructors
* supports matching on integer literals
* supports variable bindings in actions
* supports a side condition on clauses
* fixes bugs in the matching algorithm
* adds occurrences to algorithm

### DIFF
--- a/matching/src/Pattern.hs
+++ b/matching/src/Pattern.hs
@@ -27,7 +27,7 @@ import           Data.Functor.Classes  (Eq1 (..), Show1 (..))
 import           Data.Functor.Foldable (Fix (..), cata)
 import           Data.Maybe            (mapMaybe)
 import           Data.Semigroup        ((<>))
-import           Data.Text             (Text)
+import           Data.Text             (Text, pack)
 import           TextShow              (showt)
 import qualified Data.Yaml.Builder as Y
 import qualified Data.ByteString as B
@@ -35,28 +35,27 @@ import qualified Data.ByteString as B
 data Column = Column
               { getMetadata :: !Metadata
               , getTerms    :: ![Fix Pattern]
-              } deriving (Eq)
+              }
 
 instance Show Column where
   showsPrec _ (Column _ ts) =
     showString "Column " . showList ts
 
-newtype Metadata = Metadata [Metadata]
-                 deriving (Show, Eq)
+newtype Metadata = Metadata (Int, String -> [Metadata])
 
 type Index       = Int
-data Pattern a   = Pattern Index ![a]
+data Pattern a   = Pattern String ![a]
                  | Wildcard
                  | Var String
                  deriving (Show, Eq, Functor)
 
 newtype PatternMatrix = PatternMatrix [Column]
-                        deriving (Show, Eq)
+                        deriving (Show)
 
 type Action       = Int
 
 data ClauseMatrix = ClauseMatrix PatternMatrix ![Action]
-                    deriving (Show, Eq)
+                    deriving (Show)
 
 instance Show1 Pattern where
   liftShowsPrec _     _      _ Wildcard = showString "_"
@@ -94,7 +93,7 @@ failure = Fix Fail
 leaf :: Action -> Fix DecisionTree
 leaf a = Fix (Leaf (a, []))
 
-switch :: [(Index, Fix DecisionTree)]
+switch :: [(Text, Fix DecisionTree)]
        -> Maybe (Fix DecisionTree)
        -> Fix DecisionTree
 switch brs def =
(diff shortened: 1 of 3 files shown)


matching/src/Pattern.hs
```diff
@@ -120,7 +121,7 @@ swap ix tm = Fix (Swap ix tm)
 -- [ Matrix ]
```
raduom: These matrix functions are pretty difficult to read and / or reason about. How about some tests?

matching/src/Pattern.hs
```diff
@@ -27,36 +27,35 @@ import           Data.Functor.Classes  (Eq1 (..), Show1 (..))
...
-              } deriving (Eq)
+              }
 
 instance Show Column where
   showsPrec _ (Column _ ts) =
     showString "Column " . showList ts
 
-newtype Metadata = Metadata [Metadata]
-                 deriving (Show, Eq)
+newtype Metadata = Metadata (Int, String -> [Metadata])
```
raduom: What does the Int represent?
Also we should probably use a record, and store the function as two fields. Usually functions cannot be shown or compared for equality, etc.

matching/src/Pattern.hs
```diff
@@ -54,9 +56,10 @@ data Pattern a   = Pattern String (Maybe Int) ![a]
 newtype PatternMatrix = PatternMatrix [Column]
                         deriving (Show)
 
+type Occurrence   = [Int]
 type Action       = Int
 
-data ClauseMatrix = ClauseMatrix PatternMatrix ![Action]
+data ClauseMatrix = ClauseMatrix PatternMatrix ![(Action, [(String, Occurrence)])]
```
raduom: This represents the index you need on the LLVM side?

matching/src/Pattern.hs
```diff
@@ -132,15 +135,23 @@ sigma₁ :: PatternMatrix -> [String]
 sigma₁ (PatternMatrix (c : _)) = sigma c
 sigma₁ _                       = []
 
-mSpecialize :: String -> ClauseMatrix -> (Text, ClauseMatrix)
-mSpecialize ix = (pack ix, ) . expandMatrix ix . (filterMatrix (checkPatternIndex ix))
+mSpecialize :: String -> (ClauseMatrix, [Occurrence]) -> (Text, (ClauseMatrix, [Occurrence]))
+mSpecialize ix (cm, o : os) = (pack ix, (expandMatrix ix (filterMatrix (checkPatternIndex ix) (cm,o)), expandOccurrence cm o ix <> os))
```
raduom: These functions get pretty long. I would suggest giving them some names with a where clause.

matching/src/Pattern.hs
```diff
@@ -58,7 +58,7 @@ newtype PatternMatrix = PatternMatrix [Column]
                         deriving (Show)
 
 type Occurrence   = [Int]
-type Action       = Int
+type Action       = (Int, [String], Maybe [String])
```
raduom: This turns a bit complicated. Maybe we could use a record here.

matching/src/Pattern.hs
pmeredit: I, too, would like to know what the Int is. As soon as we have an MVP I think we should take the time to write a few comments on all the types/functions.

dwightguth: Can you please clarify what you mean by storing the function as two fields? I'm not sure I really follow what you're suggesting. Regardless, I didn't want to use a map type for this because a 32 bit integer would require 2^32 map entries. I'm aware that this makes it no longer possible to derive certain instances on these types, but it seems that they are not actually used. If you can suggest a better way for me to deal with exponentially many symbols in a sort, or otherwise clarify what you mean in this comment, I will fix it along with the others when I work on this PR tomorrow.

dwightguth: It will be made clear when I change this into a record type tomorrow, but the Int is the number of total constructors for this sort.

dwightguth: I actually need to change it to an Integer I think.

pmeredit:
This looks good to me, assuming the tests pass. I'm going to download your branch and play with it.

matching/src/Pattern.hs
pmeredit: Actually, I need a description of what Metadata does here at all, so I can start deriving it from Kore.  What are the Ints and Strings here?

dwightguth: The metadata is a good place to start because I already have some code in my working copy that generates the patterns and the actions. The int is the number of constructors in the sort which this metadata is represented by, and the function takes a pretty printed SymbolOrAlias and returns the list of metadata for each of the sorts of the arguments to that constructor.

dwightguth: Basically it's a lazily computed infinite data structure.

raduom: Ok. I see the point of the function.

dwightguth:
I've fixed most of your comments, but I didn't add any new tests yet. The tests that I wrote already adequately cover the behavior of the changes I made, and I'm hestitant to add tests for the lower level functions in the module because that would require exporting the internals of the module, which breaks encapsulation. I think the functional testing we're doing should be sufficient for our purposes, but if you disagree, let's discuss.

dwightguth:
Jenkins: test this please

PR #32, opened 2018-09-05
Fixes and improvements to C++ code

dwightguth:
Includes:
* generate functions to evaluate side conditions
* generate step function from decision tree
* generate static runtime functions needed for rewriting
* improvements to build system to enable compiling everything with lto
* print final configuration
* sync C++ code with latest haskell code's interface.

### DIFF
--- a/lib/ast/AST.cpp
+++ b/lib/ast/AST.cpp
@@ -384,8 +384,10 @@ void KOREDefinition::addAttribute(KOREPattern *Attribute) {
 
 void KOREDefinition::preprocess() {
   auto symbols = std::map<std::string, std::vector<KOREObjectSymbol *>>{};
+  unsigned nextOrdinal = 0;
   for (auto iter = axioms.begin(); iter != axioms.end();) {
     auto axiom = *iter;
+    axiom->ordinal = nextOrdinal++;
     if (!axiom->isRequired()) {
       iter = axioms.erase(iter);
     } else {
--- a/lib/ast/AST.cpp
+++ b/lib/ast/AST.cpp
@@ -327,7 +327,7 @@ KOREPattern *KOREAxiomDeclaration::getRequires() const {
       return nullptr;
     }
   }
-  assert(false && "could not compute side condition of axiom");
+  assert(false && "ill-formed axiom");
   abort();
 }
 
--- a/lib/ast/AST.cpp
+++ b/lib/ast/AST.cpp
@@ -298,6 +298,36 @@ KOREPattern *KOREAxiomDeclaration::getRightHandSide() const {
   assert(false && "could not compute right hand side of axiom");
 }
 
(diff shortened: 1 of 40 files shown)


dwightguth:
Jenkins: test this please

lib/ast/AST.cpp
```diff
@@ -298,6 +298,36 @@ KOREPattern *KOREAxiomDeclaration::getRightHandSide() const {
...
+          return equals->getArguments()[0];
+        } else if (equals->getConstructor()->getName() == "\\top" && equals->getArguments().empty()) {
+          return nullptr;
+        }
+      }
+    } else if (top->getConstructor()->getName() == "\\equals" && top->getArguments().size() == 2) {
+      return nullptr;
+    }
+  }
+  assert(false && "could not compute side condition of axiom");
```
theo25: It seems to me that you should return null here instead of assertion failure. Unless it is the case that for any axiom, we shlould never reach here. But even then the correct thing to do would be to make the `pattern` member of `KOREAxiomDeclalration` to be of `KOREObjectCompositePattern` type and avoid the if branch.

runtime/configurationparser/header.h
```diff
@@ -0,0 +1,81 @@
...
+  void *evaluateFunctionSymbol(uint32_t tag, void *arguments[]);
+  void *getToken(const char *sortname, uint64_t len, const char *tokencontents);
+
+  const char *getSymbolNameForTag(uint32_t tag);
+  const char *topSort(void);
+  void printMap(FILE *, map *, const char *, const char *, const char *);
+  void printSet(FILE *, set *, const char *, const char *, const char *);
+  void printList(FILE *, list *, const char *, const char *, const char *);
+  void visitChildren(block *subject, FILE *file,
+      void visitConfig(FILE *, block *, const char *),
```
theo25: Nit: If these are tabs, please remove them.

lib/ast/AST.cpp
dwightguth: I don't really see how changing the type of the declaration would help anything here. As it is, I would rather fail fast if we reach an axiom we don't understand than silently assume it doesn't have a side condition.

theo25: The point is I cannot tell what the API for this method is. When does it return `nullptr` and when does it abort? It seems to me that it returns `nullptr` when there is no side condition and aborts either when the cast fails or when the axiom's constructor isn't one of `\implies`, `\and`, or `\equals`. Are all of this invariants of our system?

If this is the case here, then the assertion is fine (although it's maybe good to change the type as suggested above to make it even clearer - not necesssary for the pr though). But please change the assertion message to something like `"axiom not well-formed"`, since I think when there is no side condition the expectation is to return `nullptr` rather than abort.

PR #33, opened 2018-09-07
Final linking of IMP

dwightguth:
This finally makes IMP link by generating all the right decision trees, but we still don't have a working interpreter. Further bug fixes will be pushed into the next PR.


dwightguth:
@theo25 @bmmoore @raduom this is ready for review.

PR #34, opened 2018-09-11
Add runtime library for string handling

pmeredit:
We don't seem to have any unit tests in the runtime, so I handled it by adding a main function for the testing.


runtime/strings/strings.cpp
```diff
@@ -0,0 +1,162 @@
+#include<assert.h>
+#include<stdlib.h>
+#include<iostream>
+
+extern "C" {
+
+// Assume assume length is passed before char *
+
+inline int get_len(const char * a) {
```
pmeredit: Assumption here is that function will be passed a string with the length prefixed _before_ the pointer of the beginning of the string. This is a pretty common way to allocate things. It can be easily changed if we want.

dwightguth:
Jenkins: add to whitelist

dwightguth:
I'm not going to review this in depth until the high level issue is fixed, but you should look at the %string llvm struct. This is the format we intend to use for strings, and you should be using %string * as the signature for the functions.

runtime/strings/strings.cpp
```diff
@@ -0,0 +1,533 @@
...
+	char data[0];
+};
+
+bool hook_STRINGS_gt(const string * a, const string * b) {
+    for (unsigned i = 0; i < a->b.len; ++i) {
+        if (a->data[i] < b->data[i]) return false;
+        if (a->data[i] > b->data[i]) return true;
+        if (i >= b->b.len)  return true;
+    }
+    return false;
```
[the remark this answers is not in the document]
dwightguth: I mean, it would need to be memcmp since strings in k can contain null bytes.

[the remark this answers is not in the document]
pmeredit: memcmp is actually a better idea.

runtime/strings/strings.cpp
```diff
@@ -0,0 +1,147 @@
...
+    return a->b.len == b->b.len;
+}
+
+bool hook_STRINGS_lt(const string * a, const string * b) {
+    for (int i = 0; i < a->b.len; ++i) {
+        if (a->data[i] > b->data[i]) return false;
+        if (a->data[i] < b->data[i]) return true;
+        if (i >= b->b.len)  return false;
+    }
+    return a->b.len < b->b.len;
```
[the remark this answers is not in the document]
pmeredit: It worked because the functions aren't symmetric. It's looping over the length of the left one always. Regardless, I'm going to switch to memcmp, which probably makes use of SSE.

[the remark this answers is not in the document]
pmeredit: That is to say, if I'd always looped over the larger string, I could have made the code symmetric. I was trying to reduce the code as much as possible, but in the end I wager memcpy will be better.

runtime/strings/strings.cpp
```diff
@@ -0,0 +1,147 @@
...
+    for (int i = 0; i < a->b.len; ++i) {
+        if (a->data[i] != b->data[i]) return true;
+        if (i >= b->b.len)   return true;
+    }
+    return a->b.len != b->b.len;
+}
+
+}
+
+int main() {
```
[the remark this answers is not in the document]
pmeredit: The new structure does use #ifdefs, but I'll still put it in a different file.

runtime/strings/strings.cpp
```diff
@@ -0,0 +1,533 @@
...
+	return makeString(out.str().c_str());
+}
+
+double hook_STRINGS_string2float(const string * input) {
+	auto si = std::string(input->data, input->b.len);
+	return std::stod(si);
+}
+
+// uses std::string for memory allocation, double copy here is not the best, but at least the
+// first copy is released at the end of the function.
```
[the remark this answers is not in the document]
pmeredit: You mean make makeString use realloc? That seems like a good idea.

runtime/strings/strings.cpp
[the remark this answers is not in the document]
dwightguth: You misunderstand, the unit tests should be in the unittests directory and make use of the boost test framework that we are using. I would recommend renaming the arithmetic folder "runtime" and adding the String tests there.

runtime/strings/strings.cpp
```diff
@@ -0,0 +1,533 @@
...
+				}
+				found = true;
+			}
+			if (found) return i;
+		}
+	}
+	return -1L;
+}
+
+// rfind will be used much less than find; it is ok to be a little less efficient and create std::strings.
```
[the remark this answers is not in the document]
pmeredit: find requires a std::string this. If you look in rfind, I am passing the char* data rather than creating two strings.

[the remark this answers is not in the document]
pmeredit: I was just too lazy to do an rfind because it isn't clear to me exactly how it works (I've never had a reason to use such a thing).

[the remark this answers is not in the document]
bmmoore: The STL function `std::find` takes a pair of iterators, it should work just fine with a pair of `char *`.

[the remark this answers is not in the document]
bmmoore: Oops, it's actually `std::search` that can do what you need. https://en.cppreference.com/w/cpp/algorithm/search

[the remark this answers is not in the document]
pmeredit: Ah! Now it makes sense. I'll use search.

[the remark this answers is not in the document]
pmeredit: Also, I'll look into rust, like you mentioned, if need be.

CMakeLists.txt
```diff
@@ -7,8 +7,8 @@ set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 add_definitions(-Wall -Werror -flto)
 set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -flto")
-set(CMAKE_AR "llvm-ar")
-set(CMAKE_RANLIB "llvm-ranlib")
+set(CMAKE_AR "llvm-ar-5.0")
```
dwightguth: I think it's better to set this to the system clang compiler version, otherwise  we could have problems on other operating systems. So maybe revert this? You can override the variables manually in your working copy if you need it on your machine in order to build.

runtime/strings/strings.cpp
```diff
@@ -0,0 +1,291 @@
...
+	memcpy(&(ret->data), &(a->data), a->b.len * sizeof(KCHAR));
+	memcpy(&(ret->data[a->b.len]), &(b->data), b->b.len * sizeof(KCHAR));
+	return ret;
+}
+
+int64_t hook_STRING_length(const string * a) {
+	return a->b.len;
+}
+
+string * hook_STRING_chr(const int64_t ord) {
```
dwightguth: chr takes an Int so this needs to be an mpz_t. You should use mpz_fits_ulong_p and mpz_get_ui to convert to a uint64_t, and then abort if mpz_fits_ulong_p returns false or if the uint64_t is greater than 255.

runtime/strings/strings.cpp
```diff
@@ -0,0 +1,291 @@
...
+	auto ret = static_cast<string *>(malloc(sizeof(string) + sizeof(KCHAR)));
+	ret->b.len = 1;
+	ret->data[0] = static_cast<KCHAR>(ord);
+	return ret;
+}
+
+int64_t hook_STRING_ord(const string * input) {
```
dwightguth: similarly, this returns an mpz_ptr which you should malloc and then call mpz_init_set_ui. Note that you should cast to an unsigned type rather than a signed type, otherwise if you're on a system where `char` is signed, you might return a negative number from this function.

runtime/strings/strings.cpp
```diff
@@ -0,0 +1,291 @@
...
+	if (input->b.len < 1) {
+		return -1;
+	}
+    return static_cast<int64_t>(input->data[0]);
+}
+
+// -1 means take the entire rest of the string.
+string * hook_STRING_substr(const string * input, const int64_t start, int64_t len) {
```
dwightguth: the hook STRING.substr actually takes a [start..end) range, not a start and a length. It should abort if start or end is greater than the length, or if end < start.

dwightguth: also note again that it takes an Int, so the type should be mpz_t.

runtime/strings/strings.cpp
```diff
@@ -0,0 +1,291 @@
...
+	if (len == -1) len = input->b.len;
+	auto minLen = std::max(std::min(len, input->b.len - start), 0L);
+	auto ret = static_cast<string *>(malloc(sizeof(string) + sizeof(KCHAR) * minLen));
+	ret->b.len = minLen;
+	memcpy(&(ret->data), &(input->data[start]), minLen * sizeof(KCHAR));
+	return ret;
+}
+
+int64_t hook_STRING_find(const string * haystack, const string * needle, int64_t pos) {
```
dwightguth: see above re: mpz_t

runtime/strings/strings.cpp
```diff
@@ -0,0 +1,291 @@
...
+		return -1;
+	}
+	auto out = std::search(haystack->data + pos * sizeof(KCHAR), haystack->data + haystack->b.len * sizeof(KCHAR),
+			               needle->data,   needle->data   + needle->b.len * sizeof(KCHAR));
+	int64_t ret = (out - haystack->data) / sizeof(KCHAR);
+	// search returns the end of the range if it is not found, but we want -1 in such a case.
+	return (ret < haystack->b.len)?ret:-1;
+}
+
+int64_t hook_STRING_rfind(const string * haystack, const string * needle, int64_t pos) {
```
dwightguth: same

runtime/strings/strings.cpp
```diff
@@ -0,0 +1,291 @@
...
+	}
+	auto end = (pos < haystack->b.len)?pos:haystack->b.len;
+	auto out = std::find_end(&haystack->data[0], &haystack->data[end],
+                  &needle->data[0], &needle->data[needle->b.len]);
+	auto ret = &*out - &haystack->data[0];
+	return (ret < end)?ret:-1;
+}
+
+int64_t hook_STRING_findChar(const string * haystack, const string * needle, int64_t pos) {
+	if (needle->b.len > 1) return -1;
```
dwightguth: reasonably sure this should actually abort.

runtime/strings/strings.cpp
```diff
@@ -0,0 +1,291 @@
...
+	return hook_STRING_find(haystack, needle, pos);
+}
+
+int64_t hook_STRING_rfindChar(const string * haystack, const string * needle, int64_t pos) {
+	if (needle->b.len > 1) return -1;
```
dwightguth: same.

runtime/strings/strings.cpp
```diff
@@ -0,0 +1,291 @@
...
+	auto len = strlen(input);
+	auto ret = static_cast<string *>(malloc(sizeof(string) + len));
+	for (unsigned i = 0; i < len; ++i) {
+		ret->data[i] = input[i];
+	}
+	ret->b.len = len;
+	return ret;
+}
+
+string * hook_STRING_float2string(const double input) {
```
dwightguth: don't worry about implementing this function yet, but note that floats are implemented by mpfr_t, so the correct behavior here is to call mpfr_get_str with a base of 10 and an n of 0 (so that it generates exactly as many digits as required not to lose any precision). For now just throw an exception here with a not implemented error.

runtime/strings/strings.cpp
```diff
@@ -0,0 +1,291 @@
...
+	std::stringstream out;
+	out << std::setprecision(14) << input;
+	return makeString(out.str().c_str());
+}
+
+double hook_STRING_string2float(const string * input) {
```
dwightguth: this is going to return an mpfr pointer and will call mpfr_set_str, but again, just throw an exception for now. Do fix the signature on all of these functions though.

runtime/strings/strings.cpp
```diff
@@ -0,0 +1,291 @@
...
+}
+
+const string * hook_STRING_string2token(const string * input) {
+	return input;
+}
+
+// uses std::string for memory allocation, double copy here is not the best, but at least the
+// first copy is released at the end of the function.
+string * hook_STRING_replaceAll(const string * input, const string * matcher, const string * replacer) {
+	std::string ret;
```
dwightguth: this is not really very great because you are treating the return value as a C string when in fact it could contain null bytes. I'm also reasonably sure that you should be using the same find algorithm as STRING.find. Basically you should have a helper function that finds the needle from some index i in the haystack, call that with index 0 when doing STRING.find, and then for replaceAll, you should call it with 0, if it finds it, then you replace and find again immediately after the end of the replacement, iterating until you reach the end of the string.

runtime/strings/strings.cpp
```diff
@@ -0,0 +1,291 @@
...
+			ret.push_back(input->data[i]);
+			++i;
+		}
+	}
+	return makeString(ret.c_str());
+}
+
+// uses std::string for memory allocation, double copy here is not the best, but at least the
+// first copy is released at the end of the function.
+string * hook_STRING_replace(const string * input, const string * matcher, const string * replacer, int64_t occurences) {
```
dwightguth: this should be the same as replaceAll except it replaces at most occurrences times.

runtime/strings/strings.cpp
```diff
@@ -0,0 +1,291 @@
...
+			ret.push_back(input->data[i]);
+			++i;
+		}
+	}
+	return makeString(ret.c_str());
+}
+
+// uses std::string for memory allocation, double copy here is not the best, but at least the
+// first copy is released at the end of the function.
+string * hook_STRING_replaceFirst(const string * input, const string * matcher, const string * replacer) {
```
dwightguth: this should just call replace with an occurrences of 1.

runtime/strings/strings.cpp
```diff
@@ -0,0 +1,291 @@
...
+		} else {
+			ret.push_back(input->data[i]);
+			++i;
+		}
+	}
+	return makeString(ret.c_str());
+}
+
+int64_t hook_STRING_countAllOccurences(const string * input, const string * matcher) {
+	int64_t ret = 0;
```
dwightguth: again, this should be using std::search.

CMakeLists.txt
bmmoore: I disagree. Because we are specifying LLVM 5 with `find_package(LLVM 5 REQUIRED CONFIG)` we should also be using the llvm 5 version of all other llvm tools to avoid any compatibility issues.

dwightguth:
Oh, one other thing, when I said abort above I actually meant it should throw the same type of exception as the int hooks.

CMakeLists.txt
pmeredit: I actually hadn't meant to check that in, but now that Brandon mentions it, I think it's worth keeping. This keeps us safer.

dwightguth:
Better but there's still a few issues.

dwightguth:
I have fixed my remaining issues with this PR. @bmmoore can you take another look?

runtime/strings/strings.cpp
```diff
@@ -1,291 +1,296 @@
...
 
-const string * hook_STRING_string2token(const string * input) {
-	return input;
-}
+  const string * hook_STRING_int2string(const mpz_t input) {
+    size_t len = mpz_sizeinbase(input, 10) + 2;
+    // +1 for null terminator needed by mpz_get_str
+    auto result = static_cast<string *>(malloc(sizeof(string) + len));
+    mpz_get_str(result->data, 10, input);
+    result->b.len = strlen(result->data);
```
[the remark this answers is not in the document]
bmmoore: Actually our allocator could probably support reducing the size of the string (at least if it's the most recently allocated). For now we should just `realloc` to the correct size.

PR #35, opened 2018-09-14
Imp working

dwightguth:
Haven't tested with all the imp programs, but collatz and sum-sbc seem to work.


lib/ast/AST.cpp
```diff
@@ -440,9 +441,14 @@ void KOREDefinition::preprocess() {
     for (auto iter = entry.second.begin(); iter != entry.second.end(); ++iter) {
       KOREObjectSymbol *symbol = *iter;
       if (!symbol->isConcrete()) {
-        assert(symbol->isPolymorphic() && "Unsupported polymorphism");
```
pmeredit: Nice that polymorphism is supported now.

runtime/configurationparser/header.h
```diff
@@ -68,14 +69,15 @@ extern "C" {
...
-  void visitMInt(FILE *, void *, const char *));
+      void visitConfig(FILE *, block *, const char *), 
+      void visitMap(FILE *, map *, const char *, const char *, const char *), 
+      void visitList(FILE *, list *, const char *, const char *, const char *), 
+      void visitSet(FILE *, set *, const char *, const char *, const char *), 
+      void visitInt(FILE *, mpz_t, const char *),
+      void visitFloat(FILE *, mpfr_t, const char *),
+      void visitBool(FILE *, bool, const char *),
+      void visitMInt(FILE *, void *, const char *),
+      void visitSeparator(FILE *));
```
pmeredit: So this is interesting to me. I used KCHAR (currently defined to char) in the strings runtime lib. Are we sure that we will stick with chars going forward? I suppose, even if we do decided to support unicode, that we'll stick with utf-8 instead of the abominations of utf-16 or utf-32.

runtime/equality.ll
```diff
@@ -0,0 +1,134 @@
...
+  unreachable
+}
+
+define i1 @hook_KEQUAL_ne(%block* %arg1, %block* %arg2) {
+  %eq = call i1 @hook_KEQUAL_eq(%block* %arg1, %block* %arg2)
+  %ne = xor i1 %eq, 1
+  ret i1 %ne
+}
+
+attributes #0 = { noreturn }
```
pmeredit: Why are these files defined directly in llvm asm?

ciscript
```diff
@@ -7,7 +7,7 @@ make -j`nproc`
 make run-unittests
 cd ../runtime/datastructures
 cargo test
-cargo rustc -- --emit=llvm-ir
+cargo build --release
```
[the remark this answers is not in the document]
dwightguth: as a matter of fact, that's not true, because cargo build --release generates .rlib files which contain .bc files inside them.

PR #36, opened 2018-09-20
Bug fixes and features required for IELE

dwightguth:
* pattern matching on Ints and Strings
* new array hooks and array bugfixes
* improve performance of haskell
* as patterns
* compute best column using "computeScore" function
* various minor bugfixes

### DIFF
--- a/lib/codegen/CreateTerm.cpp
+++ b/lib/codegen/CreateTerm.cpp
@@ -327,6 +327,7 @@ llvm::Value *CreateTerm::createHook(KOREObjectCompositePattern *hookAtt, KOREObj
   } else {
     std::string domain = name.substr(0, name.find('.'));
     if (domain == "ARRAY") {
+      // array and list are backed by the same type in llvm
       domain = "LIST";
     }
     std::string hookName = "hook_" + domain + "_" + name.substr(name.find('.') + 1);
--- a/lib/codegen/CreateTerm.cpp
+++ b/lib/codegen/CreateTerm.cpp
@@ -325,7 +325,11 @@ llvm::Value *CreateTerm::createHook(KOREObjectCompositePattern *hookAtt, KOREObj
     assert(false && "not implemented yet: MInt");
     abort();
   } else {
-    std::string hookName = "hook_" + name.substr(0, name.find('.')) + "_" + name.substr(name.find('.') + 1);
+    std::string domain = name.substr(0, name.find('.'));
+    if (domain == "ARRAY") {
+      domain = "LIST";
+    }
+    std::string hookName = "hook_" + domain + "_" + name.substr(name.find('.') + 1);
     return createFunctionCall(hookName, pattern, true);
   }
 }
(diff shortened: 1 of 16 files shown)


lib/codegen/CreateTerm.cpp
```diff
@@ -325,7 +325,11 @@ llvm::Value *CreateTerm::createHook(KOREObjectCompositePattern *hookAtt, KOREObj
     assert(false && "not implemented yet: MInt");
     abort();
   } else {
-    std::string hookName = "hook_" + name.substr(0, name.find('.')) + "_" + name.substr(name.find('.') + 1);
+    std::string domain = name.substr(0, name.find('.'));
+    if (domain == "ARRAY") {
+      domain = "LIST";
+    }
```
pmeredit: Let's comment why we are changing ARRAY to LIST.

lib/codegen/Decision.cpp
```diff
@@ -82,6 +82,37 @@ void SwitchNode::codegen(Decision *d, llvm::StringMap<llvm::Value *> substitutio
...
 
+void EqualsLiteralNode::codegen(Decision *d, llvm::StringMap<llvm::Value *> substitution) {
+  std::string funcName;
+  switch (cat) {
+  case SortCategory::Uncomputed:
+  case SortCategory::Map:
+  case SortCategory::List:
+  case SortCategory::Set:
+  case SortCategory::Bool:
+  case SortCategory::MInt:
```
pmeredit: Do we expect to support MInt later? Seems like it could be useful for performance.... also Bool.

lib/codegen/DecisionParser.cpp
```diff
@@ -16,11 +16,12 @@ class DTPreprocessor {
   KOREObjectSymbol *dv;
 
   enum Kind {
-    Switch, SwitchLit, Function, Leaf, Fail, Swap
+    Switch, SwitchLit, EqualsLiteral, Function, Leaf, Fail, Swap
```
pmeredit: I'm not a fan of the inconsistent naming between SwitchLit and EqualsLiteral. Let's standardize on Lit or Literal (fine with either).

lib/codegen/EmitConfigParser.cpp
```diff
@@ -460,7 +465,7 @@ static void getVisitor(KOREDefinition *definition, llvm::Module *module, KOREObj
   auto BlockType = getBlockType(module, definition, symbol);
   auto cast = new llvm::BitCastInst(func->arg_begin(),
       llvm::PointerType::getUnqual(BlockType), "", CaseBlock);
-  int i = 0;
+  unsigned i = 0;
```
pmeredit: Always approve of changing signed to unsigned where possible.

lib/codegen/Decision.cpp
dwightguth: No, those cases are supported by the SwitchLit node.

PR #37, opened 2018-09-26
List patterns, nonlinear patterns, and minor bug fixes

dwightguth:


matching/src/Pattern.hs
```diff
@@ -74,13 +74,22 @@ data Metadata = Metadata
...
-type Constructor = Either (SymbolOrAlias Object) String
+data Constructor = Symbol (SymbolOrAlias Object)
+                 | Literal String
+                 | List (SymbolOrAlias Object) Int
+                 deriving (Show, Eq, Ord)
 
 type Index       = Int
 data Pattern a   = Pattern Constructor (Maybe String) ![a]
-                 | As String a
+                 | ListPattern
```
raduom: What do these fields mean?

dwightguth: I added some comments explaining.

raduom: So we never match in the middle?

dwightguth: Correct

dwightguth:
@raduom I wrote up the first section of the document describing our modifications to the matching algorithm, which includes a formal description of the treatment of variables and lists found in this pr. It's here if you want to read it: https://github.com/kframework/llvm-backend/wiki/Pattern-Matching-Algorithm

dwightguth:
I will fill out the sections on map and set patterns before posting the next pr.

raduom:
> @raduom I wrote up the first section of the document describing our modifications to the matching algorithm, which includes a formal description of the treatment of variables and lists found in this pr. It's here if you want to read it: https://github.com/kframework/llvm-backend/wiki/Pattern-Matching-Algorithm

I am having some issues asking questions about your description. Do you think it would be better to use overleaf (the old sharelatex) to collaborate on this document and have comments and whatnot until it is more stable? I use overleaf quite a bit, and it would be easier for me, but I think google docs would do just as well.

matching/src/Pattern.hs
```diff
@@ -478,13 +571,38 @@ compilePattern cm =
...
     isWildcard (Fix (Pattern _ _ _)) = False
+    isWildcard (Fix (ListPattern _ _ _ _)) = False
     equalLiteral :: Int -> [Occurrence] -> String -> [(Text, (ClauseMatrix, [Occurrence]))] -> Maybe (ClauseMatrix, [Occurrence]) -> Fix DecisionTree
     equalLiteral o os _ [] (Just d) = Fix $ Switch (head os) L { getSpecializations = [], getDefault = Just $ compilePattern' o d }
     equalLiteral _ _ _ [] (Nothing) = Fix Fail
-    equalLiteral o os hookName ((name,spec):tl) d = Fix $ EqualLiteral (pack hookName) name $ Fix $ SwitchLiteral [o, 0] 1 $ L [("1", Fix $ Switch (head os) L { getSpecializations = [], getDefault = Just $ compilePattern' (o+1) spec }),("0", equalLiteral (o+1) os hookName tl d)] Nothing
+    equalLiteral o os hookName ((name,spec):tl) d = Fix $ EqualLiteral (pack hookName) [o, 0] (head os) name $ Fix $ SwitchLiteral [o, 0] 1 $ L [("1", Fix $ Switch (head os) L { getSpecializations = [], getDefault = Just $ compilePattern' (o+1) spec }),("0", equalLiteral (o+1) os hookName tl d)] Nothing
+    listPattern :: Int -> [Occurrence] -> [(Text, (ClauseMatrix, [Occurrence]))] -> Maybe (ClauseMatrix, [Occurrence]) -> [Constructor] -> Column -> Fix DecisionTree
```
raduom: It is not clear what every argument means here. Can you:
a. Comment on the argument meaning in a comment on the function `or`
b. Use more descriptive variable names.

dwightguth:
@raduom I have tried to address all your comments and also to add documentation to getLeaf and compilePattern. If there are other functions you want me to extensively document, let me know and I will do the same there. If you have any other comments, let me know. I believe this should be ready for you to re-review.

include/kllvm/codegen/CreateTerm.h
```diff
@@ -36,7 +36,8 @@ class CreateTerm {
    an llvm value corresponding to the specified KORE RHS pattern and substitution in the
    specified definition, and returns the value itself. */
   llvm::Value *operator()(KOREPattern *pattern);
-  llvm::Value *createToken(SortCategory sort, std::string contents);
+  llvm::Value *createToken(ValueType sort, std::string contents);
+  llvm::Value *createFunctionCall(std::string name, ValueType returnCat, std::vector<llvm::Value *> &args, bool sret, bool load);
```
bmmoore: Please document. Is this for calling K functions? Are the arguments all supposed to be K terms? (or is this a utility for making an LLVM call instruction?)

lib/ast/AST.cpp
```diff
@@ -77,9 +78,12 @@ SortCategory KOREObjectCompositeSort::getCategory(std::string name) {
   else if (name == "FLOAT.Float") category = SortCategory::Float;
   else if (name == "BUFFER.StringBuffer") category = SortCategory::StringBuffer;
   else if (name == "BOOL.Bool") category = SortCategory::Bool;
-  else if (name == "MINT.MInt") category = SortCategory::MInt;
+  else if (name.substr(0, 9) == "MINT.MInt") {
```
bmmoore: Maybe a line comment `// MInt sorts have names like MINT.MInt128`?

lib/codegen/CreateTerm.cpp
```diff
@@ -488,28 +499,48 @@ bool makeFunction(std::string name, KOREPattern *pattern, KOREDefinition *defini
     llvm::StringMap<llvm::Type *> params;
     std::vector<llvm::Type *> paramTypes;
     std::vector<std::string> paramNames;
+    std::vector<bool> needsLoad;
```
bmmoore: This seems to be changing (or adding) a policy for how we call functions taking builtin collections,
passing a pointer directly to the off-heap collection value.
Is how we pass around collections documented somewhere?

lib/codegen/Decision.cpp
```diff
@@ -65,12 +65,21 @@ void SwitchNode::codegen(Decision *d, llvm::StringMap<llvm::Value *> substitutio
...
-        substitution[binding] = new llvm::LoadInst(ChildPtr, binding, d->CurrentBlock);
+        llvm::Value *ChildPtr = llvm::GetElementPtrInst::CreateInBounds(BlockType, Cast, {llvm::ConstantInt::get(llvm::Type::getInt64Ty(d->Ctx), 0), llvm::ConstantInt::get(llvm::Type::getInt32Ty(d->Ctx), offset+2)}, "", d->CurrentBlock);
+        switch (dynamic_cast<KOREObjectCompositeSort *>(_case.getConstructor()->getArguments()[offset++])->getCategory(d->Definition).cat) {
+        case SortCategory::Map:
+        case SortCategory::List:
+        case SortCategory::Set:
+          substitution[binding] = ChildPtr;
+          break;
+        default:
+          substitution[binding] = new llvm::LoadInst(ChildPtr, binding, d->CurrentBlock);
```
bmmoore: Is this consistent with the function call above? It looks like this is making the substitution binding hold the address of the field within the parent term, rather than something that only references the collection.

Also, if so, pointers to the interior of objects can make garbage collection harder.

include/kllvm/ast/AST.h
```diff
@@ -84,7 +84,12 @@ class KOREMetaSortVariable : public KOREMetaSort {
 };
 
 enum class SortCategory {
-  Uncomputed, Map, List, Set, Int, Float, StringBuffer, Bool, MInt, Symbol
+  Uncomputed, Map, List, Set, Int, Float, StringBuffer, Bool, Symbol, MInt
```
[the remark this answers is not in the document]
dwightguth: so that we can represent an MInt when doing hashing and equality using its category plus its bitwidth.

lib/codegen/CreateTerm.cpp
dwightguth: Not yet, but I don't want to do it now as I had to change it in the following pull request, so this code is actually deprecated.

lib/codegen/Decision.cpp
dwightguth: it's consistent, yes. we pass map/list/set by reference (ie via a pointer) and everything else by value, hence the load.

bmmoore:
Comments are good. We need some updated documentation around the calling convention, but we can skip that in this PR if some of what we see here is already going to be replaced.

PR #38, opened 2018-09-26
Bytes and StringBuffer hooks

dwightguth:


runtime/strings/bytes.cpp
```diff
@@ -0,0 +1,176 @@
...
+      }
+    }
+    return move_int(result);
+  }
+
+  string *hook_BYTES_int2bytes(mpz_t len, mpz_t i, uint64_t endianness) {
+    if (!mpz_fits_ulong_p(len)) {
+      throw std::invalid_argument("Integer overflow on len");
+    }
+    unsigned long len_long = mpz_get_ui(len);
```
pmeredit: This common pattern of 
    
```
    if (!mpz_fits_ulong_p(len)) {
      throw std::invalid_argument("Integer overflow on len");
    }
    unsigned long len_long = mpz_get_ui(len);
```

Should be extracted out to a simple function like the one I had in strings, however, `gs` is a terrible name. Maybe `get_ui_checked`.

runtime/strings/bytes.cpp
```diff
@@ -0,0 +1,176 @@
...
+    mpz_clear(twos);
+    return result;
+  }
+
+  string *hook_BYTES_bytes2string(string *b) {
+    size_t size = sizeof(string) + b->b.len;
+    string *result = static_cast<string *>(malloc(size));
+    memcpy(result, b, size);
+    return result;
+  }
```
pmeredit: So this is essentially just copy.
I would throw an `inline` on this. It's likely the compiler will inline it, anyway, but I think it helps for readability.

runtime/strings/strings.cpp
```diff
@@ -164,8 +171,10 @@ extern "C" {
...
   }
 
-  string * makeString(const KCHAR * input) {
-    auto len = strlen(input);
+  string * makeString(const KCHAR * input, ssize_t len = -1) {
+    if (len == -1) {
+      len = strlen(input);
+    }
     auto ret = static_cast<string *>(malloc(sizeof(string) + len));
     for (unsigned i = 0; i < len; ++i) {
```
bmmoore: memcpy is a bit better.

runtime/strings/bytes.cpp
pmeredit: lgtm

runtime/strings/bytes.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+  uint64_t tag_big_endian() {
+    return (((uint64_t)getTagForSymbolName("LblbigEndianBytes{}")) << 32) | 1;
+  }
+
+  uint64_t tag_unsigned() {
+    return (((uint64_t)getTagForSymbolName("LblunsignedBytes{}")) << 32) | 1;
+  }
+
+  mpz_ptr hook_BYTES_bytes2int(string *b, uint64_t endianness, uint64_t signedness) {
```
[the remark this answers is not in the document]
dwightguth: This is just left over from the K signature we came up with for these functions, which takes constant KApplies instead of booleans.

PR #41, opened 2018-10-17
New allocator and GC

dwightguth:
Note that ints, maps, sets, and lists are still allocated with malloc and are not yet freed. This needs to be addressed in a future PR.

What this does is create the bump allocator (which brandon worked on and I optimized) and a simple copying semispace collector for K terms (except ints, maps, sets, and lists as mentioned above).

It makes performance a little bit worse, but that's an acceptable cost for now and will be improved upon in the future with the generational collection scheme we plan to work on next. Right now most of the performance cost is in malloc and page faults, which seem to arise primarily from the builtins that are not yet being collected.

@bmmoore, please  review the code that I wrote. @pmeredit , please review everything, but especially the commits that brandon wrote.


dwightguth:
Jenkins: test this please

dwightguth:
Jenkins: test this please

runtime/alloc/alloc.c
```diff
@@ -0,0 +1,138 @@
+#include <string.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <stdint.h>
+#include <stdbool.h>
+#include "jemalloc/jemalloc.h"
```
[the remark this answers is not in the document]
dwightguth: yeah, other stuff is still calling malloc for right now.

lib/codegen/CreateTerm.cpp
```diff
@@ -388,7 +389,7 @@ llvm::Value *CreateTerm::createFunctionCall(std::string name, ValueType returnCa
   }
   if (sret) {
     // we don't use alloca here because the tail call optimization pass for llvm doesn't handle correctly functions with alloca
-    AllocSret = allocateBlock(returnType, CurrentBlock);
+    AllocSret = allocateBlock(returnType, CurrentBlock, "malloc");
```
bmmoore: Using malloc here at every struct return is very unfortunate. What is the specific issue, so we can track when it's fixed?

If we can't put these on the stack then we might want need to go to the trouble of restructuring this code to pass down a pointer to a final destination.

lib/codegen/EmitConfigParser.cpp
```diff
@@ -145,7 +146,7 @@ static llvm::Value *getArgValue(llvm::Value *ArgumentsArray, int idx,
     auto cast = new llvm::BitCastInst(arg,
         llvm::PointerType::getUnqual(getValueType(cat, mod)), "", CaseBlock);
     auto load = new llvm::LoadInst(cast, "", CaseBlock);
-    CaseBlock->getInstList().push_back(llvm::CallInst::CreateFree(arg, CaseBlock));
+//    CaseBlock->getInstList().push_back(llvm::CallInst::CreateFree(arg, CaseBlock));
```
bmmoore: I don't see a reason to keep this as a commented line.

lib/codegen/EmitConfigParser.cpp
```diff
@@ -306,17 +307,20 @@ static void emitGetToken(KOREDefinition *definition, llvm::Module *module) {
     }
     case SortCategory::Int: {
       llvm::Type *Int = module->getTypeByName(INT_STRUCT);
-      llvm::Value *Block = allocateBlock(Int, CaseBlock);
+      llvm::Instruction *Malloc = llvm::CallInst::CreateMalloc(
```
bmmoore: shouldn't this use `allocateBlock`?

lib/codegen/CreateTerm.cpp
```diff
@@ -65,8 +66,8 @@ target triple = "x86_64-unknown-linux-gnu"
 %blockheader = type { i64 } 
 %block = type { %blockheader, [0 x i64 *] } ; 16-bit layout, 8-bit length, 32-bit tag, children
 
-%layout = type { i8, %layoutitem* }
-%layoutitem = type { i64, i16 }
+%layout = type { i8, %layoutitem* } ; number of children, array of children
```
bmmoore: It looks like these layouts values will be separately named globals, so why does this have a pointer to items rather than directly including the array?

runtime/configurationparser/ConfigurationPrinter.cpp
```diff
@@ -39,7 +39,7 @@ void printConfigurationInternal(FILE *file, block *subject, const char *sort) {
   uint16_t layout = subject->header.header >> 48;
   if (!layout) {
     string *str = (string *)subject;
-    size_t len = subject->header.header & 0xffffffffffffLL;
+    size_t len = subject->header.header & 0x3fffffffffffLL;
```
bmmoore: Another mask that should have a named declaration in a more central place.

runtime/datastructures/src/hook_list.rs
```diff
@@ -203,6 +203,13 @@ pub unsafe extern "C" fn printList(file: *mut FILE, list: *const List, unit: *co
   }
 }
 
+#[no_mangle]
+pub unsafe extern "C" fn list_foreach(list: *mut List, copy: extern fn(block: *mut K)) {
```
bmmoore: "copy" seems overly specific here. "process" could be fine. The rust ForEach trait just calls the function argument "f"

runtime/hash.ll
```diff
@@ -56,12 +56,12 @@ constant:
 block:
   %arghdrptr = getelementptr inbounds %block, %block* %arg, i64 0, i32 0, i32 0
   %arghdr = load i64, i64* %arghdrptr
-  call void @add_hash64(i8* %hasher, i64 %arghdr)
   %arglayout = lshr i64 %arghdr, 48
   %isString = icmp eq i64 %arglayout, 0
   br i1 %isString, label %hashString, label %hashChildren
 hashString:
-  %arglen = and i64 %arghdr, 281474976710655
+  %arglen = and i64 %arghdr, 70368744177663
```
bmmoore: Can this value be documented? (I know it's harder to abstract out a constant in a `.ll` file)

runtime/strings/bytes.cpp
```diff
@@ -22,7 +23,7 @@ extern "C" {
 
   bytes *hook_BYTES_empty() {
     static bytes empty;
-    empty.b.len = 0;
+    empty.b.len = 0x400000000000;
```
bmmoore: This should be provided by `alloc.h`

unittests/runtime-strings/bytestest.cpp
```diff
@@ -64,9 +64,9 @@ BOOST_AUTO_TEST_CASE(int2bytes) {
   mpz_init_set_ui(_0, 0);
 
   auto res = hook_BYTES_int2bytes(_0, _0, tag_big_endian());
-  BOOST_CHECK_EQUAL(0, res->b.len);
+  BOOST_CHECK_EQUAL(1LL << 46, res->b.len);
```
bmmoore: Hopefully these literals can be removed as well.

unittests/runtime-strings/stringtest.cpp
```diff
@@ -461,7 +461,7 @@ BOOST_AUTO_TEST_CASE(buffer_concat) {
   auto expected = static_cast<string *>(malloc(sizeof(string) + totalLen));
   expected->b.len = totalLen;
   memset(expected->data, 'a', totalLen);
-  BOOST_CHECK_EQUAL(totalLen, result->b.len);
+  BOOST_CHECK_EQUAL(totalLen | 0x400000000000, result->b.len);
```
bmmoore: Another literal bitmask in this test.

runtime/alloc/alloc.c
```diff
@@ -0,0 +1,138 @@
...
+
+static void freshBlock() {
+    char *nextBlock;
+    if (block_start == 0) {
+      nextBlock = megabyte_malloc();
+      first_block = nextBlock;
+      memory_block_header hdr;
+      hdr.next_block = 0;
+      hdr.semispace = true_is_fromspace;
+      memcpy(nextBlock, &hdr, sizeof(hdr));
```
bmmoore: Is this better than casting `nextBlock` to `memory_block_header*`?

runtime/alloc/alloc.c
```diff
@@ -0,0 +1,138 @@
...
+    } else {
+      nextBlock = *(char**)block_start;
+      if (block != block_end) {
+        if (block_end - block == 8) {
+          *(uint64_t *)block = 0x0000400000000000LL; // 8 bit sentinel value
+        } else {
+          *(uint64_t *)block = block_end - block - 8; // 16-bit or more sentinel value
```
bmmoore: Are these sentinals related to the usual object headers?

runtime/alloc/alloc.c
```diff
@@ -0,0 +1,138 @@
...
+    }
+    block = nextBlock + sizeof(memory_block_header);
+    block_start = nextBlock;
+    block_end = nextBlock + BLOCK_SIZE;
+    DBG("New block at %p (remaining %zd)\n", block, BLOCK_SIZE - sizeof(memory_block_header));
+}
+
+static void* __attribute__ ((noinline)) doAllocSlow(size_t requested) {
+  DBG("Block at %p too small, %zd remaining but %zd needed\n", block, block_end-block, requested);
+  if (requested > BLOCK_SIZE - sizeof(memory_block_header)) {
```
bmmoore: Do we have any provision for collecting these objects? Are we assuming they contain no pointers?

runtime/alloc/collect.c
```diff
@@ -0,0 +1,167 @@
...
+void map_foreach(void *, void(block**));
+void set_foreach(void *, void(block**));
+void list_foreach(void *, void(block**));
+
+static size_t get_size(uint64_t hdr, uint16_t layout) {
+  if (!layout) {
+    size_t size = (hdr + sizeof(block) + 7) & ~7;
+    return hdr & 0x400000000000LL ? 8 : size < 16 ? 16 : size;
+  } else {
+    return ((hdr >> 32) & 0xff) * 8;
```
bmmoore: more block header stuff that should probably be centralized into alloc.h

runtime/alloc/collect.c
```diff
@@ -0,0 +1,167 @@
...
+    *blockPtr = newBlock;
+  } else {
+    *blockPtr = *forwardingAddress;
+  }
+}
+
+static void migrate_once(block** blockPtr) {
+  block* currBlock = *blockPtr;
+  uintptr_t intptr = (uintptr_t) currBlock;
+  intptr = intptr & ~(BLOCK_SIZE-1);
```
bmmoore: more header bit stuff

runtime/alloc/collect.c
```diff
@@ -0,0 +1,167 @@
...
+static size_t get_size(uint64_t hdr, uint16_t layout) {
+  if (!layout) {
+    size_t size = (hdr + sizeof(block) + 7) & ~7;
+    return hdr & 0x400000000000LL ? 8 : size < 16 ? 16 : size;
+  } else {
+    return ((hdr >> 32) & 0xff) * 8;
+  }
+}
+
+static void migrate(block** blockPtr) {
```
bmmoore: When is it safe to call migrate instead of migrate_once?

include/runtime/alloc.h
```diff
@@ -0,0 +1,32 @@
...
+extern "C" {
+#endif
+
+extern const size_t BLOCK_SIZE;
+void* koreAlloc(size_t requested);
+void koreAllocSwap(void);
+void* koreAllocToken(size_t requested);
+void* koreResizeLastAlloc(void* oldptr, size_t newrequest, size_t oldrequest);
+
+#define len(s) ((s)->b.len & 0xffff3fffffffffff)
```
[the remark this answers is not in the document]
dwightguth: they're zero for strings, but this allows us to alos use this macro to mask out just the gc bits and leave the rest of the header as is, making this suitable for more than just strings.

include/runtime/alloc.h
```diff
@@ -0,0 +1,33 @@
...
+// the actual length is equal to the block header with the gc bits masked out.
+#define len(s) ((s)->b.len & 0xffff3fffffffffff)
+#define set_len(s, l) ((s)->b.len = (l) | (l > BLOCK_SIZE - sizeof(char *) ? 0x400000000000 : 0))
+
+typedef struct {
+  char* next_block;
+  bool semispace;
+} memory_block_header;
+
+#ifdef ALLOC_DBG
```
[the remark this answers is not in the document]
dwightguth: I mean it's being used by collect.c as well, that's why I put it here. I figured we'd use it just for logging related to memory management, but if you have another suggestion about where to place it, I'm listening.

lib/codegen/CreateTerm.cpp
dwightguth: it can't go on the stack because the alloca instruction will prevent the function from being tail recursive, which leads to stack overflows. I agree there is room for improvement here but since this was already the behavior I'd rather leave this for optimization later. It's not really the current performance bottleneck anyway.

include/runtime/alloc.h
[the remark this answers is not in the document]
bmmoore: Ah, if it's used in collect.c then this is a good place, but it should probably have a less generic name than DBG.

lib/codegen/CreateTerm.cpp
dwightguth: This can probably be changed, but I would rather address it separately. Should I create an issue?

lib/codegen/CreateTerm.cpp
bmmoore: We can leave it for now, sure. According to the latest sources it should tolerate instructions satisfying `isStaticAlloca`, whatever that means. http://llvm.org/doxygen/TailRecursionElimination_8cpp_source.html#l00095

bmmoore: Constant-sized alloca falling in the entry block.

runtime/alloc/alloc.c
dwightguth: we're assuming that they're strings, since those are the only objects capable of being that large. So they shouldn't have pointers. Eventually we probably just want to put these directly in the old generation, but right now they're not being collected, no.

ciscript
```diff
@@ -2,20 +2,21 @@
...
-./tools/test_rhs/test_rhs ../test/imp.kore | diff - ../test/imp.ll
-./tools/test_sideconditions/test_sideconditions ../test/imp.kore | diff - ../test/requires.ll
+#./tools/test_rhs/test_rhs ../test/test.kore | diff - ../test/test.ll
+#./tools/test_rhs/test_rhs ../test/imp.kore | diff - ../test/imp.ll
+#./tools/test_sideconditions/test_sideconditions ../test/imp.kore | diff - ../test/requires.ll
 
-./tools/test_configparser/test_configparser ../test/imp.kore > configparser.ll
-diff configparser.ll ../test/configparser.ll
+#./tools/test_configparser/test_configparser ../test/imp.kore > configparser.ll
+#diff configparser.ll ../test/configparser.ll
```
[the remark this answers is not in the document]
dwightguth: I deleted these because they were so fragile that they're constantly failing. We now have adequate coverage of this code elsewhere in our end to end tests. I have deleted the commented lines though.

runtime/alloc/alloc.c
bmmoore: Could there be pointers to them that we try to visit and get block metadata from during garbage collection? I think that was covered by the part where we check a header bit in migrate_once, but you should say for sure.

lib/codegen/CreateTerm.cpp
bmmoore: I asked why you did it this way. If there is a strong reason it would have deserved a comment. If nothing forced this choice then I think a more contiguous layout would probably be more efficient, but I suspect it's not currently a major cost.

I'm definitely not insisting on doing anything without profiling, let alone requesting a specific implementation change - once it matters we should consider the implementation of getLayout more holistically.

runtime/alloc/alloc.c
dwightguth: this is handled by the set_len macro in header.h. When the block header of strings are initialized, we compare on the length of the string and if it's large enough that this code path would have been exercised by the allocation, we set the NOT_YOUNG_OBJECT_BIT, which prevents us from trying to migrate the object during gc.

lib/codegen/CreateTerm.cpp
dwightguth: I think the only reason I did it this way was because it was easier, since it meant that the type returned by getLayoutData and the type of the layout object at allocation could be the same.

include/runtime/header.h
```diff
@@ -5,49 +5,63 @@
 #include <gmp.h>
 #include <mpfr.h>
 
+// the actual length is equal to the block header with the gc bits masked out.
+#define len(s) ((s)->h.hdr & 0xffff3fffffffffff)
```
bmmoore: A macro for masking out the GC bits sounds like a good thing to have, but a macro called `len` should just mask with `0x3fffffffffff` rather than ever returning parts of the high bytes. I also doubt that the immediate in the code can be shared, and while it might not actually be more expensive I doubt this immediate could possibly be cheaper to construct than `0x3fffffffffff`.

bmmoore: Plus, distinguishing a `len` and a macro to `clear_gc_bits` now will make it easier in the future if we want to try changing representations in a way that makes a difference between between suppressing GC state and computing string length.

PR #42, opened 2018-10-29
Reduce code size and compilation time by implementing sharing in decision trees in C++

dwightguth:
I measured the impact on code size and compilation time for EVM, and found that before this change, running `opt -O2` on the generated llvm code took 15.5 minutes and generated a bitcode file that was 13 megabytes. after the change is implemented, doing the same took 7 seconds and generated a bitcode file that was 2.5 megabytes. that's a 132x improvement in compilation time and a 5x reduction in code size.

It's been tested and passes all the tests in the repo as well as passing the evm test suite.


include/kllvm/codegen/Decision.h
```diff
@@ -69,60 +82,39 @@ class SwitchNode : public DecisionNode {
...
-  static MakeLiteralNode *Create(
-      const std::string &name,
-      ValueType cat,
-      const std::string literal,
-      DecisionNode *child) {
-    return new MakeLiteralNode(name, cat, literal, child);
-  }
-
-  virtual void codegen(Decision *d, llvm::StringMap<llvm::Value *> substitution);
+  virtual void collectUses(std::set<std::string> &vars) { if(cases.size() != 1 || cases[0].getConstructor()) vars.insert(name); for (auto _case : cases) { _case.getChild()->collectUses(vars); } }
```
bmmoore: Why do either of these conditions guard whether `name` should be added to vars?

include/kllvm/codegen/Decision.h
```diff
@@ -69,60 +82,39 @@ class SwitchNode : public DecisionNode {
...
       KOREObjectPattern *pattern,
+      std::vector<std::string> &uses,
       DecisionNode *child) {
-    return new MakePatternNode(name, pattern, child);
+    return new MakePatternNode(name, pattern, uses, child);
   }
 
   virtual void codegen(Decision *d, llvm::StringMap<llvm::Value *> substitution);
+  virtual void collectUses(std::set<std::string> &vars) { vars.insert(uses.begin(), uses.end()); child->collectUses(vars); }
+  virtual void collectDefs(std::set<std::string> &vars) { vars.insert(name); child->collectDefs(vars); }
```
bmmoore: Is `collectDef` supposed to collect everything bound by this node? I would think the `name` being matched already existed above, while the variables contained in `pattern` would be the ones that are bound.

include/kllvm/codegen/DecisionParser.h
```diff
@@ -5,8 +5,8 @@
 
 namespace kllvm {
 
-DecisionNode *parseYamlDecisionTreeFromString(std::string yaml, int numSubjects, const llvm::StringMap<KOREObjectSymbol *> &syms, const llvm::StringMap<KOREObjectCompositeSort *> &sorts);
-DecisionNode *parseYamlDecisionTree(std::string filename, int numSubjects, const llvm::StringMap<KOREObjectSymbol *> &syms, const llvm::StringMap<KOREObjectCompositeSort *> &sorts);
+DecisionNode *parseYamlDecisionTreeFromString(std::string yaml, const llvm::StringMap<KOREObjectSymbol *> &syms, const llvm::StringMap<KOREObjectCompositeSort *> &sorts);
```
bmmoore: Why was `numSubjects` removed? I think we will want to support trees that start with multiple variables available to bind.

lib/codegen/Decision.cpp
```diff
@@ -18,16 +20,54 @@ void Decision::operator()(DecisionNode *entry, llvm::StringMap<llvm::Value *> su
...
+  }
+  return vars;
+}
+
+bool DecisionNode::beginNode(Decision *d, std::string name, llvm::StringMap<llvm::Value *> &substitution) {
+  if (isCompleted()) {
+    llvm::BranchInst::Create(cachedCode, d->CurrentBlock);
+    for (std::string var : collectVars()) {
+      phis.lookup(var)->addIncoming(substitution[var], d->CurrentBlock);
+      substitution[var] = phis.lookup(var);
```
bmmoore: Is substitution used if we take the early return here?

include/kllvm/codegen/Decision.h
dwightguth: if cases.size() == 0 && !cases[0].getConstructor(), then we don't actually generate a switch statement, just an unconditional branch to another block, so the variable isn't actually used by the generated code.

include/kllvm/codegen/Decision.h
dwightguth: collectDef collects every variable that is introduced recursively by this node. MakePattern creates a new pattern, so name is a def, not a use.

include/kllvm/codegen/DecisionParser.h
dwightguth: it is no longer needed; we can now infer everything needed to generate the decision tree from the decision tree itself.

lib/codegen/Decision.cpp
dwightguth: I suppose not, but it doesn't hurt anything. I can remove it and see if the tests still pass if you want though?

include/kllvm/codegen/Decision.h
bmmoore: Ok, I see that `case x of y -> ...` can be handled just by setting up the substitution without any instructions that actually read.

But if `cases.size()` is 0, isn't `cases[0]` reading past the end of the list?

include/kllvm/codegen/Decision.h
bmmoore: Ah, I was confusing it with matching on a pattern.

lib/codegen/Decision.cpp
dwightguth: If the function returns true, the caller should also return. Which would then deallocate the stringmap. so we could delete this if you wanted. I just think it's nicer to keep it for consistency in case the interface changes later.

PR #43, opened 2018-10-30
Minor bugfixes to llvm backend for EVM and IELE

dwightguth:
Includes:
* rename the symbols associated with string token globals so that they do not have null bytes, which is apparently forbidden in llvm
* some bugfixes in how we pass collections around
* case for strategies in llvm and haskell code
* fix a bug in INT.bitRange hook where memory was being read uninitialized
* add lto for rust code
* fix a bug in computeMapElementScore that was causing a crash after we changed the Map sort to have keys which are KItem.

### DIFF
--- a/lib/ast/AST.cpp
+++ b/lib/ast/AST.cpp
@@ -286,9 +286,21 @@ KOREPattern *KOREAxiomDeclaration::getRightHandSide() const {
     if (top->getConstructor()->getName() == "\\implies" && top->getArguments().size() == 2) {
       if (auto andPattern = dynamic_cast<KOREObjectCompositePattern *>(top->getArguments()[1])) {
         if (andPattern->getConstructor()->getName() == "\\and" && andPattern->getArguments().size() == 2) {
-          if (auto equals = dynamic_cast<KOREObjectCompositePattern *>(andPattern->getArguments()[0])) {
-            if (equals->getConstructor()->getName() == "\\equals" && equals->getArguments().size() == 2) {
-              return equals->getArguments()[1];
+          if (auto bottomPattern = dynamic_cast<KOREObjectCompositePattern *>(top->getArguments()[0])) {
+            if (bottomPattern->getConstructor()->getName() == "\\bottom" && bottomPattern->getArguments().empty()) {
+              if (auto andPattern2 = dynamic_cast<KOREObjectCompositePattern *>(andPattern->getArguments()[1])) {
+                if (andPattern2->getConstructor()->getName() == "\\and" && andPattern2->getArguments().size() == 2) {
+                  if (auto rewrites = dynamic_cast<KOREObjectCompositePattern *>(andPattern2->getArguments()[1])) {
+                    if (rewrites->getConstructor()->getName() == "\\rewrites" && rewrites->getArguments().size() == 2) {
+                      return rewrites->getArguments()[1];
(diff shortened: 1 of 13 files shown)


lib/ast/AST.cpp
```diff
@@ -323,6 +335,19 @@ KOREPattern *KOREAxiomDeclaration::getRequires() const {
           return equals->getArguments()[0];
         } else if (equals->getConstructor()->getName() == "\\top" && equals->getArguments().empty()) {
           return nullptr;
+        } else if (equals->getConstructor()->getName() == "\\bottom" && equals->getArguments().empty()) {
+          // sttrategy axiom hack
```
bmmoore: spelling

runtime/arithmetic/int.cpp
```diff
@@ -230,6 +231,7 @@ mpz_ptr hook_INT_log2(mpz_t a) {
 void extract(mpz_t result, mpz_t i, size_t off, size_t len) {
   size_t size = (len + LIMB_BITS - 1) / LIMB_BITS;
   mpz_init2(result, len+LIMB_BITS);
+  memset(result->_mp_d, 0, size);
```
bmmoore: Why is this needed? mpz_init2 promises to set the number to zero. Is the problem from calling raw mpn functions later?

runtime/CMakeLists.txt
```diff
@@ -36,3 +36,7 @@ install(
   DESTINATION lib/kllvm/rust/deps
   PATTERN *.rlib
 )
+
+if (CMAKE_BUILD_TYPE STREQUAL "Release")
+  install(SCRIPT ../cmake/mkrlib.cmake)
```
bmmoore: I don't understand the CMake here. Is this adding additional build steps or trying to change dependencies? Is this the right way to do this?

runtime/arithmetic/int.cpp
dwightguth: Yeah pretty much. The issue is that if size > 0 and the final result of the operation is also zero, then the while loop at the end of this function will read uninitialized memory while trying to compute the appropriate value for _mp_size. properly speaking, we should probably be using mpz_limbs_write and mpz_limbs_finish, but this code also works fine (and I wrote it this way because I was aping the Zarith ocaml implementation)

runtime/CMakeLists.txt
dwightguth: This is adding a script that runs when the code is installed. That script proceeds to modify the rlib files installed from cargo so that they are suitable for lto. I did it this way because it's easier to get the list of all the libraries that need to be munged at install time after they've been created than trying to statically list all the recursive rust dependencies of our rust code.

dwightguth: It's also much more stable because if we depend on a rust package and that package adds new dependencies in a new version, we don't have to explicitly change the cmake script.

bmmoore: Ok, running this at installation time sounds pretty reliable. I was worried about interactions with compilation if it ran during the build.

runtime/arithmetic/int.cpp
bmmoore: Hmm, this stuff *is* documented, though some of it is not guaranteed to be stable across versions. I suppose it's ok.

PR #46, opened 2018-11-05
Topic/total ordering

pmeredit:
No ready for review, still needs llvm code updates, and a little bit for GMP (I think).


runtime/datastructures/src/decls.rs
```diff
@@ -56,6 +58,36 @@ impl PartialEq for KElem {
...
+        } else if ret == 0 {
+            Some(Ordering::Equal)
+        } else {
+            Some(Ordering::Greater)
+        }
+    }
+}
+
+impl Ord for KElem {
+    fn cmp(&self, other: &KElem) -> Ordering {
```
dwightguth: Do we actually need this function? The Eq trait is implemented by means of implicitly calling PartialEq and doesn't actually need to duplicate code. I'm willing to bet you can do the same thing with Ord and just delete this method

runtime/datastructures/src/decls.rs
```diff
@@ -69,6 +101,86 @@ impl Hash for KElem {
   }
 }
 
+fn hash_map_compare<K,V>(a: &HashMap<K, V>, b: &HashMap<K, V>) -> Ordering
```
dwightguth: This is not a good approach. You should modify the library to use OrdMap and OrdSet and then just make use of their Ord implementation

runtime/datastructures/src/decls.rs
```diff
@@ -69,6 +101,86 @@ impl Hash for KElem {
...
+        if b_keys.is_empty() {
+            return Ordering::Equal;
+        }
+        return Ordering::Less;
+    }
+    return Ordering::Greater;
+}
+
+fn hash_set_compare<K>(a: &HashSet<K>, b: &HashSet<K>) -> Ordering
+where K: Ord + Hash + Clone {
```
dwightguth: Same here, this can also be deleted

runtime/datastructures/src/decls.rs
```diff
@@ -69,6 +101,86 @@ impl Hash for KElem {
...
+        if b_vals.is_empty() {
+            return Ordering::Equal;
+        }
+        return Ordering::Less;
+    }
+    return Ordering::Greater;
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn hook_LIST_cmp(a: *const c_void, b: *const c_void) -> i64 {
```
dwightguth: Don't use c_void here, it's not actually necessary. Just use *const List. Then you need to transmute and it just becomes part of the abi of the call.

runtime/datastructures/src/decls.rs
```diff
@@ -69,6 +101,86 @@ impl Hash for KElem {
...
+  match std::mem::transmute::<*const c_void, &List>(a)
+      .cmp(std::mem::transmute::<*const c_void, &List>(b)) {
+    Ordering::Less => -1,
+    Ordering::Equal => 0,
+    Ordering::Greater => 1,
+  }
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn hook_MAP_cmp(a: *const c_void, b: *const c_void) -> i64 {
```
dwightguth: Same here

runtime/datastructures/src/decls.rs
```diff
@@ -69,6 +101,86 @@ impl Hash for KElem {
...
+  match hash_map_compare(std::mem::transmute::<*const c_void, &Map>(a),
+                         std::mem::transmute::<*const c_void, &Map>(b)) {
+    Ordering::Less => -1,
+    Ordering::Equal => 0,
+    Ordering::Greater => 1,
+  }
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn hook_SET_cmp(a: *const c_void, b: *const c_void) -> i64 {
```
dwightguth: And here

runtime/datastructures/src/decls.rs
```diff
@@ -98,6 +210,7 @@ extern "C" {
   pub fn move_int(result: *mut Int) -> *mut Int;
   pub fn printConfigurationInternal(file: *mut FILE, subject: *const Block, sort: *const c_char);
   pub fn hook_KEQUAL_eq(k1: K, k2: K) -> bool;
+  pub fn hook_KORD_cmp(k1: K, k2: K) -> i64;
```
dwightguth: You should still call it hook_KEQUAL_cmp. That's much more in alignment with how hooks are named in K and means we can eventually add it to the prelude once we get MInt working.

dwightguth:
A good start but the existing code needs cleanup.

runtime/datastructures/src/decls.rs
pmeredit: I tried removing it, it doesn't work. I think the issue is that PartialOrd returns an Option, while Ord does not. It's kind of annoying.  What I can do, however, is call cmp with unwrap to reduce code duplication.

runtime/datastructures/src/decls.rs
pmeredit: Interestingly, I figured this out independently.

runtime/datastructures/src/decls.rs
pmeredit: Done

runtime/datastructures/src/decls.rs
pmeredit: I assume you mean BTreeMap, which doesn't implement ORD. It only implements PartialOrd, because there is no total ordering on maps other than the one I forced here. So even if we use BTreeMap I would need to write nearly the same code (just no need for the BinaryHeap).

runtime/datastructures/src/decls.rs
pmeredit: Going to resolve this while leaving the other open, whatever we do will be applied to both :)

runtime/datastructures/src/decls.rs
pmeredit: Yea, this make more sense. The reason Eq requires no implementation is the only difference with PartialEQ is it's possible that there exists b such that b != b. By implementing Eq you are guaranteeing that isn't true.

With ParitalOrd you are dealing with different types, and partial orders are just much more common than PartialEQ, which is relegated primarily to NaN in IEEE floating point.

runtime/datastructures/src/decls.rs
pmeredit: Answered: im::OrdMap

PR #48, opened 2018-11-19
Fixes to llvm backend for iele

dwightguth:
Includes: 
* support for matching and compacting overloaded productions
* update to latest version of Array in K
* make llvm backend compatible with building a library
* code to free kore memory
* bug fixes


runtime/take_steps.ll
```diff
@@ -8,14 +8,20 @@ declare %block* @step(%block*)
 declare void @finish_rewriting(%block*) #0
 declare void @koreCollect(%block **)
```
pmeredit: Why are we writing these files directly in llvm instead of using C++ with C linkage?

dwightguth: Just the decision that was made at the time, no particular reason. We could convert them all at a later date, but that should be a separate PR.

dwightguth:
Jenkins: test this please

lib/codegen/CreateTerm.cpp
```diff
@@ -453,6 +453,7 @@ llvm::Value *CreateTerm::notInjectionCase(KOREObjectCompositePattern *constructo
   return new llvm::BitCastInst(Block, llvm::PointerType::getUnqual(Module->getTypeByName(BLOCK_STRUCT)), "", CurrentBlock);
 }
 
+// returns a value and a boolean indicating whetther that value could be an injection
```
pmeredit: whetther => whether

PR #49, opened 2018-12-03
Reduce size of decision tree

dwightguth:


dwightguth:
@raduom https://github.com/kframework/llvm-backend/wiki/Occurrences

dwightguth:
I mean it needs to be both easily serializable and easily convertible to a unique string representation. What exactly do you have in mind?

raduom:
I was looking at the comment you wrote and I though that you can pretty much write it in Haskell, with a couple of comments. I do not really like stringly types in a program, especially if written in Haskell. I would just use a normal Haskell data structure with a Show instance.

raduom:
Yeah. The recursive structure is clear and the code also looks more intuitive. Some things I would have done differently, and probably will later:
* I would have use a record since matching positional arguments can be somewhat confusing for large data structures.
* Recursive data structures can be written in fix point notation to be able to abstract over recursion.

However, all this can be done later.

PR #52, opened 2018-12-12
Improvements to rust allocation

dwightguth:
Fixes a bug where we were still using malloc to allocate rust objects. Also fixes an issue where code generation generated a call to 0 for the __rust_alloc function.

### DIFF
--- a/runtime/datastructures/src/decls.rs
+++ b/runtime/datastructures/src/decls.rs
@@ -14,9 +14,17 @@ pub struct KoreAllocator;
 unsafe impl GlobalAlloc for KoreAllocator {
   #[inline(always)]
   unsafe fn alloc(&self, _layout: Layout) -> *mut u8 {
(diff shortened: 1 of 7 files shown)


runtime/datastructures/src/decls.rs
```diff
@@ -14,9 +14,17 @@ pub struct KoreAllocator;
...
-    koreAllocOld(_layout.size())
+    if during_gc() {
+      malloc(_layout.size())
+    } else {
+      koreAllocOld(_layout.size())
+    }
+  }
+  unsafe fn dealloc(&self, _ptr: *mut u8, _layout: Layout) {
+    if during_gc() {
+      free(_ptr);
```
pmeredit: Couldn't this result in freeing koreAlloc'd memory?

pmeredit:
lgtm

PR #54, opened 2018-12-13
install instructions

dwightguth:


PhilMiller:
You should also mention use of the flag `--backend llvm` when compiling a definition, and the common variable in at least the examples in `k/k-distribution` expecting that to be set in `KOMPILE_FLAGS`

dwightguth:
These are the install instructions for the llvm backend. How to use the backend with the k frontend is a good topic of documentation but I don't think it belongs here in particular.

dwightguth:
It's a dependency of llvm's static libraries.

bmmoore:
Shouldn't it be a dependency of the llvm-6.0 package then?

bmmoore:
Ah, `-l tinfo` seems to work as well as `-l ncurses`, and libtinfo is among the dependencies of the llvm package.

PR #56, opened 2018-12-18
Update build system to work when built by k

dwightguth:


dwightguth:
I mean those are both variables set by CMAKE implicitly.

PR #65, opened 2019-01-11
Floating point types

dwightguth:
Includes hooks and compiler integration, but not dependent typing (will do later)

Note that not all the tests I made live in this repo, I also added some tests to the K repo now that everything is set up in K to have tests for the llvm backend there. That will hopefully make it easier for the Haskell backend to eventually share tests.


include/runtime/header.h
```diff
@@ -75,6 +75,11 @@ extern "C" {
     void *f;
     char *g;
   } list;
+
+  typedef struct floating {
+    uint64_t exp;
```
bmmoore: What is the meaning of this field? Is it the mpfr "exponent range"? An undocumented field named "exp" sounds like the value of the exponent.

lib/codegen/CreateTerm.cpp
```diff
@@ -230,7 +230,51 @@ llvm::Value *CreateTerm::createToken(ValueType sort, std::string contents) {
...
+      std::vector<llvm::Constant *> allocdLimbs;
+      for (size_t i = 0; i < size; i++) {
+        allocdLimbs.push_back(llvm::ConstantInt::get(llvm::Type::getInt64Ty(Ctx), value->_mpfr_d[i]));
+      }
+      limbsVar->setInitializer(llvm::ConstantArray::get(limbsType, allocdLimbs));
+      llvm::Constant *expbits = llvm::ConstantInt::get(llvm::Type::getInt64Ty(Ctx), exp);
+      llvm::Constant *mpfr_prec = llvm::ConstantInt::get(llvm::Type::getInt64Ty(Ctx), prec);
+      llvm::Constant *mpfr_sign = llvm::ConstantInt::getSigned(llvm::Type::getInt32Ty(Ctx), value->_mpfr_sign);
+      llvm::Constant *mpfr_exp = llvm::ConstantInt::getSigned(llvm::Type::getInt64Ty(Ctx), value->_mpfr_exp);
+      globalVar->setInitializer(llvm::ConstantStruct::get(Module->getTypeByName(FLOAT_STRUCT), expbits, llvm::ConstantStruct::getAnon({mpfr_prec, mpfr_sign, mpfr_exp, llvm::ConstantExpr::getPointerCast(limbsVar, llvm::Type::getInt64PtrTy(Ctx))})));
```
bmmoore: This looks like you are setting up a struct literal. Is this method of construction documented in the MPFR documentation, or are we making some assumptions here? Either way, this branch of the case should be documented.

ciscript
```diff
@@ -47,4 +47,5 @@ llvm-kompile ../test/test9.kore TEST main -o interpreter
 ./interpreter ../test/test9.kore -1 /dev/stdout | diff - ../test/test9.out
 llvm-kompile ../test/test10.kore TEST main -o interpreter
 ./interpreter ../test/test10.in.kore -1 /dev/stdout | diff - ../test/test10.out
+llvm-kompile ../test/test11.kore TEST main -o interpreter
```
bmmoore: What about running the generated interpreter?

runtime/strings/strings.cpp
```diff
@@ -182,12 +184,15 @@ extern "C" {
     return hook_STRING_string2base_long(input, ubase);
   }
 
-  const string * hook_STRING_float2string(const mpfr_t input) {
-    throw std::logic_error("TODO: floats");
+  const string * hook_STRING_float2string(const floating *input) {
```
bmmoore: Are these new hooks tested anywhere?

lib/codegen/CreateTerm.cpp
dwightguth: It's documented in the mpfr internals and isn't going to change because the C++ ABI depends on it. We are literally just copying an mpfr_t from the host to the target though. It assumes we're not cross compiling and that's it. I will add a comment here though.

ciscript
dwightguth: This test just tests that the definition compiles. I wrote the more detailed tests to be part of the k frontend and I believe we should pursue this path moving forward as the testing infrastructure there will be more robust and stable.

runtime/strings/strings.cpp
dwightguth: See comment above re: k frontend tests.

dwightguth:
The build succeeded but failed to update status on github, so I will merge this with admin powers.

PR #68, opened 2019-01-14
Updates to make build on Arch Linux

ehildenb:
These are the changes I made to make the LLVM backend build properly on Arch linux (including updated build instructions).

Note that I don't think this can be merged as-is, because Ubuntu still needs LLVM 6.0, and this updates to LLVM 7.0. @dwightguth do you know how to set the version of LLVM based on the distro, or how to make CMakeLists.txt not care about the version?


dwightguth:
Closing as this has been superceded by #80

PR #69, opened 2019-01-15
Moved more map code.

raduom:


raduom:
I got stuck a bit on the scoring functions, and I am thinking of dedicating a refactoring session only for that. I think it would be a pretty big win isolating this code.

dwightguth:
Being able to separate out the heuristics will be very helpful moving forward, yes. I think it should be possible to factor separately the scoring code that decides which map/set rows have the same key as the current row and the scoring code that decides how to actually score things. I think this is probably what you should aim for next.

PR #79, opened 2019-02-05
add proper hex escape

dwightguth:


lib/parser/KOREScanner.l
```diff
@@ -15,6 +15,20 @@ using namespace kllvm::parser;
...
+    -1,-1, 0,1,2,3,4,5,6,7,8,9,-1,-1,-1,-1,-1,-1,-1,10,11,12,13,14,15,-1,
+    -1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,
+    -1,-1,10,11,12,13,14,15,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,
+    -1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,
+    -1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,
+    -1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,
+    -1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,
+    -1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,
+    -1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1
+};
```
bmmoore: 1. Why have a table instead of using `strtol`? If it's intended to improve performance I would like to see benchmarks.
2. The formatting is horribly misleading - most commas are aligned, but the third line has 26 entries while the other lines have 23! If we must have a table, at least format it in a regularly with all lines all having 16 entries (or similar) so it's easier to visually inspect for correctness.

lib/parser/KOREScanner.l
```diff
@@ -96,6 +96,7 @@ yyin = in;
 <STR>"\\\""     { stringBuffer.push_back('\"');               }
 <STR>"\\\\"     { stringBuffer.push_back('\\');               }
 <STR>\\[0-9]{3} { stringBuffer.push_back((yytext[1] - '0') * 64 + (yytext[2] - '0') * 8 + yytext[3] - '0'); }
+<STR>\\x[0-9a-fA-F]{2} { stringBuffer.push_back(strtol(yytext[2], NULL, 16)); }
```
bmmoore: I think that should be `yytext+2` or maybe `&yytext[2]`.

PR #80, opened 2019-02-07
Build system refactoring

dwightguth:
Ready for review. @raduom , please test on Mac OS.


raduom:
> Ready for review. @raduom , please test on Mac OS.

I had to add

```
#define __MATH_LONG_DOUBLE_CONSTANTS
```

in floattest.cpp

dwightguth:
I will make a separate pull request to get Mac OS working, then. @bmmoore or @theo25 , can you review?

dwightguth:
That's a very good catch! We weren't catching it during testing because the kframework binary package depends on opam which recommends ocaml which depends on libtinfo-dev. I updated the install instructions and will merge when the tests pass, then I will try to fix the cause of the bug in the testing infrastructure.

PR #82, opened 2019-02-13
GC refactoring

theo25:
This is ready for review.

This is a refactoring of the garbage collection code to make it parametric in the allocation arena. I have verified that it does not increase the assembly size of the allocator/collector, in fact the size is reduced for about 140 instructions.

The PR also includes a patch to the build system that I needed to make the project build against an llvm build built from source.


include/runtime/arena.h
```diff
@@ -0,0 +1,79 @@
...
+// the space is allocated in a general (not garbage collected pool).
+void *arenaAlloc(struct arena *, size_t);
+
+// Resizes the last allocation as long as the resize does not require a new
+// block allocation.
+// Returns the address of the byte following the last newlly allocated byte when
+// the resize succeeds, returns 0 otherwise.
+void *arenaResizeLastAlloc(struct arena *, ssize_t);
+
+// Exchenges the current allocation and collection semispaces and clears the new
```
dwightguth: s/Exchenges/Exchanges/

include/runtime/arena.h
```diff
@@ -0,0 +1,79 @@
...
+// Returns the address of the first byte that belongs in the given arena.
+// Returns 0 if nothing has been allocated ever in that arena.
+char *arenaStartPtr(const struct arena *);
+
+// Returns a pointer to the address of the byte following the last allocated
+// byte in the given arena.
+// This address is 0 if nothing has been allocated ever in that arena.
+char **arenaEndPtr(struct arena *);
+
+// Returns a pointer to the byte following the given number of bytes starting
```
dwightguth: this comment is not really clear to me what the function is doing, you should probably clarify.

runtime/alloc/alloc.c
```diff
@@ -3,177 +3,72 @@
...
 
-static struct semispace_info fromspace;
-static struct semispace_info oldspace = {0, 0, 0, 0, 2};
-
-static char* first_tospace_block;
-
-char *fromspace_ptr() {
-  return fromspace.first_block;
+char *youngspace_ptr() {
+  return arenaStartPtr(&youngspace);
```
dwightguth: have you checked that the assembly for these four functions do not contain a call instruction and that it gets successfully inlined?

runtime/alloc/alloc.c
```diff
@@ -3,177 +3,72 @@
...
-  MEM_LOG("Allocation at %p (size %zd), next alloc at %p (if it fits)\n", result, requested, space->block);
-  return result;
+  freeAllMemory();
+  arenaReset(&youngspace);
+  arenaReset(&oldspace);
 }
 
 __attribute__ ((always_inline)) void* koreAlloc(size_t requested) {
-  return doAlloc(requested, &fromspace);
+  return arenaAlloc(&youngspace, requested);
```
dwightguth: similarly here, is arenaAlloc getting inlined?

include/runtime/arena.h
theo25: fixed

include/runtime/arena.h
theo25: hopefully fixed

runtime/alloc/alloc.c
theo25: Yes I can verify that these are getting inlined properly. I checked the assembly code of the evm interpreter to make 100% sure.

runtime/alloc/alloc.c
theo25: Same here. In fact, `doAlloc` is also getting inlined. There only calls to `doAllocSlow` in the binary.

runtime/main/main.ll
```diff
@@ -29,7 +29,7 @@ entry:
   %output_str = load i8*, i8** %output_ptr
   store i8* %output_str, i8** @output_file
 
-  call void @__gmp_set_memory_functions(i8* (i64) * @koreAllocOld, i8* (i8*, i64, i64) * @koreReallocOld, void (i8*, i64) * @koreFree)
+  call void @__gmp_set_memory_functions(i8* (i64) * @koreAllocNoGC, i8* (i8*, i64, i64) * @koreReallocNoGC, void (i8*, i64) * @koreFree)
```
dwightguth: Can you please refactor this out into a separate method? Right now it is duplicated with the main method of the blockchain vm server and I'd like to centralize the code so that future changes will not require updating both repos.

theo25:
I have added the third arena for not collected objects. Currently testing against evm.

include/runtime/arena.h
dwightguth: You clarified the wrong comment

runtime/main/main.ll
theo25: done

theo25:
I tested the PR against evm's vm and blockchain tests. All tests pass except
`tests/ethereum-tests/BlockchainTests/GeneralStateTests/stSpecialTest/JUMPDEST_AttackwithJump_d0g0v0.json`
which requires an amount of memory that my development laptop cannot provide.

PR #83, opened 2019-02-15
Substitution hook

dwightguth:
Implemented: capture-avoiding substitution and renaming of variables during printing phase.


runtime/strings/strings.cpp
```diff
@@ -47,6 +47,14 @@ extern "C" {
   }
 
   bool hook_STRING_eq(const string * a, const string * b) {
+    uintptr_t aint = (uintptr_t) a;
+    uintptr_t bint = (uintptr_t) b;
+    if ((aint & 3) == 3 || (bint & 3) == 3) {
```
theo25: Not claiming that there is anything wrong here (or in other olaces with a similar check), but could you explain how/where you enforce the invariants about the leats significant bits of the block pointers?

dwightguth: if the least significant bit is a 1, then the other 63 bits are reserved for various constant values that would otherwise require an allocation. In this case, if the second least significant bit is also a 1, then it is a bound variable and the most significant 32 bits contains its debruijn index, otherwise it is an object-level symbol with 0 arguments and the most significanat 32 bits contains its tag. Is there a particular place in the code you think I should put this info that you would look for it there?

theo25: Maybe as a comment in the `include/runtime/header.h`?

theo25: I saw the commit, thanks for adding the clarification

runtime/equality.ll
```diff
@@ -127,8 +128,15 @@ compareSymbol:
   %child2ptr = load %block*, %block** %child2ptrptr
   %comparedSymbol = call i1 @hook_KEQUAL_eq(%block* %child1ptr, %block* %child2ptr)
   br i1 %comparedSymbol, label %loop, label %exit
+compareVariable:
+  %var1ptrptr = inttoptr i64 %child1intptr to %block**
+  %var2ptrptr = inttoptr i64 %child2intptr to %block**
+  %var1ptr = load %block*, %block** %var1ptrptr
+  %var2ptr = load %block*, %block** %var2ptrptr
+  %comparedVar = call i1 @hook_STRING_eq(%block* %var1ptr, %block* %var2ptr)
```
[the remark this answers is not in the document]
dwightguth: well, the hash function hashes the entire 64-bit address if the least significant bit is one, so the only thing you're missing out on is a different hash value for different free variables with the VARIABLE_BIT set that happen to also share the same name. It should be sound. I didn't really consider it earlier, I was just taking the easier path, but I really think this is the best we can do. If we hash based on the address, then when gc moves an object it will change hash buckets, which will make things break. So I'm not sure there's any room for improvement here.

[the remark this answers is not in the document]
bmmoore: It doesn't look like this comparison is considering addresses either, unless `hook_STRING_eq` is badly named.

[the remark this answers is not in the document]
dwightguth: hook_STRING_eq compares two terms of type string *. If you look at the change I made to that function, you will see that it is comparing addresses iff the VARIABLE_BIT is set.

PR #85, opened 2019-02-17
Pull all current changes from KORE.

raduom:


matching/src/KorePatterns.hs
```diff
@@ -1,51 +1,74 @@
...
-pattern App_ fnSymbol args =
+       ( Object (..), Meta (..) )
+import Kore.Sort
+       ( Sort (..) )
+
+pattern KoreObjectPattern :: Functor domain
+                          => Pattern Object domain variable
+                              (KorePattern domain variable annotation)
+                          -> KorePattern domain variable annotation
+pattern KoreObjectPattern pat <- (Recursive.project -> _ :< UnifiedObjectPattern pat)
```
bmmoore: A bit unfortunate that this is a unidirectional pattern, but I don't see any good way to invent an annotation.

PR #97, opened 2019-02-28
Garbage collection for the old generation

theo25:
This PR implements a garbage collection for the old generation. It uses the same copy-collector infrastructure for the old generation. For now, the old generation is collected every 10 young generation collections. We can change this in future by adding desirable logic in the `shouldCollectOldGen` function. It is tested with the EVM vm and blockchain tests.

### DIFF
--- a/include/runtime/alloc.h
+++ b/include/runtime/alloc.h
@@ -22,7 +22,8 @@ void* koreAllocTokenOld(size_t requested);
 // allocates exactly requested bytes into the not garbage-collected arena
 void* koreAllocNoGC(size_t requested);
 // swaps the two semispace of the young generation as part of garbage collection
-void koreAllocSwap(void);
+// if the swapOld flag is set, it also swaps the two semispaces of the old generation
+void koreAllocSwap(bool swapOld);
 // resizes the last allocation into the young generation
 void* koreResizeLastAlloc(void* oldptr, size_t newrequest, size_t oldrequest);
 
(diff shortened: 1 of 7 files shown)


dwightguth:
Once every ten collections might be too often. Have you measured the before/after of the gc overhead?

theo25:
No I haven't. Can you suggest which test should I use as a benchmark for this kind of measurement?

theo25:
These are the time measurements for the blockchain tests, done on my laptop (Intel Core i7-6500U CPU @ 2.50GHz, 12GB memory).

master branch:
```
real 23m54.506s
user 12m39.724s
sys   3m52.104s
```

gc-old branch:
```
real 32m1.973s
user 14m19.712s
sys   4m18.388s
```

I skipped `tests/ethereum-tests/BlockchainTests/GeneralStateTests/stSpecialTest/JUMPDEST_AttackwithJump_d0g0v0.json` because it thrashes my memory.

dwightguth:
I'm suspicious of these timing numbers because one of them shows a much higher wall time but no corresponding increase in user or system time. Are you sure you didn't have anything running in the background when you ran the gc-old branch?

theo25:
I repeated the measurements after setting the frequency of collection for the old generation to once every 50 young generation collections (from once every 10). Here are the new resutls:

master branch:

real 23m36.456s
user 12m41.164s
sys   3m52.752s

gc-old branch:

real 25m50.217s
user 14m24.944s
sys   4m4.368s

This looks like a significant improvement over the previous timings, although still it introduces a slowdown as was expected

dwightguth:
25 minutes instead of 23 is probably acceptable. Hopefully when the peak memory usage goes down we will see a significant drop in system time which will make up for the increased overhead.

PR #98, opened 2019-03-01
IO hooks

### LINKED ISSUE #91: Implement I/O
Again, we have two paths here, as I/O is implemented via a series of impure function hooks.

* we can do the unprincipled thing and implement the hooks directly, which would take ~1 week but would potentially not work with the haskell backend
* we could do the principled thing and do something similar to the I/O Monad in haskell where an implicit "state" object is passed to these functions. This could take 3-4 weeks and could potentially disrupt any semantics currently using I/O, including those that use the java backend.

mickyabir:
IO hooks to resolve #91


runtime/io/io.cpp
```diff
@@ -0,0 +1,178 @@
...
+#include "runtime/header.h"
+
+extern "C" {
+
+#define KCHAR char
+  
+  mpz_ptr move_int(mpz_t);
+
+  char * getTerminatedString(string * str) {
+    char * buf = (char *) malloc(sizeof(KCHAR) * (len(str) + 1));
```
dwightguth: can we refactor out the call to len so it only happens once?

runtime/io/io.cpp
```diff
@@ -0,0 +1,178 @@
...
+          flags = O_RDWR | O_TRUNC | O_CREAT;
+        }
+      case 'a':
+        flags = O_WRONLY | O_APPEND | O_CREAT;
+        if (len(control) > 1 && *(++ctl) == '+'){
+          flags = O_RDWR | O_APPEND | O_CREAT;
+        }
+      default:
+        flags = O_RDONLY;
+    }
```
dwightguth: can we also support b, e, and x mode flags?

runtime/io/io.cpp
```diff
@@ -0,0 +1,178 @@
...
+    if (!mpz_fits_sint_p(i)) {
+      throw std::invalid_argument("Arg too large for int32_t");
+    }
+
+    int fd = mpz_get_si(i);
+
+    off_t loc = lseek(fd, 0, SEEK_CUR);
+
+    mpz_t result;
+    mpz_init_set_si(result, int(loc));
```
dwightguth: off_t is actually 64-bits, so this should be a cast to long, not int.

runtime/io/io.cpp
```diff
@@ -0,0 +1,178 @@
...
+    return move_int(result);
+  }
+
+  string * hook_IO_read(mpz_t i, mpz_t len) {
+    if (!mpz_fits_sint_p(i) || !mpz_fits_ulong_p(len)) {
+      throw std::invalid_argument("Arg too large");
+    }
+
+    int fd = mpz_get_si(i);
+    int length = mpz_get_si(len);
```
dwightguth: this should be mpz_get_ui.

runtime/io/io.cpp
```diff
@@ -0,0 +1,178 @@
...
+
+    char * buf = new char[length];
+    int bytes = read(fd, buf, length);
```
dwightguth: I don't think we need this double copy. I'm fairly sure we can just alloc `length` bytes via koreAllocToken and then realloc down to `bytes` bytes after `read`ing into the kore heap.

runtime/io/io.cpp
```diff
@@ -0,0 +1,178 @@
...
+    memcpy(&(result->data), buf, bytes * sizeof(KCHAR));
+
+    delete[] buf;
+
+    return result;
+  }
+
+  void hook_IO_close(mpz_t i) {
+    if (!mpz_fits_sint_p(i)) {
+      throw std::invalid_argument("Arg too large for int64_t");
```
dwightguth: this should be int, not int64_t. same with the other uses of int32_t to refer to a fd.

runtime/io/io.cpp
```diff
@@ -0,0 +1,178 @@
...
+    auto result = static_cast<string *>(koreAllocToken(sizeof(string) + bytes));
+    set_len(result, bytes);
+    memcpy(&(result->data), buf, bytes * sizeof(KCHAR));
+
+    delete[] buf;
+
+    return result;
+  }
+
+  void hook_IO_close(mpz_t i) {
```
dwightguth: the functions that return void should not have hooks of void type. They should return a block * equal to a statically allocated term with 0 children whose block header is gotten by calling `getBlockHeaderForSymbol(getTagForSymbolName("dotk"))`. given that it is statically allocated, its NOT_YOUNG_OBJECT_BIT should be set.

runtime/io/io.cpp
```diff
@@ -0,0 +1,178 @@
...
+    if (!mpz_fits_sint_p(i)) {
+      throw std::invalid_argument("Arg too large for int64_t");
+    }
+
+    int fd = mpz_get_si(i);
+    close(fd);
+  }
+
+  void hook_IO_seek(mpz_t i, mpz_t loc) {
+    if (!mpz_fits_sint_p(i) || !mpz_fits_sint_p(i)) {
```
dwightguth: should be i and loc, not i and i. Also, loc should be a slong.

runtime/io/io.cpp
```diff
@@ -0,0 +1,178 @@
...
+      throw std::invalid_argument("Arg too large");
+    }
+
+    int fd = mpz_get_si(i);
+    int l = mpz_get_si(loc);
+    lseek(fd, l, SEEK_SET);
+  }
+
+  void hook_IO_seekEnd(mpz_t i, mpz_t loc) {
+    if (!mpz_fits_sint_p(i) || !mpz_fits_sint_p(i)) {
```
dwightguth: same here, loc should be a slong.

runtime/io/io.cpp
```diff
@@ -0,0 +1,178 @@
...
+    write(fd, &ch, 1);
+  }
+
+  void hook_IO_write(mpz_t i, string * str) {
+    if (!mpz_fits_sint_p(i)) {
+      throw std::invalid_argument("Arg too large for int32_t");
+    }
+
+    int fd = mpz_get_si(i);
+    char * s = getTerminatedString(str);
```
dwightguth: this call is unnecessary, as write does not require the string to be null terminated. you can just use str->data to access the char * from the string *.

runtime/io/io.cpp
```diff
@@ -0,0 +1,178 @@
...
+  char * getTerminatedString(string * str) {
+    char * buf = (char *) malloc(sizeof(KCHAR) * (len(str) + 1));
+    strncpy(buf, str->data, len(str));
+    buf[len(str)] = '\0';
+    return buf;
+  }
+  
+  mpz_ptr hook_IO_open(string * filename, string * control) {
+    int flags = 0;
+    char * ctl = getTerminatedString(control);
```
dwightguth: I don't think you need this call, as you aren't actually doing anything other than examining the characters of the string.

runtime/io/io.cpp
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    int fd = mpz_get_si(i);
+    char * s = getTerminatedString(str);
+
+    write(fd, s, len(str));
+  }
+
+  void hook_IO_lock(mpz_t i, mpz_t len) {
+    if (!mpz_fits_sint_p(i) || !mpz_fits_sint_p(len)) {
```
dwightguth: slong for len.

runtime/io/io.cpp
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    int fd = mpz_get_si(i);
+    int l = mpz_get_si(len);
+
+    lockf(fd, F_LOCK, l);
+  }
+
+  void hook_IO_unlock(mpz_t i, mpz_t len) {
+    if (!mpz_fits_sint_p(i) || !mpz_fits_sint_p(len)) {
```
dwightguth: slong for len

runtime/io/io.cpp
dwightguth: this is my bad. I forgot how 0-child symbols work. it should actually be `(((uint64_t)getTagForSymbol("dotk")) << 32) | 1`

runtime/io/io.cpp
```diff
@@ -0,0 +1,389 @@
...
+
+  mpz_ptr move_int(mpz_t);
+
+  static block dotK = {getBlockHeaderForSymbol((((uint64_t)getTagForSymbolName("dotk")) << 32) | 1), {}};
+  static blockheader kseqHeader = {getBlockHeaderForSymbol((uint64_t)getTagForSymbolName("kseq{}"))};
+
+  block * block_errno() {
+    block * retBlock = static_cast<block *>(koreAlloc(sizeof(block)));
+    switch (errno) {
+    case EOF:
```
dwightguth: create a local char * and assign ERRTAG(EOF) to it, then do the rest after the switch statement. Also, don't forget to break. Finally, put the case and the statement and the break on one line to make it easier to read.

runtime/io/io.cpp
```diff
@@ -0,0 +1,389 @@
...
+      retBlock->h = getBlockHeaderForSymbol((((uint64_t)getTagForSymbolName(ERRTAG(EOF))) << 32) | 1);
```
dwightguth: these are all 0 argument symbols so you shouldn't be calling getBlockHeaderForSymbol, you should just take the ored value and cast it to a pointer.

runtime/io/io.cpp
```diff
@@ -0,0 +1,389 @@
...
+#define KCHAR char
+#define ERRTAG(err) "Lbl'hash'" #err
+
+  mpz_ptr move_int(mpz_t);
+
+  static block dotK = {getBlockHeaderForSymbol((((uint64_t)getTagForSymbolName("dotk")) << 32) | 1), {}};
+  static blockheader kseqHeader = {getBlockHeaderForSymbol((uint64_t)getTagForSymbolName("kseq{}"))};
+
+  block * block_errno() {
+    block * retBlock = static_cast<block *>(koreAlloc(sizeof(block)));
```
dwightguth: you don't need this allocation because only the default case actually has to allocate a block.

runtime/io/io.cpp
```diff
@@ -0,0 +1,389 @@
...
+    case EHOSTDOWN:
+      retBlock->h = getBlockHeaderForSymbol((((uint64_t)getTagForSymbolName(ERRTAG(EHOSTDOWN))) << 32) | 1);
+    case EHOSTUNREACH:
+      retBlock->h = getBlockHeaderForSymbol((((uint64_t)getTagForSymbolName(ERRTAG(EHOSTUNREACH))) << 32) | 1);
+    case ELOOP:
+      retBlock->h = getBlockHeaderForSymbol((((uint64_t)getTagForSymbolName(ERRTAG(ELOOP))) << 32) | 1);
+    case EOVERFLOW:
+      retBlock->h = getBlockHeaderForSymbol((((uint64_t)getTagForSymbolName(ERRTAG(EOVERFLOW))) << 32) | 1);
+    default:
+      retBlock = static_cast<block *>(koreResizeLastAlloc(retBlock, sizeof(block) + sizeof(uint64_t), sizeof(block)));
```
dwightguth: this is the only case that needs to allocate so you don't need to call resize.

runtime/io/io.cpp
```diff
@@ -0,0 +1,389 @@
...
+      retBlock->h = getBlockHeaderForSymbol((uint64_t)getTagForSymbolName("Lbl'Hash'unkownIOError"));
```
dwightguth: s/unkown/unknown/

runtime/io/io.cpp
```diff
@@ -0,0 +1,389 @@
...
+      mpz_t err;
+      mpz_init_set_si(err, errno);
+      mpz_ptr p = move_int(err);
+      memcpy(&(retBlock->children[0]), &p, sizeof(uint64_t));
```
dwightguth: I would just say memcpy(retBlock->children, &p, sizeof(mpz_ptr))

runtime/io/io.cpp
```diff
@@ -0,0 +1,389 @@
...
+    }
+
+    return retBlock;
+  }
+
+  blockheader header_int() {
+    static blockheader header = {(uint64_t)-1};
+
+    if (header.hdr == -1) {
+      getBlockHeaderForSymbol((uint64_t)getTagForSymbolName("inj{SortInt, SortIOInt}"));
```
dwightguth: you're not assigning to header here.

runtime/io/io.cpp
```diff
@@ -0,0 +1,389 @@
...
+    }
+
+    return header;
+  }
+
+  blockheader header_err() {
+    static blockheader header = {(uint64_t)-1};
+
+    if (header.hdr == -1) {
+      getBlockHeaderForSymbol((uint64_t)getTagForSymbolName("inj{SortIOError, SortKItem}"));
```
dwightguth: same problem.

runtime/io/io.cpp
```diff
@@ -0,0 +1,389 @@
...
+    }
+
+    return header;
+  }
+
+  blockheader header_string() {
+    static blockheader header = {(uint64_t)-1};
+
+    if (header.hdr == -1) {
+      getBlockHeaderForSymbol((uint64_t)getTagForSymbolName("inj{SortString, SortIOString}"));
```
dwightguth: and here

runtime/io/io.cpp
```diff
@@ -0,0 +1,389 @@
...
+    if ((length > 1 && control->data[1] == '+') || (length > 2 && control->data[2] == '+')) {
+      flags &= (~O_WRONLY) & ~(O_RDONLY);
+      flags |= O_RDWR;
+    }
+
+    mpz_t result;
+    char * f = getTerminatedString(filename);
+    int fd = open(f, flags, S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP | S_IROTH);
+
+    block * retBlock = static_cast<block *>(koreAlloc(sizeof(block) + sizeof(uint64_t)));
```
dwightguth: I don't think it's going to be acceptable to have allocation function calls in between the call to open and reading the value of errno if open failed. koreAlloc can in some cases call malloc which presumably can call all kinds of system calls that reset errno.

runtime/io/io.cpp
```diff
@@ -0,0 +1,389 @@
...
+
+    if (-1 == fd) {
+      retBlock->h = header_err();
+      block * p = block_errno();
+      memcpy(&(retBlock->children[0]), &p, sizeof(uint64_t));
```
dwightguth: same, you probably want memcpy(retBlock->children, &p, sizeof(block *))

runtime/io/io.cpp
```diff
@@ -0,0 +1,389 @@
...
+
+    } else {
+      retBlock->h = header_int();
+      mpz_init_set_si(result, fd);
+      mpz_ptr p = move_int(result);
+      memcpy(&(retBlock->children[0]), &p, sizeof(uint64_t));
```
dwightguth: and here

runtime/io/io.cpp
```diff
@@ -0,0 +1,389 @@
...
+    if (header.hdr == -1) {
+      getBlockHeaderForSymbol((uint64_t)getTagForSymbolName("inj{SortString, SortIOString}"));
+    }
+
+    return header;
+  }
+
+  char * getTerminatedString(string * str) {
+    int length = len(str);
+    char * buf = (char *) malloc(sizeof(KCHAR) * (length + 1));
```
dwightguth: can we make this a koreAllocToken instead? then you don't need to explicitly free and you aren't paying the overhead of a malloc. You will have to call set_len though.

runtime/io/io.cpp
```diff
@@ -0,0 +1,389 @@
...
+  block * hook_IO_close(mpz_t i) {
+    if (!mpz_fits_sint_p(i)) {
+      throw std::invalid_argument("Arg too large for int32_t");
+    }
+
+    int fd = mpz_get_si(i);
+    int ret = close(fd);
+
+    if (ret == -1) {
+      block * retBlock = static_cast<block *>(koreAlloc(sizeof(block) + 2 * sizeof(uint64_t)));
```
dwightguth: same problem here, you can't call koreAlloc between close and block_errno.

runtime/io/io.cpp
```diff
@@ -0,0 +1,389 @@
...
+      retBlock->h = kseqHeader;
+      block * inj = static_cast<block *>(koreAlloc(sizeof(block) + sizeof(uint64_t)));
+      inj->h = header_err();
+      block * err = block_errno();
```
dwightguth: please move this up to the top of the if.

runtime/io/io.cpp
```diff
@@ -0,0 +1,389 @@
...
+#include "runtime/header.h"
+
+extern "C" {
+
+#define KCHAR char
+#define ERRTAG(err) "Lbl'hash'" #err
+
+  mpz_ptr move_int(mpz_t);
+
+  static block dotK = {getBlockHeaderForSymbol((((uint64_t)getTagForSymbolName("dotk")) << 32) | 1), {}};
```
dwightguth: this should not be a block, this should be a block * and should also just cast the ored value to a pointer rrather than calling getBlockHeaderForSymbol.

runtime/io/io.cpp
```diff
@@ -0,0 +1,389 @@
...
+      block * inj = static_cast<block *>(koreAlloc(sizeof(block) + sizeof(uint64_t)));
+      inj->h = header_err();
+      block * err = block_errno();
+      memcpy(&(inj->children[0]), &err, sizeof(uint64_t));
+      memcpy(&(retBlock->children[0]), &inj, sizeof(uint64_t));
+      memcpy(&(retBlock->children[1]), &dotK, sizeof(uint64_t));
+      return retBlock;
+    }
+
+    return &dotK;
```
dwightguth: remove the & once you change the type of dotk.

runtime/io/io.cpp
```diff
@@ -0,0 +1,389 @@
...
+    int fd = mpz_get_si(i);
+    int ret = close(fd);
+
+    if (ret == -1) {
+      block * retBlock = static_cast<block *>(koreAlloc(sizeof(block) + 2 * sizeof(uint64_t)));
+      retBlock->h = kseqHeader;
+      block * inj = static_cast<block *>(koreAlloc(sizeof(block) + sizeof(uint64_t)));
+      inj->h = header_err();
+      block * err = block_errno();
+      memcpy(&(inj->children[0]), &err, sizeof(uint64_t));
```
dwightguth: see my comments about the other memcpys for these three lines.

mickyabir:
Unit tests are not updated in the latest commits to deal with blocks

runtime/io/io.cpp
```diff
@@ -0,0 +1,440 @@
...
+    case ESHUTDOWN: errStr = ERRTAG(ESHUTDOWN); break;
+    case ETOOMANYREFS: errStr = ERRTAG(ETOOMANYREFS); break;
+    case ETIMEDOUT: errStr = ERRTAG(ETIMEDOUT); break;
+    case ECONNREFUSED: errStr = ERRTAG(ECONNREFUSED); break;
+    case EHOSTDOWN: errStr = ERRTAG(EHOSTDOWN); break;
+    case EHOSTUNREACH: errStr = ERRTAG(EHOSTUNREACH); break;
+    case ELOOP: errStr = ERRTAG(ELOOP); break;
+    case EOVERFLOW: errStr = ERRTAG(EOVERFLOW); break;
+    default:
+      block * retBlock = static_cast<block *>(koreAlloc(sizeof(block) + sizeof(uint64_t)));
```
dwightguth: it's more accurate to make this sizeof(block) + sizeof(mpz_ptr)

runtime/io/io.cpp
```diff
@@ -0,0 +1,440 @@
...
+
+  inline block * getKSeqErrorBlock() {
+    block * err = block_errno();
+    block * retBlock = static_cast<block *>(koreAlloc(sizeof(block) + 2 * sizeof(block *)));
+    block * inj = static_cast<block *>(koreAlloc(sizeof(block) + sizeof(block *)));
+    retBlock->h = kseqHeader;
+    inj->h = header_err();
+    memcpy(inj->children, &err, sizeof(block *));
+    memcpy(retBlock->children, &inj, sizeof(block *));
+    memcpy(&(retBlock->children[1]), &dotK, sizeof(block *));
```
dwightguth: retBlock->children + 1

runtime/io/io.cpp
```diff
@@ -0,0 +1,440 @@
...
+    return retBlock;
+  }
+
+  inline block* getInjErrorBlock() {
+    block * retBlock = static_cast<block *>(koreAlloc(sizeof(block) + sizeof(block *)));
+    retBlock->h = header_err();
+    block * p = block_errno();
```
dwightguth: it's probably incorrect to put this line after the alloc since it will screw up the errno.

runtime/io/io.cpp
```diff
@@ -0,0 +1,440 @@
...
+    auto result = static_cast<string *>(koreAllocToken(sizeof(string) + length));
+    int bytes = read(fd, &(result->data), length);
+
+    if (-1 == bytes) {
+      return getInjErrorBlock();
+    }
+
+    block * retBlock = static_cast<block *>(koreAlloc(sizeof(block) + sizeof(uint64_t)));
+    retBlock->h = header_string();
+    result = static_cast<string *>(koreResizeLastAlloc(result, bytes, length));
```
dwightguth: you can't call this after the other koreAlloc because it is no longer the last allocation. You need to do it before doing a second allocation.

runtime/io/io.cpp
```diff
@@ -0,0 +1,440 @@
...
+    set_len(result, bytes);
+    memcpy(retBlock->children, &result, sizeof(string *));
+    return retBlock;
+  }
+
+  block * hook_IO_close(mpz_t i) {
+    if (!mpz_fits_sint_p(i)) {
+      throw std::invalid_argument("Arg too large for int32_t");
```
dwightguth: as I said, this should be int, not int32_t. Same for the other ones that call sint. ulong should be unsigned long.

theo25:
I reviewed the code and it looks fine too me. But it seems you need to fix some expected error codes in the unit tests, because the PR fails on something like that. Also, as a nitpick, you could add `static` to methods in `runtime/io/io.cpp` that are not to be used outside of the file (e.g. `getTerminatedString`). This helps the inline optimization.

ciscript
```diff
@@ -53,4 +53,14 @@ llvm-kompile ../test/test12.kore TEST main -o interpreter
 ./interpreter ../test/test12-1.kore -1 /dev/stdout | diff - ../test/test12.out
 ./interpreter ../test/test12-2.kore -1 /dev/stdout | diff - ../test/test12.out
 ./interpreter ../test/test12-3.kore -1 /dev/stdout | diff - ../test/test12.out
+llvm-kompile ../test/io.kore IO main -o interpreter
+./interpreter ../test/ioTest1.kore -1 /dev/stdout | grep -q "Lbl'-LT-'k'-GT-'{}(dotk{}())"
+cat `ls | grep io_output1.txt`
```
dwightguth: should we not be checking the contents of this file?

ciscript
```diff
@@ -53,4 +53,14 @@ llvm-kompile ../test/test12.kore TEST main -o interpreter
...
+echo
+./interpreter ../test/ioTest2.kore -1 /dev/stdout | diff - ../test/ioTest2.out
+./interpreter ../test/ioTest3.kore -1 /dev/stdout | grep -q "Lbl'-LT-'k'-GT-'{}(dotk{}())"
+cat `ls | grep io_log`
```
dwightguth: same here.

mickyabir:
io.k:

```module IO-SYNTAX
  imports DOMAINS
  imports K-IO
  imports STRING

  syntax Stmt ::= "close" "(" Id ")" 
            | "seek" "(" Id "," Int ")" 
            | "seekEnd" "(" Id "," Int ")" 
            | "putc" "(" Id "," Int ")" 
            | "write" "(" Id "," String ")" 
            | "lock" "(" Id "," Int ")" 
            | "unlock" "(" Id "," Int ")" 
            | "log" "(" String "," String ")" 

  syntax IOExp ::= IOVal
                | "open" "(" String "," String ")" 
                | "tell" "(" Id ")" 
                | "getc" "(" Id ")" 
                | "read" "(" Id "," Int ")" 

  syntax IOVal ::= IOString | IOInt

  syntax Stmt ::= IOExp | Stmt Stmt [left]
                  | Id "=" IOExp ";" [strict(2)]

  syntax Pgm ::= "int" Ids ";" Stmt | Stmt
  syntax Ids ::= List{Id,","}

  syntax KResult ::= IOVal
endmodule

module IO
  imports IO-SYNTAX
  configuration <T color="yellow">
                  <k color="green"> $PGM:Pgm </k>
                  <state color="red"> .Map </state>
                </T>

  rule open(S:String, S':String) => #open(S, S') 
  rule log(S:String, S':String) => #logToFile(S, S') 

  rule <k> tell(X:Id) => #tell(I:Int) ...</k> <state>... X |-> I ...</state>
  rule <k> getc(X:Id) => #getc(I:Int) ...</k> <state>... X |-> I ...</state>
  rule <k> read(X:Id, I:Int) => #read(I':Int, I) ...</k> <state>... X |-> I' ...</state>
  rule <k> close(X:Id) => #close(I:Int) ...</k> <state>... X |-> I ...</state>
  rule <k> seek(X:Id, I:Int) => #seek(I':Int, I) ...</k> <state>... X |-> I' ...</state>
  rule <k> seekEnd(X:Id, I:Int) => #seekEnd(I':Int, I) ...</k> <state>... X |-> I' ...</state>
  rule <k> putc(X:Id, I:Int) => #putc(I':Int, I) ...</k> <state>... X |-> I' ...</state>
  rule <k> write(X:Id, S:String) => #write(I:Int, S) ...</k> <state>... X |-> I ...</state>
  rule <k> lock(X:Id, I:Int) => #lock(I':Int, I) ...</k> <state>... X |-> I' ...</state>
  rule <k> unlock(X:Id, I:Int) => #unlock(I':Int, I) ...</k> <state>... X |-> I' ...</state>

  rule <k> X = I:IOVal; => . ...</k> <state>... X |-> (_ => I) ...</state>
  rule <k> X:Id => I ...</k> <state>... X |-> I ...</state>
  rule S1:Stmt S2:Stmt => S1 ~> S2  [structural]

  rule <k> int (X,Xs => Xs);_ </k> <state> Rho:Map (.Map => X|->0) </state>
    requires notBool (X in keys(Rho))
  rule int .Ids; S => S  [structural]

endmodule

**Comment by mickyabir** [comment]
```
ioTest1.io:
int x;
x = open("io_output1.txt", "w+");
write(x, "hello world!")

ioTest2.io:
int x;
x = open("io_output2.txt", "w+");
write(x, "hello world!")
seek(x, 0)
read(x, 5)

ioTest3.io:
log("io_log.txt", "First log")

ioTest4.io:
int x;
x = open("io_output3.txt", "w+");
write(x, "hello ")
close(x)
x = open("io_output3.txt", "a+");
write(x, "world!")
close(x)
x = open("io_output3.txt", "r+");
seek(x, 0)
write(x, "howdy")
seekEnd(x, 0)
putc(x, 33)
seek(x, 0)
read(x, 13)

ciscript
```diff
@@ -53,4 +53,9 @@ llvm-kompile ../test/test12.kore TEST main -o interpreter
 ./interpreter ../test/test12-1.kore -1 /dev/stdout | diff - ../test/test12.out
 ./interpreter ../test/test12-2.kore -1 /dev/stdout | diff - ../test/test12.out
 ./interpreter ../test/test12-3.kore -1 /dev/stdout | diff - ../test/test12.out
+llvm-kompile ../test/io.kore IO main -o interpreter
+./interpreter ../test/ioTest1.kore -1 /dev/stdout | grep -q "Lbl'-LT-'k'-GT-'{}(dotk{}())"
```
dwightguth: I would like it if tests 1 and 3 checked the output of their respective files to make sure they're correct.

runtime/io/io.cpp
```diff
@@ -19,7 +20,7 @@ extern "C" {
   static block * dotK = (block *)((((uint64_t)getTagForSymbolName("dotk{}")) << 32) | 1);
   static blockheader kseqHeader = {getBlockHeaderForSymbol((uint64_t)getTagForSymbolName("kseq{}"))};
 
-  static std::map<std::string, std::string> logFiles;
+  static std::map<char *, std::string> logFiles;
```
dwightguth: I'm not sure this works. a std::map whose key is char * is going to compare strings with pointer equality, which is not what you want.

runtime/io/io.cpp
```diff
@@ -450,14 +450,21 @@ extern "C" {
     for (auto const& log : logFiles) {
       std::string pathStr = log.first;
       std::string msg = log.second;
-      char * path = strdup(pathStr.c_str());
-      char * dir = dirname(path);
-      char * base = basename(path);
+      size_t length = pathStr.length();
+      char * path1 = (char *) malloc(sizeof(char) * (length + 1));
+      strcpy(path1, pathStr.c_str());
+      path1[length] = '\0';
```
dwightguth: strcpy writes the null byte, you can delete this line and the other one below.

unittests/runtime-io/io.cpp
```diff
@@ -0,0 +1,343 @@
...
+BOOST_AUTO_TEST_CASE(open) {
+  int fd = overwriteTestFile();
+  ::close(fd);
+
+  auto realFilename = makeString("test.txt");
+  auto fakeFilename = makeString("testFake.txt");
+  auto control = makeString("r");
+
+  block * b1 = hook_IO_open(realFilename, control);
+  BOOST_CHECK(0 < mpz_cmp_si((mpz_ptr) *(b1->children), 0));
```
dwightguth: I'm concerned that this test will probably invoke undefined behavior if a bug causes open to not return an integer. It would be nice to explicitly assign mocked headers for inj{IOError} and inj{Int} and check against them in the test, just to make sure.

dwightguth: We should do this for all the tests, too.

unittests/runtime-io/io.cpp
```diff
@@ -0,0 +1,343 @@
...
+  mpz_t f2;
+  int fd1 = overwriteTestFile();
+  mpz_init_set_si(f1, fd1);
+  int fd2 = overwriteTestFile();
+  mpz_init_set_si(f2, fd2);
+
+  hook_IO_close(f1);
+  hook_IO_close(f2);
+
+  BOOST_CHECK_EQUAL(-1, fcntl(fd1, F_GETFD));
```
dwightguth: can we check that errno == EBADF too? otherwise this test isn't really testing that the file descriptor has been closed.

PR #99, opened 2019-03-06
Update to latest release of im-rc

dwightguth:
We update to the latest release of im-rc to fix a bug involving lists of length 4100 or greater. Since the latest release of im uses rust edition 2018, we also update the build system to build rust 1.33 with the system llvm 6.0.


install-rust
```diff
@@ -0,0 +1,15 @@
...
+curl -L https://static.rust-lang.org/dist/rustc-1.33.0-src.tar.gz -o rustc-src.tar.gz
+tar xvf rustc-src.tar.gz
+cd rustc-1.33.0-src
+./configure --llvm-root=/usr/lib/llvm-6.0 --enable-llvm-link-shared
+sudo ./x.py install
+cd ..
+rm -rf rustc-1.33.0-src rustc-src.tar.gz
+
+sudo apt-get install cargo
+git clone https://github.com/rust-lang/cargo -b rust-1.33.0 --depth 1
```
bmmoore: Are you sure that's the right branch? That branch is at the same commit as the 0.34.0 release tag, I guess they continue development toward another release on a "rust-X.Y.Z" branch?
I think it's a bit more stable to request a commit hash or a tag than to pick a branch.

Also, why is this using git while for rust we download a tar.gz of rust source?
The files this fetches should match https://github.com/rust-lang/cargo/archive/0.34.0.tar.gz (keeping in mind that branch rust-1.33.0 and tag 0.34.0 are at the same point).

bmmoore: Actually, per https://forge.rust-lang.org/release-process.html the cargo version for rust `1.Y.Z` is usually `0.(Y+1).0`, so `0.34.0` is the match for rust `1.33.0`. I still think any git checkout should use a tag or commit hash so the branch won't move on us.

PR #100, opened 2019-03-12
install rustc using rustup

dwightguth:
@bmmoore suggested we use rustup with a custom toolchain rather than installing rust globally, and so I made this PR to do exactly that. It also removes the source dependency on cargo, which is nice.


dwightguth:
we are installing rustup which installs stable by default. I assume it uses the default toolchain.

bmmoore:
So, that will be installing newer and newer releases over time, but we can expect cargo to stay backwards compatible for a long time.

Issue #45, opened 2018-11-01
Improve GC


dwightguth:
- [x] collect old generation
- [x] collect ints and floats
- [x] give struct stringbuffer a proper header
- [x] use block-entry alloca for sret
- [x] collect collections
- [ ] remove unnecessary load in struct layout

bmmoore:
Rather than simply removing the load in struct layout, we should probably try to arrange things so most layouts can be decoded from the header word alone, without needing any hash lookup at all.

Baltoli:
@theo25 What's the status of this?

ehildenb:
Closing for now, can re-open if we revisit.

ehildenb:
@theo25 please link to the upstream patches you've submitted here, and follow up on them. Then we can close.

theo25:
Here is the link to the upstream patch with a fix required for unwindig the stack of interpreters generated by the LLVM backend using libunwind: [https://reviews.llvm.org/D143618](https://reviews.llvm.org/D143618)

Baltoli:
@theo25 has been asked for a test by the LLVM team.

dwightguth:
We should be able to use the minimized test case I wrote when I was evaluating the issue. It will just need to be given appropriate lit directives.

theo25:
@dwightguth Unfortunately that test case does not trigger the bug, because its functions do not require any additional stack space when tail called. But I think I can modify it and get the required behavior for the bug to  trigger

Baltoli:
https://github.com/llvm/llvm-project/commit/2e0940c6a0857e3cc6aecfe7dd0c4cde94e1ae9a patch has now landed - @theo25 can the issue be closed now?

PR #102, opened 2019-03-14
Garbage collection for integers and floats

theo25:
This PR implements garbage collection for integers and floats. The strategy is as following:
- We add block headers to the `mpz_t` and `floating` blocks, so that garbage collection can figure out their size and access their gc bits. For `mpz_t`, we create a wrapper structure, namely `mpz_hdr`, that contains the block header. Similarly for `floating`, we create a wrapper structure `floating_hdr`. Only the GC code deals with the new structures. The rest of the backend sees only `mpz_t` and `floating` objects, same as before.
- These block headers have the following format: the most significant 10 bits are zeroed and serve as the layout code, the next 4 bits are the GC bits, and the least significant 40 bits hold the size of the structure minus the size of the header (so that `get_size` compute the correct size).
- We add special allocators for integers and floats that appropriately initialize their block headers.
- We add a special allocator to serve as the hook for GMP that will be used to allocate limbs. It allocates limbs as `string` blocks and return a pointer to the `data` field (skipping the block header, since GMP does not know about it).
- Finally, two integration tests are added, one for integer gc and one for floating gc. Both tests are imp programs that compute the first 1000 fibbonacci numbers in a loop.

It is worth noting that when an `mpfr_t` is initialized by the MPFR library (and its limbs are allocated), the `_mpfr_d` field is not assigned the pointer returned by the allocator hook, but rather a pointer offset by 8 bytes. We have to take this into account when migrating `floating` objects to correctly figure out the block header of the limbs block.

Addresses #45


INSTALL.md
```diff
@@ -3,10 +3,10 @@ On Ubuntu 18.04:
 ```
 sudo apt-get update
 sudo apt-get install git cmake clang-6.0 llvm-6.0-tools zlib1g-dev bison flex libboost-test-dev libgmp-dev libmpfr-dev libyaml-cpp-dev libjemalloc-dev curl libtinfo-dev
-./install-rust
```
dwightguth: can you explain this change? if it's just because you want to change the directory that the rust sources get placed in, I honestly think it makes more sense for it to be outside so that users don't delete it when they run `git clean -fxd`.

runtime/alloc/alloc.c
```diff
@@ -95,11 +95,42 @@ void* koreResizeLastAlloc(void* oldptr, size_t newrequest, size_t last_size) {
...
+  memcpy(new->data, ptr, min);
+  set_len(new, new_size);
+  return new->data;
 }
 
 void koreFree(void* ptr, size_t size) {}
+
+__attribute__ ((always_inline)) void* koreAllocInteger(size_t requested) {
+  integer *result = (integer *) koreAlloc(sizeof(integer));
+  set_len(result, sizeof(mpz_t));
```
dwightguth: can we make this sizeof(integer) - sizeof(blockheader)?

runtime/alloc/alloc.c
```diff
@@ -95,11 +95,42 @@ void* koreResizeLastAlloc(void* oldptr, size_t newrequest, size_t last_size) {
...
+  return &result->i;
+}
+
+__attribute__ ((always_inline)) void* koreAllocFloating(size_t requested) {
+  floating *result = (floating *) koreAlloc(sizeof(floating));
+  set_len(result, sizeof(mpfr_t)+8);
```
dwightguth: same but with floating

runtime/alloc/alloc.c
```diff
@@ -95,11 +95,42 @@ void* koreResizeLastAlloc(void* oldptr, size_t newrequest, size_t last_size) {
...
+  return result;
+}
+
+__attribute__ ((always_inline)) void* koreAllocIntegerOld(size_t requested) {
+  integer *result = (integer *) koreAllocOld(sizeof(integer));
+  set_len(result, sizeof(mpz_t));
```
dwightguth: and here

runtime/alloc/alloc.c
```diff
@@ -95,11 +95,42 @@ void* koreResizeLastAlloc(void* oldptr, size_t newrequest, size_t last_size) {
...
+  return &result->i;
+}
+
+__attribute__ ((always_inline)) void* koreAllocFloatingOld(size_t requested) {
+  floating *result = (floating *) koreAllocOld(sizeof(floating));
+  set_len(result, sizeof(mpfr_t)+8);
```
dwightguth: and here

runtime/equality.ll
```diff
@@ -7,7 +7,7 @@ target triple = "x86_64-unknown-linux-gnu"
 %set = type { i8 *, i8 *, i64 } ; im::hashset::HashSet
 %list = type { i64, [7 x i64] } ; im::vector::Vector
 %mpz = type { i32, i32, i64 * } ; mpz_t
-%floating = type { i64, { i64, i32, i64, i64 * } } ; exp, mpfr_t
+%floating = type { %blockheader, i64, { i64, i32, i64, i64 * } } ; 16-bit layout, 48-bit length, exp, mpfr_t
```
dwightguth: layout is actually 10 bits and length is 40

runtime/hash.ll
```diff
@@ -7,7 +7,7 @@ target triple = "x86_64-unknown-linux-gnu"
 %set = type { i8 *, i8 *, i64 } ; im::hashset::HashSet
 %list = type { i64, [7 x i64] } ; im::vector::Vector
 %mpz = type { i32, i32, i64 * } ; mpz_t
-%floating = type { i64, { i64, i32, i64, i64 * } } ; exp, mpfr_t
+%floating = type { %blockheader, i64, { i64, i32, i64, i64 * } } ; 16-bit layout, 48-bit length, exp, mpfr_t
```
dwightguth: and here

runtime/move_float.ll
```diff
@@ -1,16 +1,23 @@
 target datalayout = "e-m:e-i64:64-f80:128-n8:16:32:64-S128"
 target triple = "x86_64-unknown-linux-gnu"
 
-%floating = type { i64, { i64, i32, i64, i64 * } } ; exp, mpfr_t
+%blockheader = type { i64 }
+%floating = type { %blockheader, i64, { i64, i32, i64, i64 * } } ; 16-bit layout, 48-bit length, exp, mpfr_t
```
dwightguth: and here

runtime/move_float.ll
```diff
@@ -1,16 +1,23 @@
...
 
 ; helper function for float hooks
 define %floating* @move_float(%floating* %val) {
-  %loaded = load %floating, %floating* %val
-  %malloccall = tail call i8* @koreAllocNoGC(i64 ptrtoint (%floating* getelementptr (%floating, %floating* null, i32 1) to i64))
+  %p1 = getelementptr inbounds %floating, %floating* %val, i64 0, i32 1
```
dwightguth: why has this function gotten more complicated? I would have thought you just needed to change the allocation function.

INSTALL.md
theo25: I just moved after the `git clone` because the script is not available before you do that. I guess we can change further to make the script execute outside of the repo dir. Do you want me to do that?

dwightguth: that makes sense, but yes, let's do it before the cd.

runtime/move_float.ll
theo25: Now that `floating` has a block header, if we just changed the allocation function to `koreAllocFloating` and then copied the whole floating object on the allocated space, the initialization of the header performed by `koreAllocFloating` would be undone. So I had to either copy only the non-header part or reinitialize the header. I thought the latter is duplicating the logic of header initialization, so I chose the former approach.

theo25:
I fixed and/or replied to all your comments @dwightguth. Please take a look again when you can.

runtime/move_float.ll
dwightguth: I now understand what is going on here, but I think I have decided I don't like what this does to the code asymmetry to treat floats differently than ints. Is it possible we could have a nested struct containing just the exponent and mpfr_t and treat that nested struct the same way we treat mpz_t?

theo25: Yes I don't see why not. It would require a lot of changes in the implementation of float hooks and elsewhere, although mostly refactoring changes. I can do if you are OK with that.

dwightguth: yeah, I think that's probably the right approach. Either that or to make blocks contain a integer * instead of an mpz_ptr. Personally I like the former better because it means that the hooks don't need to worry about block headers.

theo25: I implemented the discussed refactoring. Please take a look again when you can.

dwightguth:
I approved but as I've said elsewhere I would like @bmmoore to review also.

runtime/alloc/collect.c
```diff
@@ -111,6 +112,78 @@ static void migrate_string_buffer(stringbuffer** bufferPtr) {
   *bufferPtr = *(stringbuffer **)(buffer->contents);
 }
 
+static void migrate_mpz(mpz_ptr *mpzPtr) {
+  mpz_hdr *intgr = struct_base(mpz_hdr, i, *mpzPtr);
+  const uint64_t hdr = intgr->h.hdr;
+  bool isNotInYoungGen = hdr & NOT_YOUNG_OBJECT_BIT;
+  bool hasAged = hdr & YOUNG_AGE_BIT;
+  if (isNotInYoungGen && !(hasAged && collect_old)) {
```
bmmoore: I think something is poorly named around here. Why is a mask called `YOUNG_AGE_BIT` considered when the object `isNotInYoungGen`?

dwightguth: YOUNG_AGE_BIT gets set when the object survives its first young generation collection. It then remains set when the object is promoted to the old generation. We then check both YOUNG_AGE_BIT and NOT_YOUNG_OBJECT_BIT because a combination of both tells us that the object is in the old generation. If you have a better name to suggest, I'm willing to listen though.

bmmoore: The `YOUNG` only refers to the role of that bit in the young generation. If it means the same thing in headers in the old generation it could have a more generic name like `AGE_BIT`. If the bit doesn't really mean the same thing, there could be multiple names for the same mask value and I expect the optimizer would remove any apparent redundancy (especially considering hdr is used from a local variable). Perhaps a macro or inline function `isInOldGen(hdr)` computing `(hdr & NOT_YOUNG_OBJECT_BIT) && (hdr & YOUNG_AGE_BIT)`,

theo25: In general I think I agree with @bmmoore that the naming of the flags is a little confusing. I don't think though that renaming them should be part of this PR. So if that's ok with everybody, I will go ahead and merge it.

dwightguth: I don't think it's too much effort to make a macro like Brandon suggested. Why don't we aim for it as part of this pr, even if we don't rename anything.

theo25: OK that sounds fair. I will do it

dwightguth:
Please fix conflicts and merge with master. I will then take a quick look at the merge commit and then hopefully merge it.

runtime/alloc/collect.c
theo25: I pushed a commit that adds the discussed macro. @bmmoore please take a look again

PR #109, opened 2019-04-05
Fix for correct stringbuffer collection.

theo25:
This PR fixes a bug in the string buffer object's header where an evacuated stringbuffer object with incorrect size information in its header would lead to an incorrect scan, since the scanner would not know how to skip it. With this PR, the stringbuffer object header correctly reflects its size. An additional field in the buffer contains the length of the pointed string, while the capacity of the buffer is recorded in the header of the pointed string object (since it is also the actual size of that object). String objects not pointed by stringbuffers still record their length (which in this case is equal to the actual object size) in their header.

Addresses #45


theo25:
It is a simple language that has a stringbuffer as part of the state and supports a command that appends content to the buffer and a command that loops a given number of times, doing nothing but rewriting itself and reducing a counter. This way I can have programs that have enough steps to trigger multiple collections which the stringbuffer should survive. This is the case that was failing before and now succeeds.

PR #111, opened 2019-04-05
Allocation of sret parameters of hooks in the stack

theo25:
Draft - Do not review yet.

This PR replaces the previous allocation strategy for sret parameters of hooks (allocation in the non-collected heap) with allocation in the stack using LLVM's `alloca`, so that the sret objects are automatically freed when the function containing the call to the hook returns.

This PR also removes a remaining allocation in the non-collected heap found in the implementation of the string hooks.

Addresses #45


dwightguth:
Looks good but did you examine the optimized llvm code for evm to make sure that the code for tail recursive functions affected by this change is still tail recursive?

theo25:
I just did the check that @dwightguth asked. Unfortunately, I have found at least one example of function that is tail recursive and is optimized as such without the stack allocations but is not optimized with the stack allocations. I will investigate and mark the PR as draft again.

dwightguth:
To clarify: it's fairly important that functions that are written in a tail recursive style in K generate tail-recursive llvm code. Otherwise, there is no way to avoid stack overflows from deeply nested function calls in K. This only applies to eval_Lbl functions and apply_rule_ functions though. Other functions are not important.

PR #112, opened 2019-04-08
FFI hooks

mickyabir:
Implements the ffiCall and functionAddress hooks for arithmetic and void types.


bin/llvm-kompile-clang
```diff
@@ -21,6 +21,7 @@ fi
   "$LIBDIR"/libconfigurationparser.a \
   "$LIBDIR"/libstrings.a \
   "$LIBDIR"/libio.a \
+  "$LIBDIR"/libffi.a \
```
dwightguth: let's put this in libmeta.

runtime/ffi/ffi.cpp
```diff
@@ -0,0 +1,200 @@
...
+  char * getTerminatedString(string * str);
+
+  mpz_ptr hook_LIST_size(struct list * l);
+  block * hook_LIST_get(struct list * l, int idx);
+
+  static void * so_lib_handle() {
+    static void * handle = NULL;
+
+    if (handle == NULL) {
+      handle = dlopen(NULL, RTLD_LAZY);
```
dwightguth: this returns null on failure so it would be good to check for null after this line and abort with an error.

runtime/ffi/ffi.cpp
```diff
@@ -0,0 +1,200 @@
...
+      return ffi_type_longdouble;
+    } else {
+      throw std::invalid_argument("Arg is not a supported type");
+    }
+  }
+
+  block * hook_FFI_call(mpz_t addr, struct list * args, struct list * types, block * ret) {
+    ffi_cif cif;
+    ffi_type *argtypes, rtype;
+    void (* address)();
```
dwightguth: this should be (void), which is not equivalent to ().

runtime/ffi/ffi.cpp
```diff
@@ -0,0 +1,200 @@
...
+
+    if (!mpz_fits_ulong_p(addr)) {
+      throw std::invalid_argument("Addr is too large");
+    }
+    *(void **)(&address) = (void *) mpz_get_ui(addr);
+
+    mpz_ptr nargs_ptr = hook_LIST_size(args);
```
dwightguth: please use hook_LIST_size_long for this.

runtime/ffi/ffi.cpp
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    size_t nargs = mpz_get_si(nargs_ptr);
+    size_t ntypes = mpz_get_si(ntypes_ptr);
+
+    if (nargs != ntypes) {
+      throw std::invalid_argument("Args size does not match types size");
+    }
+
+    argtypes = (ffi_type *) malloc(sizeof(ffi_type) * nargs);
```
dwightguth: this is wrong. The function takes an `ffi_type *[nargs]` (ie, an array nargs of pointer to struct ffi_type), not an `ffi_type (*)[nargs]` (ie, a pointer to an array nargs of struct ffi_type).

runtime/ffi/ffi.cpp
```diff
@@ -0,0 +1,200 @@
...
+    static void * handle = NULL;
+
+    if (handle == NULL) {
+      handle = dlopen(NULL, RTLD_LAZY);
+    }
+
+    return handle;
+  }
+
+  static ffi_type getTypeFromSymbol(uint64_t symbol) {
```
dwightguth: this should return an ffi_type * (see comment below) obtained by getting the address of the constant struct.

runtime/ffi/ffi.cpp
```diff
@@ -0,0 +1,200 @@
...
+    } else if (symbol == tag_type_longdouble()) {
+      return ffi_type_longdouble;
+    } else {
+      throw std::invalid_argument("Arg is not a supported type");
+    }
+  }
+
+  block * hook_FFI_call(mpz_t addr, struct list * args, struct list * types, block * ret) {
+    ffi_cif cif;
+    ffi_type *argtypes, rtype;
```
dwightguth: argtypes should be ffi_type **.

runtime/ffi/ffi.cpp
```diff
@@ -0,0 +1,200 @@
...
+    mpz_ptr nargs_ptr = hook_LIST_size(args);
+    mpz_ptr ntypes_ptr = hook_LIST_size(types);
+    if (!mpz_fits_ulong_p(nargs_ptr)) {
+      throw std::invalid_argument("Args list is too large");
+    }
+    if (!mpz_fits_ulong_p(ntypes_ptr)) {
+      throw std::invalid_argument("Types list is too large");
+    }
+
+    size_t nargs = mpz_get_si(nargs_ptr);
```
dwightguth: this will be removed by the above comment, but note that size_t is unsigned so it really should have been mpz_get_ui.

runtime/ffi/ffi.cpp
```diff
@@ -0,0 +1,200 @@
...
+    size_t ntypes = mpz_get_si(ntypes_ptr);
+
+    if (nargs != ntypes) {
+      throw std::invalid_argument("Args size does not match types size");
+    }
+
+    argtypes = (ffi_type *) malloc(sizeof(ffi_type) * nargs);
+
+    block * type, * elem;
+    ffi_type arg_t;
```
dwightguth: based on the new return type of getTypeFromSymbol this should be a pointer.

runtime/ffi/ffi.cpp
```diff
@@ -0,0 +1,200 @@
...
+          throw std::invalid_argument("Types list contains invalid FFI type");
+        }
+        type = (block *) *elem->children;
+        arg_t = getTypeFromSymbol((uint64_t)type);
+        argtypes[i] = arg_t;
+        avalues_size += arg_t.size;
+    }
+
+    string * bytes;
+    void * avalues = malloc(avalues_size);
```
dwightguth: this has the same problem. You are calling ffi_call with a void (*)[] when you should be calling it with a void *[nargs].

runtime/ffi/ffi.cpp
```diff
@@ -0,0 +1,200 @@
...
+    size_t len, offset = 0;
+    for (int i = 0; i < nargs; i++) {
+        elem = hook_LIST_get(args, i);
+        if (elem->h.hdr != (uint64_t)getTagForSymbolName("inj{SortBytes{}}")) {
+          throw std::invalid_argument("Args list contains non-bytes type");
+        }
+        bytes = (string *) *elem->children;
+        len = len(bytes);
+        memcpy((char *)avalues + offset, bytes, sizeof(char) * len);
```
dwightguth: you should not need to memcpy here. Since it is expecting a pointer to an array of bytes, and you have a pointer to an array of bytes, you can just write the pointer to the array of pointers.

runtime/ffi/ffi.cpp
```diff
@@ -0,0 +1,200 @@
...
+        offset += len;
+    }
+
+    rtype = getTypeFromSymbol((uint64_t)ret);
+
+    ffi_status status = ffi_prep_cif(&cif, FFI_DEFAULT_ABI, nargs, &rtype, &argtypes);
```
dwightguth: you should not have to dereference argtypes here.

runtime/ffi/ffi.cpp
```diff
@@ -0,0 +1,200 @@
...
+        break;
+      case FFI_BAD_TYPEDEF:
+          throw std::invalid_argument("Types list contains invalid FFI type");
+        break;
+      case FFI_BAD_ABI:
+          throw std::invalid_argument("Invalid ABI mode");
+        break;
+    }
+
+    void * rvalue = malloc(rtype.size);
```
dwightguth: we should be able to avoid this double allocation by calling koreAlloc before ffi_call and then passing the pointer to the data buffer.

runtime/ffi/ffi.cpp
```diff
@@ -0,0 +1,200 @@
...
+    ffi_call(&cif, address, rvalue, &avalues);
+
+    free(avalues);
+
+    block * retBlock = static_cast<block *>(koreAlloc(sizeof(block) + rtype.size));
+    retBlock->h = getBlockHeaderForSymbol((uint64_t)getTagForSymbolName("inj{SortBytes{}}"));
```
dwightguth: this isn't right. you should be casting to a string * and calling set_len.

runtime/ffi/ffi.cpp
```diff
@@ -0,0 +1,200 @@
...
+    return retBlock;
+  }
+
+  mpz_ptr hook_FFI_address(string * fn) {
+    char * func = getTerminatedString(fn);
+    void * handle = so_lib_handle();
+    void * address = dlsym(handle, func);
+
+    mpz_t result;
+    mpz_init_set_ui(result, (uint64_t)address);
```
dwightguth: cast address to a uintptr_t.

unittests/runtime-ffi/ffi.cpp
```diff
@@ -0,0 +1,200 @@
...
+  addr = hook_FFI_address(fn);
+  BOOST_CHECK(0 < mpz_cmp_ui(addr, 0));
+
+  fn = makeString("getX");
+  addr = hook_FFI_address(fn);
+  BOOST_CHECK(0 < mpz_cmp_ui(addr, 0));
+
+  fn = makeString("increaseX");
+  addr = hook_FFI_address(fn);
+  BOOST_CHECK(0 < mpz_cmp_ui(addr, 0));
```
dwightguth: can we have a test where the symbol is not found?

PR #114, opened 2019-04-10
refactor out sets and clean up some excessive refactoring

dwightguth:
This is the first step towards hopefully encapsulating all of the
map and set code inside Map.hs and Set.hs, which I hope to do prior to
working on map/set choices.

There were a few cases with extremely tiny functions that I elected not to move the functionality relating to sets into its own module and instead to move the functionality relating to maps back into the main module. We will explore further what to do about these cases when we take the next step in the refactoring progression; for now I'm focused on migrating the main body of the logic pertaining to all the different pattern types into their own modules.

I tested this on EVM and all tests pass.

### DIFF
--- a/matching/src/Everything.hs
+++ b/matching/src/Everything.hs
@@ -29,7 +29,7 @@ import Data.Function
 import Data.Functor.Foldable
        ( Fix (..) )
 import Data.List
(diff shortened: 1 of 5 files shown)


dwightguth:
Closing in favor of scala rewrite.

PR #125, opened 2019-05-10
FFI Struct Types

mickyabir:
Adds support for FFI struct types


runtime/meta/ffi.cpp
```diff
@@ -103,8 +106,48 @@ extern "C" {
     } else if (symbol == tag_type_longdouble()) {
       return &ffi_type_longdouble;
     } else {
-      throw std::invalid_argument("Arg is not a supported type");
+      return NULL;
```
dwightguth: let's add a case here for structs rather than exposing the logic relating to structs outside of getTypeFromSymbol.

runtime/meta/ffi.cpp
```diff
@@ -61,50 +64,86 @@ extern "C" {
...
+        return &ffi_type_ulong;
+      } else if (symbol == tag_type_slong()) {
+        return &ffi_type_slong;
+      } else if (symbol == tag_type_longdouble()) {
+        return &ffi_type_longdouble;
+      }
+    } else if (elem->h.hdr == (uint64_t)getTagForSymbolName(TYPETAG(struct))){
+      struct list * elements = (struct list *) *elem->children;
+      size_t numFields = hook_LIST_size_long(elements);
+      std::vector<ffi_type *> structElements;
```
dwightguth: I think we can avoid the double allocation here and just write directly to structTypes->elements.

runtime/meta/ffi.cpp
```diff
@@ -61,50 +64,86 @@ extern "C" {
...
+
+        structElements.push_back(type);
+        structSize += type->size;
+      }
+
+      ffi_type * structType = (ffi_type *) malloc(sizeof(ffi_type));
+      structType->size = 0;
+      structType->alignment = 0;
+      structType->type = FFI_TYPE_STRUCT;
+      structType->elements = (ffi_type **) malloc(sizeof(ffi_type *) * (structSize + 1));
```
dwightguth: this should actually be numFields+1.

runtime/meta/ffi.cpp
```diff
@@ -61,50 +64,86 @@ extern "C" {
...
+      } else if (symbol == tag_type_longdouble()) {
+        return &ffi_type_longdouble;
+      }
+    } else if (elem->h.hdr == (uint64_t)getTagForSymbolName(TYPETAG(struct))){
+      struct list * elements = (struct list *) *elem->children;
+      size_t numFields = hook_LIST_size_long(elements);
+      std::vector<ffi_type *> structElements;
+      block * structField;
+      ffi_type * type;
+      size_t structSize = 0;
```
dwightguth: based on the below comment, you don't need this field.

runtime/meta/ffi.cpp
```diff
@@ -166,6 +206,11 @@ extern "C" {
     set_len(rvalue, rtype->size);
     free(avalues);
 
+    for (auto &s : structTypes) {
+      free(s->elements);
+      free(s);
+    }
+
```
dwightguth: don't forget to call structTypes.clear() here otherwise you'll have a double free when you call this method a second time.

runtime/meta/ffi.cpp
```diff
@@ -20,6 +21,8 @@ extern "C" {
   return tag; \
 }
 
+  static std::vector<ffi_type *> structTypes;
```
dwightguth: just to be on the safe side, can we make this thread_local static? There is at least a small possibility we may at some point have multiple threads executing K code at the same time.

PR #139, opened 2019-06-10
FFI Pointers Hooks

mickyabir:


unittests/runtime-ffi/ffi.cpp
```diff
@@ -312,6 +312,33 @@ BOOST_AUTO_TEST_CASE(call) {
   ret = *(int *) bytes->data;
 
   BOOST_CHECK_EQUAL(ret, p2.p.x * p2.p.y);
+
+  /* int pointerTest(int * x) */
+  x = 2;
+  int * ptr = &x;
+  string * ptrargstr = makeString((char *) &ptr, sizeof(int *)); 
+
+  block * ptrarg = static_cast<block *>(koreAlloc(sizeof(block) + sizeof(string *)));
```
dwightguth: can we actually call hook_FFI_alloc, get a string *, write to it, call hook_FFI_bytes_address, call mpz_get_ui, write the value to the argument to call, then call, then call hook_FFI_free? I'd like to test this use case a bit more explicitly.

dwightguth:
Shit, I just realized, this PR isn't going to work as-is because it introduces a new GC root that we're not processing correctly. I'll get theo to make an easy way to add new GC roots, but we shouldn't merge this yet.

mickyabir:
The module definition is essentially IMP with added pointer functions. The test used is the following:

`int x;
pointerAlloc(1, 4)
x = 0;
while (x <= 5000) {
  x = x + 1;
}
pointerFree(1)`

PR #140, opened 2019-06-10
Elimination of no-gc allocations for sret parameters

theo25:
This PR eliminates allocations in the not collected arena for sret paremeters of llvm functions generated by the backend. Instead, a new arena is introduced that is always fully collected in every collection and the sret parameters are allocated there when needed. This is OK, since these allocations are not supposed to to be live between steps.

Addresses #45

### DIFF
--- a/runtime/alloc/alloc.c
+++ b/runtime/alloc/alloc.c
@@ -13,6 +13,7 @@
 REGISTER_ARENA(youngspace, 0);
 REGISTER_ARENA(oldspace, 1);
 REGISTER_ARENA(nogcspace, 2);
+REGISTER_ARENA(alwaysgcspace, 3);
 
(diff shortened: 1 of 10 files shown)


runtime/alloc/alloc.c
```diff
@@ -13,6 +13,7 @@
 REGISTER_ARENA(youngspace, 0);
 REGISTER_ARENA(oldspace, 1);
 REGISTER_ARENA(nogcspace, 2);
+REGISTER_ARENA(alwaysgcspace, 3);
```
dwightguth: we could save some memory usage if we don't make this space have two parts.

theo25: I agree and changed this

PR #142, opened 2019-06-13
Added support for multiple gc roots

theo25:


dwightguth:
I think this needs to be either more generic, or less stl'd. the issue is that the intended use case is we have two maps and we want to call register once on the keys of one map and once on the values of the other. But as this is, it's going to end up registering the address of some temporary vector rather than the address of an actual location within a map, and then when you try to migrate, it's not going to correctly migrate the map.

PR #147, opened 2019-06-19
Heuristic small default

mickyabir:


matching/src/main/scala/org/kframework/backend/llvm/matching/Heuristics.scala
```diff
@@ -91,3 +91,19 @@ object AHeuristic extends Heuristic {
     result
   }
 }
+
+object SDHeuristic extends Heuristic {
```
dwightguth: the paper calls this heuristic D so let's do the same.

matching/src/main/scala/org/kframework/backend/llvm/matching/Heuristics.scala
```diff
@@ -91,3 +91,19 @@ object AHeuristic extends Heuristic {
...
+  def scoreLiteral[T](p: LiteralP[T], f: Fringe, c: Clause, key: Option[Pattern[Option[Occurrence]]], isEmpty: Boolean): Double = ???
+  def scoreMap[T](p: MapP[T], f: Fringe, c: Clause, key: Option[Pattern[Option[Occurrence]]], isEmpty: Boolean): Double = ???
+  def scoreOr[T](p: OrP[T], f: Fringe, c: Clause, key: Option[Pattern[Option[Occurrence]]], isEmpty: Boolean): Double = ???
+  def scoreSet[T](p: SetP[T], f: Fringe, c: Clause, key: Option[Pattern[Option[Occurrence]]], isEmpty: Boolean): Double = ???
+  def scoreSymbol[T](p: SymbolP[T], f: Fringe, c: Clause, key: Option[Pattern[Option[Occurrence]]], isEmpty: Boolean): Double = ???
+  def scoreVariable[T](p: VariableP[T], f: Fringe, c: Clause, key: Option[Pattern[Option[Occurrence]]], isEmpty: Boolean): Double = ???
+  def scoreWildcard[T](p: WildcardP[T], f: Fringe, c: Clause, key: Option[Pattern[Option[Occurrence]]], isEmpty: Boolean): Double = ???
+
+   def computeScoreForKey(c: Column, key: Option[Pattern[Option[Occurrence]]]): Double = {
+    -(c.patterns.filter(_.isWildcard).size)
```
dwightguth: based on the name of this heuristic, small default, I think we should probably go with isDefault, not isWildcard.

dwightguth: let's also do c.patterns.count(_.isDefault) instead of filter(...).size.

PR #149, opened 2019-06-20
remove spurious phi nodes

dwightguth:
Don't generate phi nodes for basic blocks with only a single predecessor.

### DIFF
--- a/lib/codegen/Decision.cpp
+++ b/lib/codegen/Decision.cpp
@@ -34,6 +34,14 @@ void Decision::operator()(DecisionNode *entry, llvm::StringMap<llvm::Value *> su
       }
     }
   }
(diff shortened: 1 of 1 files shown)


lib/codegen/Decision.cpp
```diff
@@ -34,6 +34,14 @@ void Decision::operator()(DecisionNode *entry, llvm::StringMap<llvm::Value *> su
       }
     }
   }
+  for (auto block = this->CurrentBlock->getParent()->begin(); block != this->CurrentBlock->getParent()->end(); ++block) {
+    if (block->getUniquePredecessor()) {
+      for (auto phi = block->phis().begin(); phi != block->phis().end(); phi = block->phis().begin()) {
+        phi->replaceAllUsesWith(phi->getIncomingValue(0));
+	phi->removeFromParent();
```
theo25: nit: wrong indentation

PR #156, opened 2019-06-21
Heuristic R

mickyabir:
adds heuristic r and reorders the heuristics


matching/src/main/scala/org/kframework/backend/llvm/matching/Heuristics.scala
```diff
@@ -99,6 +89,37 @@ object BHeuristic extends Heuristic {
...
+    var result = 0.0
+    for (con <- c.signatureForKey(key)) {
+      for (i <- c.patterns.indices) {
+        if (c.patterns(i).isSpecialized(con, c.fringe, c.clauses(i))) {
+          result += 1.0
+        }
+      }
+    }
+
+    for (i <- c.patterns.indices) {
```
dwightguth: I forgot to mention this, but there's one more thing needed here. You need to guard this for loop by an `if (c.category.hasIncompleteSignature(sigma, c.fringe))`

Where sigma is the stored value of c.signatureForKey(key), which you should call only once.

PR #165, opened 2019-06-26
Best column scoring fix

mickyabir:


matching/src/main/scala/org/kframework/backend/llvm/matching/Heuristics.scala
```diff
@@ -15,23 +15,35 @@ sealed trait Heuristic {
...
+  def getBest(cols: Seq[(Column, Int)], allCols: Seq[(Column, Int)]): Seq[(Column, Int)] = {
     var result: List[(Column, Int)] = Nil
     var best = cols(0)._1.score
     for (col <- cols) {
       import Ordering.Implicits._
-      if (col._1.score > best) {
-        best = col._1.score
+      import scala.math.max
+
+      val bestInvalid = allCols.filter(c => !c._1.isValid && col._1.needed(c._1.keyVars)).sortBy(_._1.score).headOption
```
dwightguth: ironically, Column.isValid doesn't actually compute whether the column is valid or not. You should probably rename the function to isMaybeValid to make it clearer, but the idea is that some properties relating to validity are local to the column and thus cached, and others depend on the other columns and thus recomputed each time we call bestColIx. So you should probably do the following instead:
```
val validIndices = cols.map(_._2).toSet
val invalidCols = allCols.filterNot(c => validIndices(c._2))
val bestInvalid = invalidCols.filter(c => col._1.needed(c._1.keyVars)).sortBy(_._1.score).headOption
```

matching/src/main/scala/org/kframework/backend/llvm/matching/Heuristics.scala
```diff
@@ -15,23 +15,35 @@ sealed trait Heuristic {
...
+      var colBest = col._1.score
+
+      if (!bestInvalid.isDefined) {
```
dwightguth: you can just say ```
val colBest = if (bestInvalid.isDefined && bestInvalid.get._1.score > col._1.score) {
  bestInvalid.get._1.score
} else 
  col._1.score
}
```

dwightguth:
This "fix" seems to work but makes the decision tree bigger than the "hack", so I am going to close this and we will try again with a heuristic that scores columns based on whether they are needed.

PR #174, opened 2019-07-02
MatrixColumn refactor cont.

mickyabir:


matching/src/main/scala/org/kframework/backend/llvm/matching/Heuristics.scala
```diff
@@ -18,15 +18,16 @@ sealed trait Heuristic {
 
   def computeScoreForKey(c: AbstractColumn, key: Option[Pattern[Option[Occurrence]]]): Double
 
-  def getBest(cols: Seq[(Column, Int)]): Seq[(Column, Int)] = {
+  def getBest(cols: Seq[(Column, Int)], matrix: Matrix): Seq[(Column, Int)] = {
```
dwightguth: let's make this method take and return a Seq[MatrixColumn] instead of a Seq[(Column, Int)].

matching/src/main/scala/org/kframework/backend/llvm/matching/Heuristics.scala
```diff
@@ -18,15 +18,16 @@ sealed trait Heuristic {
...
     var result: List[(Column, Int)] = Nil
-    var best = cols(0)._1.score
+    var best = cols(0)._1.score(new MatrixColumn(matrix, 0))
     for (col <- cols) {
       import Ordering.Implicits._
-      if (col._1.score > best) {
-        best = col._1.score
+      val score = col._1.score(new MatrixColumn(matrix, cols.indexOf(col)))
```
dwightguth: the index should be col._2. You don't need to call indexOf.

matching/src/main/scala/org/kframework/backend/llvm/matching/Matrix.scala
```diff
@@ -8,28 +8,16 @@ import java.util
...
-  val fringe: Fringe = matrix.columns(colIx).fringe
-  def signatureForKey(key: Option[Pattern[Option[Occurrence]]]): List[Constructor] = matrix.columns(colIx).signatureForKey(key)
-  lazy val isEmpty = matrix.columns(colIx).isEmpty
-  lazy val category: SortCategory = matrix.columns(colIx).category
-  def maxPriority: Int = matrix.columns(colIx).maxPriority
+  val column: Column = matrix.columns(colIx)
 }
 
 class Column(val fringe: Fringe, val patterns: IndexedSeq[Pattern[String]], val clauses: IndexedSeq[Clause]) extends AbstractColumn {
+  val column: Column = this
```
dwightguth: let's make this a def instead of a val so it doesn't take up an extra word of memory.

PR #176, opened 2019-07-08
l heuristic and refactor

mickyabir:


matching/src/main/scala/org/kframework/backend/llvm/matching/pattern/SortCategory.scala
```diff
@@ -48,9 +48,9 @@ abstract class EqualLiteral() extends SortCategory {
 
   private def tree(matrix: Matrix, ls: Seq[(String, Matrix)]): DecisionTree = {
     val litO = matrix.bestCol.fringe.occurrence
-    if (matrix.default.isDefined && ls.isEmpty) {
+    if (matrix.default(matrix.bestColIx).isDefined && ls.isEmpty) {
```
dwightguth: let's cache this function call.

matching/src/main/scala/org/kframework/backend/llvm/matching/Heuristics.scala
```diff
@@ -149,6 +149,31 @@ object AHeuristic extends Heuristic {
...
 
+object LHeuristic extends Heuristic {
+  val needsMatrix: Boolean = true
+
+  def computeScoreForKey(c: AbstractColumn, key: Option[Pattern[Option[Occurrence]]]): Double = {
+    var result = 0.0
+    val matrixColumn = c.asInstanceOf[MatrixColumn]
+    val matrix = matrixColumn.matrix
+    val colIx = matrixColumn.colIx
+    matrix.columns.updated(0, matrix.columns(colIx)).updated(colIx, matrix.columns(0))
```
dwightguth: not necessary. we don't actually implement "swaps" in the paper as actual swaps.

matching/src/main/scala/org/kframework/backend/llvm/matching/Heuristics.scala
```diff
@@ -149,6 +149,31 @@ object AHeuristic extends Heuristic {
...
+
+    for (con <- matrix.sigma) {
```
dwightguth: sigma calls bestColIx as well, so you should call matrixColumn.column.signature.

matching/src/main/scala/org/kframework/backend/llvm/matching/Heuristics.scala
```diff
@@ -149,6 +149,31 @@ object AHeuristic extends Heuristic {
...
+      val spec = matrix.specialize(con, colIx)._2
+      if (spec.bestRowIx != -1) {
+        result += 1.0
+      }
+    }
+    if (matrix.default(colIx).isDefined) {
```
dwightguth: let's cache this call too

PR #179, opened 2019-07-09
adds rowUseless and notCol

mickyabir:


matching/src/main/scala/org/kframework/backend/llvm/matching/Matrix.scala
```diff
@@ -549,6 +549,42 @@ class Matrix private(val symlib: Parser.SymLib, private val rawColumns: IndexedS
...
+      }
+    }
+
+    if (!r.patterns(0).isWildcard) {
+      val con = r.patterns(0).signature(r.clause).head
+      val rowSpec = Matrix.fromRows(symlib, IndexedSeq(r), fringe).specialize(con, 0)._2.rows.head
+      specialize(con, 0)._2.useless(rowSpec)
+    } else {
+      if (columns(0).category.hasIncompleteSignature(columns(0).signature, columns(0).fringe)) {
+        default(0).isDefined && default(0).get.useless(new Row(r.patterns.drop(1), r.clause))
```
dwightguth: let's do the same thing we did above and create a Matrix from the row and call default on it. That is more likely to be correct in all cases than r.patterns.drop(1).

matching/src/main/scala/org/kframework/backend/llvm/matching/Matrix.scala
```diff
@@ -549,6 +549,42 @@ class Matrix private(val symlib: Parser.SymLib, private val rawColumns: IndexedS
...
+          val rowSpec = Matrix.fromRows(symlib, IndexedSeq(r), fringe).specialize(con, 0)._2.rows.head
+          if (specialize(con, 0)._2.useless(rowSpec)) {
+            true
+          }
+        }
+        false
+      }
+    }
+
+    false
```
dwightguth: this should not be necessary as all the above code paths should have returned a value. with this line here, the function is always going to return false.

matching/src/main/scala/org/kframework/backend/llvm/matching/Matrix.scala
```diff
@@ -549,6 +549,42 @@ class Matrix private(val symlib: Parser.SymLib, private val rawColumns: IndexedS
...
+      val rowSpec = Matrix.fromRows(symlib, IndexedSeq(r), fringe).specialize(con, 0)._2.rows.head
+      specialize(con, 0)._2.useless(rowSpec)
+    } else {
+      if (columns(0).category.hasIncompleteSignature(columns(0).signature, columns(0).fringe)) {
+        default(0).isDefined && default(0).get.useless(new Row(r.patterns.drop(1), r.clause))
+      } else {
+        for (con <- columns(0).signature) {
+          val rowSpec = Matrix.fromRows(symlib, IndexedSeq(r), fringe).specialize(con, 0)._2.rows.head
+          if (specialize(con, 0)._2.useless(rowSpec)) {
+            true
```
dwightguth: please add the return keyword here.

matching/src/main/scala/org/kframework/backend/llvm/matching/Matrix.scala
```diff
@@ -549,6 +549,42 @@ class Matrix private(val symlib: Parser.SymLib, private val rawColumns: IndexedS
...
+      if (rows.size > 0) {
+        false
+      } else {
+        true
+      }
+    }
+
+    if (!r.patterns(0).isWildcard) {
+      val con = r.patterns(0).signature(r.clause).head
+      val rowSpec = Matrix.fromRows(symlib, IndexedSeq(r), fringe).specialize(con, 0)._2.rows.head
```
dwightguth: let's add a specialize function to Row.

matching/src/main/scala/org/kframework/backend/llvm/matching/Matrix.scala
```diff
@@ -549,6 +549,42 @@ class Matrix private(val symlib: Parser.SymLib, private val rawColumns: IndexedS
...
+        }
+        false
+      }
+    }
+
+    false
+  }
+
+  def rowUseless(rowIx: Int): Boolean = {
+    val matrix = Matrix.fromRows(symlib, rows.take(rowIx) ++ rows.drop(rowIx + 1), fringe)
```
dwightguth: remember that we only want to take the rows after rowIx which have the same priority as rowIx. You can test this with clause.action.priority.

matching/src/main/scala/org/kframework/backend/llvm/matching/Matrix.scala
dwightguth: It should probably take a Constructor and a column index and return an Option[Row] which is None if the row was deleted during the specialization.

matching/src/main/scala/org/kframework/backend/llvm/matching/Matrix.scala
```diff
@@ -549,6 +557,36 @@ class Matrix private(val symlib: Parser.SymLib, private val rawColumns: IndexedS
...
+        if (rowSpec.isDefined && specialize(con, 0)._2.useful(rowSpec.get)) {
+          return true
+        }
+      }
+      false
+    }
+  }
+
+  def rowUseless(rowIx: Int): Boolean = {
+    val filteredRows = (rows.take(rowIx) ++ rows.drop(rowIx + 1)).filter(_.clause.action.priority <= rows(rowIx).clause.action.priority)
```
dwightguth: let's do rows.drop(rowIx + 1).takeWhile(_.clause.action.priority == rows(rowIx).clause.action.priority)

PR #182, opened 2019-07-09
n heuristic

mickyabir:


matching/src/main/scala/org/kframework/backend/llvm/matching/Heuristics.scala
```diff
@@ -208,6 +208,19 @@ object RHeuristic extends Heuristic {
...
 
+@NamedHeuristic(name='n')
+object NHeuristic extends Heuristic {
+  val needsMatrix: Boolean = true
+
+  def computeScoreForKey(c: AbstractColumn, key: Option[Pattern[Option[Occurrence]]]): Double = {
+    val matrixColumn = c.asInstanceOf[MatrixColumn]
+    val matrix = matrixColumn.matrix
+    val colIx = matrixColumn.colIx
+    matrix.rows.indices.filter(matrix.necessary(_, colIx)).size
```
dwightguth: I actually meant `matrix.rows.indices.count(matrix.necessary(_, colIx))`

PR #184, opened 2019-07-10
install-rust: respect CARGO_HOME and RUSTUP_HOME variables

ehildenb:


ehildenb:
This maintains the old behavior of defaulting to `$HOME/.cargo`, but allows (correctly) changing it by setting `CARGO_HOME`, to allow for local builds.

ehildenb:
Should wait on https://github.com/kframework/k/pull/614

PR #188, opened 2019-07-12
Iterated pattern matching optimization

dwightguth:
It's finally ready!


CMakeLists.txt
```diff
@@ -101,9 +101,14 @@ install(
 
 find_program(OPT opt-8)
 find_program(OPT opt)
+find_program(LLC llc-8)
+find_program(LLC llc)
 if(${OPT} STREQUAL "OPT-NOTFOUND")
```
charala1: This can be simplified to `if ( NOT OPT )`

PR #189, opened 2019-07-18
Misc bug fixes

dwightguth:
Includes:
* bug fix where llc was undesirably optimizing by default
* bug fix where we were overflowing the end of a memory block because we were writing sentinels to the NoGC space which is not 8-byte aligned
* Bug fix to decision tree generation that fixes wasm bug
* Fix bug where our hash algorithm for lists was broken the same way the set and map algorithm was previously.

### DIFF
--- a/include/runtime/alloc.h
+++ b/include/runtime/alloc.h
@@ -9,7 +9,10 @@ extern "C" {
 // A contiguous area larger than that size cannot be allocated in any arena.
 extern const size_t BLOCK_SIZE;
 
+#define YOUNGSPACE_ID 0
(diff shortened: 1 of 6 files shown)


include/runtime/alloc.h
```diff
@@ -9,6 +9,8 @@ extern "C" {
 // A contiguous area larger than that size cannot be allocated in any arena.
 extern const size_t BLOCK_SIZE;
 
+#define NOGCSPACE_ID 2
```
theo25: I t would be nice to add macros for the rest of the spaces here for uniformity.

PR #192, opened 2019-07-24
Support llvm 7.0.1

dwightguth:
It looks like the bug in llvm 7 that broke the llvm backend was fixed in llvm 7.0.1, which, while not available on Ubuntu 18.04, is available in Debian Buster, whereas llvm 8 is not. So I'm going to add support for it here.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,12 +1,16 @@
 cmake_minimum_required (VERSION 3.4)
 
 find_program(CMAKE_C_COMPILER clang-8)
+find_program(CMAKE_C_COMPILER clang-7)
 find_program(CMAKE_C_COMPILER clang)
 find_program(CMAKE_CXX_COMPILER clang++-8)
+find_program(CMAKE_CXX_COMPILER clang++-7)
 find_program(CMAKE_CXX_COMPILER clang++)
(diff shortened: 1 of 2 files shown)


charala1:
> It looks like the bug in llvm 7 that broke the llvm backend was fixed in llvm 7.0.1, which, while not available on Ubuntu 18.04, is available in Debian Buster, whereas llvm 8 is not. So I'm going to add support for it here.

LLVM 8 is available on debian buster:
```
deb http://apt.llvm.org/buster/ llvm-toolchain-buster-8 main
deb-src http://apt.llvm.org/buster/ llvm-toolchain-buster-8 main
```
add the key: `wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key|sudo apt-key add -`

see <https://apt.llvm.org/>

dwightguth:
Sure, of course you can always add a third party package repository, but we should not be asking our users to do this if we can avoid it because it adds an extra step to the installation process.

charala1:
Well yes, but this is the official release. Maybe the benefits outweigh the
cost in this case.

dwightguth:
We didn't use llvm 7 because it made the backend segfault. It looks like this was fixed in 7.0.1, so we might as well just use 7.0.1, since it's the version in buster by default.

PR #203, opened 2019-08-02
Fixes +Infinity ==Float -Infinity bug

mickyabir:


mickyabir:
If the error message isn't informative enough, I can change it

lib/codegen/CreateTerm.cpp
```diff
@@ -245,9 +245,9 @@ llvm::Value *CreateTerm::createToken(ValueType sort, std::string contents) {
     llvm::Constant *global = Module->getOrInsertGlobal("float_" + contents, Module->getTypeByName(FLOAT_WRAPPER_STRUCT));
     llvm::GlobalVariable *globalVar = llvm::dyn_cast<llvm::GlobalVariable>(global);
     if (!globalVar->hasInitializer()) {
-      size_t is_float = contents.find_first_of("fF");
       size_t prec, exp;
-      if (is_float != std::string::npos) {
+      const char last = contents.back();
+      if (last == 'f' || last == 'F' || last == 'y') {
```
mickyabir: I'm open to a cleaner solution, but this essentially added the case when the string is just `+Infinity` or `-Infinity`

runtime/strings/strings.cpp
```diff
@@ -345,9 +345,9 @@ extern "C" {
 }
 
 void init_float2(floating *result, std::string contents) {
-  size_t is_float = contents.find_first_of("fF");
   size_t prec, exp;
-  if (is_float != std::string::npos) {
+  const char last = contents.back();
+  if (last == 'f' || last == 'F' || last == 'y') {
```
mickyabir: see above

PR #207, opened 2019-08-05
fixes +123 int bug in rules

mickyabir:


lib/codegen/EmitConfigParser.cpp
```diff
@@ -408,6 +408,20 @@ static void emitGetToken(KOREDefinition *definition, llvm::Module *module) {
       break;
     }
     case SortCategory::Int: {
+      llvm::Value *FirstChar = new llvm::LoadInst(llvm::Type::getInt8Ty(Ctx), func->arg_begin()+2, "", CaseBlock);
+      llvm::Constant *asciiPlus = llvm::ConstantInt::get(llvm::Type::getInt8Ty(Ctx), 43);
```
dwightguth: let's make this '+', not 43.

PR #219, opened 2019-08-19
Including modules actually includes their hooks

mickyabir:


mickyabir:
The kreflection hooks will be moved if necessary in the upcoming PR that will implement them

PR #225, opened 2019-08-31
Use immer instead of rust

dwightguth:
Removes rust from the codebase and replaces references everywhere to the C++ immer library. I tested the performance against KEVM and saw a roughly 25% speedup in the codebase resulting from this. All tests pass as well.

@chathhorn can you please pull the latest changes of this branch and test that there are no regressions against the rv-match codebase?


chathhorn:
I tested f0f852f with the C semantics unit tests and don't see any regressions.

runtime/meta/substitution.cpp
```diff
@@ -3,6 +3,7 @@
 #include "runtime/header.h"
 #include "runtime/alloc.h"
 
+
```
mickyabir: am I missing something here?

unittests/runtime-collections/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+add_kllvm_unittest(runtime-collections-tests
+  lists.cpp
+  maps.cpp
+  main.cpp
```
mickyabir: what about the `sets` tests?

mickyabir:
Approved but please look at the changes I made

PR #235, opened 2019-09-10
KOREScanner.l: Admit @ in identifiers for set variables

ttuegel:
In kframework/kore#922, we begin to use set variables in `prelude.kore`, which requires this small extension to the C++ Kore parser.


dwightguth:
Is @0 a valid set variable? It would be permitted by this regex.

ttuegel:
Hm, if we take the Haskell parser to be the reference implementation, then that is not valid. (The document describing the syntax of Kore has fallen out of date.)

PR #250, opened 2019-10-09
K_COVERAGEDIR Environment Variable

gtrepta:


dwightguth:
We don't have tests for coverage, so please make sure to test this manually before merging.

gtrepta:
> We don't have tests for coverage, so please make sure to test this manually before merging.

It wont create the file if the directory doesn't exist, if that's ok then everything else works fine.

PR #251, opened 2019-10-09
Recursive descent parser and reentrant scanner

dwightguth:
This is the second of two major refactorings to prepare the way for removing the remaining memory leaks in the LLVM backend. Here we remove our build-time dependency on bison and replace the bison parser with a recursive descent parser. We continue to use flex, but we add the reentrant option to flex so that all memory allocated by flex will be cleaned up when the scanner finishes.

The purpose of this refactoring is to allow us in the next PR to replace use of raw pointers in the AST and Parser libraries with smart pointers. std::unique_ptr has a deleted copy constructor and bison versions prior to Bison 3.2 do not support move-only types in semantic values. However, bison 3.2 is not available on Ubuntu 18.04 LTS, so we simply remove our bison dependency instead.


lib/parser/KOREScanner.l
```diff
@@ -97,51 +97,48 @@ yyin = in;
...
 %%
 
-KOREScanner::KOREScanner(const char *filename) {
-  if (!(in = fopen(filename, "r"))) {
+KOREScanner::KOREScanner(std::string filename) {
+  if (!(in = fopen(filename.c_str(), "r"))) {
     std::cerr << "Cannot read file: " << filename << "\n";
     exit(1);
   }
+  yylex_init(&scanner);
```
theo25: I am a little confused as to how the parser state is getting initialized here. This `scanner` variable is just a `void *` right?

dwightguth: yylex_init allocates some storage and then writes the address of that storage to this pointer.

theo25: OK I see. I think I get it now.

PR #271, opened 2020-01-07
Better error message when semantics is used as a library and function gets stuck

dwightguth:
@ehildenb can you try this out and let me know if this helps you?


ehildenb:
This helps a bit (pin-pointing the problematic term), but it seems that it does not print the term correctly (missing some arguments). Either way it's definitely an improvement, so if you want to merge we can fix it more later.

PR #277, opened 2020-01-13
always keep sort_table symbol for debugging

dwightguth:


radumereuta:
This hasn't really fixed the issue.
If I kompile normally, I don't get `gdb.error: Missing ELF symbol "sort_table"`
But I get `<optimized out>` terms.
If I use `-O0` or `-O1` the error is back.

dwightguth:
I don't know what to tell you; I was no longer able to reproduce this issue with this change. Are you sure you're using the correct version of the llvm backend?

PR #281, opened 2020-01-16
Fix bugs involving substitution

dwightguth:
Normally I would ask Brandon to review this, but since he's not doing well with health stuff, we'll have to make do. Essentially, I fixed a few minor bugs in the gdb pretty printing of variables that hadn't been uncovered because we had never tested that code path before. I also fixed two major bugs in the substitution algorithm:

1. a bug made it possible for two different lambdas in the same term to have the same address as their bound variable, thus causing both to be printed as the same variable despite actually being different variables.
2. a bug made it so that when a bound variable was substituted under a lambda, the debruijn index was not correctly incremented by one.


dwightguth:
@ehildenb I'd appreciate your review on this also since it's somewhat theoretically complex and I don't have Brandon to review.

runtime/meta/substitution.cpp
```diff
@@ -167,52 +172,81 @@ block *substituteInternal(block *currBlock) {
...
+      block *to_replace_stack = to_replace;
+      block *replacement_stack = replacement;
+      block *replacementInj_stack = replacementInj;
+      block *result = (block *)evaluateFunctionSymbol(tag, &arguments[0]);
+      to_replace = to_replace_stack;
+      replacement = replacement_stack;
+      replacementInj = replacementInj_stack;
+      idx = idx_stack;
+      return result;
+    }
```
ehildenb: Are we calling this here because after going through applying substitution to all sub-terms we could now have the result be evaluatable?

I guess I'm wondering how this was done before. Is the function evaluator being called after substitution anyway? If so, is this redundant?

ehildenb:
This looks reasonable to me, but I get lost in some of the details of how the LLVM backend is working I think. Debruijn indices seem to be incremented/decremented in sensible places, and I can't spot any obvious inconsistencies.

As long as the test which triggered finding this bug is added to the regression test-suite, it should be good.

runtime/meta/substitution.cpp
dwightguth: no, function evaluation is not called after substitution. previously we forgot to evaluate anywhere rules after applying substitution.

ehildenb: Sounds good then.

PR #288, opened 2020-01-27
Clean up choice code to fix bugs

dwightguth:
In the process of testing the llvm backend on the K tutorial, I uncovered a number of bugs associated with nondeterministic set and map choices.

Roughly, there were three bugs:
1. Phi nodes for variables that are needed on only one ancestor path of a shared node with multiple choice nodes as ancestors were not being computed correctly if that shared node only had a single immediate parent.
2. The choice to be jumped back to after an iterator for a choice runs out of elements was incorrect for nested choice nodes that are shared across multiple different parent choices.
3. If both children of a switch that checks whether the iterator of a choice has run out of elements have nested choice nodes, the code was not correctly computing the jump destination after entering the outer switch.

The fix for these problems was complexified by a fourth problem that was not immediately causing trouble, but needed to be addressed in order to cause a robust fix. Essentially, we were using only the occurrence to identify variables in decision tree functions, when in fact a shared node can have two different variables with the same occurrence and different types. We were correctly handling this case in the old implementation (with the exception of the above bugs, obviously), but we had to change the variable key to an occurrence/type tuple in order to make it unique so that we would be able to precompute the liveness of each variable in the function.

As a result, the fix for the above three issues roughly requires the following four changes:
1. The scala and llvm code was modified to treat occurrence/type tuples as the unique identifier of variables in decision tree functions and substitutions. This required annotating the yaml decision tree with additional information, hence the alteration to the scala code. In itself, this fixes none of the bugs, but it makes it possible to perform step 2.
2. We precompute the live-in variables for each of nodes in the decision tree using a minor variant of the standard liveness analysis algorithm. The reason for the variation is that if I have a switch whose children are a fail node and a choice node, and the choice node has another fail node as a decendent, we can statically know that the path from the outer failure node directly to the inner choice without passing through the other path of the switch is not feasible because of our knowledge of how the successors of the failure block are computed. This is still not a perfect representation of liveness in this modified context; I ran into a situation where variables were treated as live at the root of a decision tree that were not actually live there because there was a path to a shared node with a choice as an ancestor that statically depended on a particular variable in order to jump back to that choice, when the path to the root in question can be statically known not to pass through any choices, and thus those variables should not actually be live. However, the algorithm is able to solve this case with only a minor loss of code quality because it simply emits undefined values into the phi nodes for edges of the dataflow graph that the code generator does not have a value in the substitution for. Thus these incorrectly live variables can safely be ignored. There might be a minor loss to performance and code size as a result, but further improvements can be made if this becomes an issue.
3. Instead of assuming that the choice to jump back to when an runs out of elements can be statically known (it cannot because multiple choices may contain the same shared node as descendents), we maintain a stack of jump destinations in local memory for the function. We push onto the stack when an iterator has another element, and pop from the stack when local matching fails and it's time to try the next element. Because we can statically know for each function the maximum nesting depth of choices in that function, we simply use an integer and array alloca to represent the stack. the first element popped onto the stack is always the basic block that tells you that the function got stuck, and this element is popped off when the entire function gets stuck.
4. We fix the third bug above by means of creating a local variable in the SwitchNode::codegen function which stores the basic block of the current choice prior to evaluating any of the branches of the switch, and then refers to it after each branch, rather than relying on a class variable that might get clobbered by one of the children of the switch and thus contain an incorrect value in the next loop iteration.

I would like both @mickyabir and @theo25 to review because Theo has reviewed code for the choices in the past and this change is quite complex.

This code has been tested and it fixes the problems associated with executing IMP++ in the K tutorial.

### DIFF
--- a/lib/codegen/Decision.cpp
+++ b/lib/codegen/Decision.cpp
@@ -20,24 +20,22 @@ FailNode FailNode::instance;
 
 static unsigned max_name_length = 1024 - std::to_string(std::numeric_limits<unsigned long long>::max()).length();
 
-void Decision::operator()(DecisionNode *entry, llvm::StringMap<llvm::Value *> substitution) {
+void Decision::operator()(DecisionNode *entry, std::map<std::pair<std::string, llvm::Type *>, llvm::Value *> substitution) {
   if (entry == FailNode::get()) {
     llvm::BranchInst::Create(this->FailureBlock, this->CurrentBlock);
   } else {
     entry->codegen(this, substitution);
   }
   for (auto &entry : failPhis) {
-    for (auto &entry2 : entry.second) {
-      llvm::PHINode *Phi = entry2.second;
-      std::vector<llvm::BasicBlock *> undefBlocks;
-      for (llvm::BasicBlock *block : llvm::predecessors(Phi->getParent())) {
-        if (Phi->getBasicBlockIndex(block) == -1) {
-          undefBlocks.push_back(block);
-        }
+    llvm::PHINode *Phi = entry.second;
+    std::vector<llvm::BasicBlock *> undefBlocks;
+    for (llvm::BasicBlock *block : llvm::predecessors(Phi->getParent())) {
+      if (Phi->getBasicBlockIndex(block) == -1) {
+        undefBlocks.push_back(block);
       }
-      for (llvm::BasicBlock *block : undefBlocks) {
-        Phi->addIncoming(llvm::UndefValue::get(Phi->getType()), block);
-      }
+    }
+    for (llvm::BasicBlock *block : undefBlocks) {
+      Phi->addIncoming(llvm::UndefValue::get(Phi->getType()), block);
     }
   }
   for (auto block = this->CurrentBlock->getParent()->begin(); block != this->CurrentBlock->getParent()->end(); ++block) {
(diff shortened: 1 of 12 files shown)


lib/codegen/Decision.cpp
```diff
@@ -48,26 +54,41 @@ void Decision::operator()(DecisionNode *entry, std::map<std::pair<std::string, l
...
+    DecisionNode *node = *workList.begin();
+    workList.erase(node);
+    std::set<std::pair<std::string, llvm::Type *>> newVars;
+    for (DecisionNode *succ : node->successors) {
+      if (succ == FailNode::get()) {
+        for (DecisionNode *trueSucc : succ->successors) {
+          if (node->choiceAncestors.count(dynamic_cast<IterNextNode *>(trueSucc))) {
+            newVars.insert(trueSucc->vars.begin(), trueSucc->vars.end());
+	  }
+	}
```
theo25: nit: fix indentation

lib/codegen/Decision.cpp
```diff
@@ -205,38 +219,44 @@ void SwitchNode::codegen(Decision *d, std::map<std::pair<std::string, llvm::Type
...
-      if (d->ChoiceBlock && _case.getLiteral() == 1) {
-        new llvm::StoreInst(llvm::BlockAddress::get(d->CurrentBlock->getParent(), d->ChoiceBlock), d->FailAddress, d->CurrentBlock);
-        d->FailJump->addDestination(d->ChoiceBlock);
-        d->ChoiceBlock = nullptr;
+      if (currChoiceBlock && _case.getLiteral() == 1) {
+        auto PrevDepth = new llvm::LoadInst(d->ChoiceDepth, "", d->CurrentBlock);
+	auto CurrDepth = llvm::BinaryOperator::Create(llvm::Instruction::Add, PrevDepth, llvm::ConstantInt::get(llvm::Type::getInt64Ty(d->Ctx), 1), "", d->CurrentBlock);
+	new llvm::StoreInst(CurrDepth, d->ChoiceDepth, d->CurrentBlock);
+
+	auto ty = d->ChoiceBuffer->getType()->getElementType();
```
theo25: nit: fix indentation

lib/codegen/Decision.cpp
```diff
@@ -380,6 +383,33 @@ llvm::Value *Decision::getTag(llvm::Value *val) {
...
+  FailNode::get()->predecessors.clear();
+  FailNode::get()->successors.clear();
+  std::set<LeafNode *> leaves;
+  dt->preprocess(leaves);
+  dt->computeLiveness(leaves);
+  auto ty = llvm::ArrayType::get(llvm::Type::getInt8PtrTy(module->getContext()), dt->getChoiceDepth() + 1);
+  llvm::AllocaInst *choiceBuffer = new llvm::AllocaInst(ty, 0, "choiceBuffer", block);
+  llvm::AllocaInst *choiceDepth = new llvm::AllocaInst(llvm::Type::getInt64Ty(module->getContext()), 0, "choiceDepth", block);
+  new llvm::StoreInst(llvm::ConstantInt::get(llvm::Type::getInt64Ty(module->getContext()), 0), choiceDepth, block);
+  auto zero = llvm::ConstantInt::get(llvm::Type::getInt64Ty(module->getContext()), 0);
```
theo25: nit: you could define the `zero` constant above the previous line and use it there too

lib/codegen/Decision.cpp
```diff
@@ -78,7 +82,12 @@ void DecisionNode::computeLiveness(std::unordered_set<LeafNode *> &leaves) {
     node->eraseDefsAndAddUses(newVars);
     if (newVars != node->vars || !livenessVisited) {
       node->vars = newVars;
-      workList.insert(node->predecessors.begin(), node->predecessors.end());
+      for (DecisionNode *pred : node->predecessors) {
+        if (!workListSet.count(pred)) {
+          workListSet.insert(pred);
+	  workList.push_back(pred);
+	}
```
theo25: another indentation issue here

PR #297, opened 2020-02-17
Alias3

dwightguth:
Essentially we were calling shared_from_this on a unique_ptr by mistake


dwightguth:
The rationale is that attributes are not substituted and thus do not need to be stored as a shared pointer. There is an implicit conversion from unique_ptr to shared_ptr, but not the other way around.

PR #298, opened 2020-02-19
Fix another bug involving phi nodes and choices in decision trees

dwightguth:
the K build is failing on Mac OS because llvm 9 introduced some new optimizations that take advantage of undefined values in llvm IR, and this uncovered a bug in how we are computing the phi nodes for decision tree functions in the llvm backend. Essentially, we were not correctly computing the set of choice nodes that are ancestors of the current node, and that led to values getting assigned as undefined when really liveness should have treated them as live. This meant that the same stack slot was being reused for multiple different variables and their values were clobbering one another, leading to memory corruption in the substitution that was being generated by the decision tree.

Essentially, to fix this, instead of doing a DFS on the tree to compute the choice ancestors, we first do a topological sort and then iterate through in sorted order. This guarantees that each node's ancestors are computed after its predecessors ancestors are fully computed, which fixes the issue.

### DIFF
--- a/include/kllvm/codegen/Decision.h
+++ b/include/kllvm/codegen/Decision.h
@@ -50,6 +50,8 @@ public:
   bool beginNode(Decision *d, std::string name, std::map<var_type, llvm::Value *> &substitution);
   std::vector<DecisionNode *> topologicalSort(void);
 
+  static void computeChoiceAncestors(std::vector<DecisionNode *> sorted);
+
   void setCompleted() { completed = true; }
   bool isCompleted() const { return completed; }
   uint64_t getChoiceDepth() const { return choiceDepth; }
@@ -141,7 +143,6 @@ public:
     if(preprocessed) return;
     bool hasDefault = false;
     for (auto _case : cases) {
-      _case.getChild()->choiceAncestors.insert(choiceAncestors.begin(), choiceAncestors.end());
       _case.getChild()->preprocess(leaves);
       _case.getChild()->predecessors.insert(this);
       _case.getChild()->predecessors2.insert(this);
@@ -197,7 +198,6 @@ public:
   }
   virtual void preprocess(std::unordered_set<LeafNode *> &leaves) {
     if (preprocessed) return;
-    child->choiceAncestors.insert(choiceAncestors.begin(), choiceAncestors.end());
     child->preprocess(leaves);
     child->predecessors.insert(this);
     child->predecessors2.insert(this);
@@ -260,7 +260,6 @@ public:
   }
   virtual void preprocess(std::unordered_set<LeafNode *> &leaves) {
     if (preprocessed) return;
-    child->choiceAncestors.insert(choiceAncestors.begin(), choiceAncestors.end());
     child->preprocess(leaves);
     child->predecessors.insert(this);
     child->predecessors2.insert(this);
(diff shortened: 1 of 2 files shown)


lib/codegen/Decision.cpp
```diff
@@ -148,6 +148,28 @@ void DecisionNode::computeLiveness(std::unordered_set<LeafNode *> &leaves) {
...
 
+std::vector<DecisionNode *> DecisionNode::topologicalSort(void) {
+  std::vector<DecisionNode *> result;
+  std::vector<DecisionNode *> workList;
+  workList.push_back(this);
+  while (!workList.empty()) {
+    DecisionNode *node = workList.back();
+    workList.pop_back();
+    result.push_back(node);
+    auto it = node->successors2.begin();
```
virgil-serbanuta: Why use `it` in the loop below and not just `node->successors2.begin()`? Because it's shorter?

lib/codegen/Decision.cpp
```diff
@@ -466,12 +488,30 @@ llvm::Value *Decision::getTag(llvm::Value *val) {
   return res;
 }
 
+void DecisionNode::computeChoiceAncestors(std::vector<DecisionNode *> sorted) {
```
virgil-serbanuta: I didn't write much C++ code in the last 4 years, but wouldn't this be better with a `const&` argument?

include/kllvm/codegen/Decision.h
```diff
@@ -191,10 +198,11 @@ class MakePatternNode : public DecisionNode {
   }
   virtual void preprocess(std::unordered_set<LeafNode *> &leaves) {
     if (preprocessed) return;
-    child->choiceAncestors.insert(choiceAncestors.begin(), choiceAncestors.end());
     child->preprocess(leaves);
     child->predecessors.insert(this);
+    child->predecessors2.insert(this);
```
virgil-serbanuta: Is there a reason to compute these incrementally instead of copying the set before sorting?

lib/codegen/Decision.cpp
dwightguth: As I recall I have observed experimentally that it is faster to reuse the iterator that is returned by `erase` rather than to call `begin` again, when you are iterating through a set and deleting every element as you reach it. Moreover, this is essentially a simplification of the more general pattern of looping through a data structure and possibly deleting elements as you go, where you need an iterator because it will not necessarily always point to the beginning of the collection, as it does in this case because we are deleting every element.

Thus I view this code as more idiomatic and faster than repeatedly calling `begin`. Not to mention, the iterator is computed by the `erase` method anyway so it's duplicating work not to use it.

include/kllvm/codegen/Decision.h
dwightguth: Well, for one thing, predecessors2 is the predecessors of the dag that results from removing the outgoing edges of the failure node, so if we we used the same set, it would require more complicated logic. But also, since these sets will both become empty after `topologicalSort` is called, the difference in memory cost is at most linear in the number of nodes in the graph, and thus is relatively minor compared to the preexisting cost of representing the entire decision tree in memory. I'd rather write simpler logic now and worry about memory usage if we have examples where it is a problem. And it's not a matter of copying just one set prior to sorting because there is one set for each node. We could probably maintain a map from each node to the copied set of predecessors/successors as a local variable to the sort function, but if we're talking about storing one set per node, as a field on the node is the more object-oriented way to go about that. Plus, this way the performance cost is a pointer dereference, whereas if it were a map, the cost would include hashing a pointer and then doing equality in a bucket. When decision trees get large, the liveness analysis already takes a long time, so I'd rather keep things fast even if it requires slightly more memory, unless we reach a point where that becomes a problem, which we haven't yet.

PR #301, opened 2020-02-25
remove a bunch of stuff and just use alloca + mem2reg

dwightguth:
The liveness analysis we were doing was proving to be very slow, so I went digging into the details of how SSA is usually constructed, and realized that it's usually done based on dominance frontiers rather than data flow equations. I was going to implement all of that, but then I realized we were probably better off just relying on the mem2reg llvm transformation pass. This turns out to delete a bunch of code /and/ be much much faster, while being much less likely to have bugs. So it's win/win/win.

### DIFF
--- a/lib/codegen/Decision.cpp
+++ b/lib/codegen/Decision.cpp
@@ -20,189 +20,17 @@ FailNode FailNode::instance;
 
 static unsigned max_name_length = 1024 - std::to_string(std::numeric_limits<unsigned long long>::max()).length();
 
-void Decision::operator()(DecisionNode *entry, std::map<std::pair<std::string, llvm::Type *>, llvm::Value *> substitution) {
+void Decision::operator()(DecisionNode *entry) {
   if (entry == FailNode::get()) {
     llvm::BranchInst::Create(this->FailureBlock, this->CurrentBlock);
   } else {
-    if (entry->containsFailNode) {
-      for (auto &entry : FailNode::get()->vars) {
-        auto &var = entry.first;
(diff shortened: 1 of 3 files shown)


lib/codegen/Decision.cpp
```diff
@@ -580,20 +361,20 @@ void makeEvalOrAnywhereFunction(KORESymbol *function, KOREDefinition *definition
...
-  addStuck(stuck, module, function, subst, definition);
+  addStuck(stuck, module, function, codegen, definition);
 
-  Decision codegen(definition, block, fail, jump, choiceBuffer, choiceDepth, module, returnSort);
-  codegen(dt, subst);
+  codegen(dt);
 }
 
-void abortWhenStuck(llvm::BasicBlock *CurrentBlock, llvm::Module *Module, KORESymbol *symbol, std::map<std::pair<std::string, llvm::Type *>, llvm::Value *> &subst, KOREDefinition *d) {
+void abortWhenStuck(llvm::BasicBlock *CurrentBlock, llvm::Module *Module, KORESymbol *symbol, Decision &codegen, KOREDefinition *d) {
```
virgil-serbanuta: There may not be a special reason, but I'm curious: Why is codegen sent by reference, while the others are pointers?

dwightguth: just convention. A bunch of llvm functions take and return pointers to llvm ASTs, so that's why the llvm namespace objects are pointers. We use smart pointers to keep track of the KORE ASTs, so that's why the KORE AST nodes are pointers. I just used a reference for Decision because it seemed simpler, given that we didn't have any compelling reason to use a pointer in that case.

PR #310, opened 2020-03-20
Profile to generate and analyze gc statistics

dwightguth:
Essentially, we create a new cmake build profile, GcStats, which makes some minor modifications to garbage collection so as to emit on standard error a log of how much data is live at each collection broken down by its age. We also introduce a new tool that analyzes this log and can emit certain information useful for GC tuning.

This is not ready for review yet as I am still gathering data and may make further modifications to the code, but I'm creating this PR so as to facilitate easier testing since we now have profiles with significantly different code due to preprocessor macros.


include/runtime/collect.h
```diff
@@ -36,4 +37,33 @@ extern "C" {
...
+#endif
+
+#define initialize_migrate() \
+  bool isInYoungGen = is_in_young_gen_hdr(hdr); \
+  initialize_age() \
+  bool isInOldGen = is_in_old_gen_hdr(hdr); \
+  if (!(isInYoungGen || (isInOldGen && collect_old))) { \
+    return; \
+  } \
+  bool shouldPromote = isInYoungGen && age; \
```
virgil-serbanuta: I'm trying to understand how this works. `age` will count how many times something was migrated, right? If so, then I guess that this line says "shouldPromote= is young and was migrated at least once"...

tools/llvm-kompile-gc-stats/main.cpp
```diff
@@ -0,0 +1,95 @@
...
+        if (i > 0) {
+          assert(mpz_cmp_ui(total[i-1], frame[i]) >= 0);
+          mpz_sub_ui(total[i-1], total[i-1], frame[i]);
+        }
+      }
+    } else if (generation) {
+      mpz_set_ui(size, 0);
+      for (int i = lowerBound; i <= upperBound; i++) {
+        if (i == 0) {
+          mpz_add_ui(size, size, frame[0]);
```
virgil-serbanuta: Why is the `i==0` case special?

include/runtime/collect.h
dwightguth: Correct.

tools/llvm-kompile-gc-stats/main.cpp
dwightguth: Because of the data format, frame[0] contains the total number of allocated bytes rather than the total number of bytes that survived exactly 0 collection cycles. This is why it is treated specially.

PR #312, opened 2020-04-01
Fix memory corruption when initial configuration has a List in it

dwightguth:
This is a fairly straightforward memory corruption issue in the llvm backend. When the initial configuration has a non-hooked function that returns a term of sort List, it was storing a pointer to that list to a list** object allocated by malloc and then accidentally treating that list** as if it was a list*. This obviously led to memory corruption. This is left over from the days when we were trying to return structs by value from functions in the llvm backend. However, all structs are returned as pointers now in the llvm backend, which means that this code had become out of date and we didn't realize because there were no tests covering this case. I added a test to cover the case when the initial configuration has a function that returns sort List, and fixed the code so that it directly returns the list* pointer without performing an additional allocation.

### DIFF
--- a/lib/codegen/EmitConfigParser.cpp
+++ b/lib/codegen/EmitConfigParser.cpp
@@ -263,7 +263,6 @@ static std::pair<llvm::Value *, llvm::BasicBlock *> getEval(KOREDefinition *def,
     subst.insert({name, arg});
     pattern->addArgument(KOREVariablePattern::Create(name, sort));
   }
-  KORESymbolDeclaration *symbolDecl = def->getSymbolDeclarations().at(symbol->getName());
   CreateTerm creator(subst, def, CaseBlock, mod, false);
   llvm::Value *result = creator(pattern.get()).first;
   llvm::Value *retval;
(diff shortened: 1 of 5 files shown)


lib/codegen/EmitConfigParser.cpp
```diff
@@ -280,14 +279,9 @@ static std::pair<llvm::Value *, llvm::BasicBlock *> getEval(KOREDefinition *def,
...
   case SortCategory::Set:
-    if (symbolDecl->getAttributes().count("hook")) {
-      // if this is a hook then we are using the sret abi and the reutrned value
-      // is a pointer. Otherwise we need to store it like Bool/MInt
-      retval = new llvm::BitCastInst(result, llvm::Type::getInt8PtrTy(Ctx), "",
-          creator.getCurrentBlock());
-      break;
-    }
-    // fall through
+    retval = new llvm::BitCastInst(result, llvm::Type::getInt8PtrTy(Ctx), "",
```
virgil-serbanuta, line 282: Is there any difference now between Map/List/Set types and the Int/Float/... types above?

PR #313, opened 2020-04-08
fix p and n heuristic

dwightguth:
We accidentally misrepresented the `necessary` function by checking that a row was not useless when we should have been checking that it *was* useless. This was seriously nerfing the p and n heuristics and making them compute the wrong thing. I tested the modified p heuristic on the tezos semantics and while it takes longer than the q heuristic, the resulting decision tree is not any larger and has a slightly smaller average path length.


dwightguth:
It's difficult to test something like this because if the code were to be incorrect, it's more likely to cause a performance regression than any kind of easily detectable failure.

PR #314, opened 2020-04-10
Better dependency management when building K definitions with cmake

dwightguth:
This basically adds support for `llvm-kompile -c` which will emit a `.o` file for the definition that can be linked with other files with cmake to generate the final interpreter. We then add a second version of the LLVMKompile cmake script that takes advantage of this to avoid rerunning the backend when definition.kore is unchanged. This significantly speeds up the incremental build time when modifying files that link together with a k definition, like custom hooks.


dwightguth:
Backwards compatibility.

PR #315, opened 2020-04-10
llvm backend pretty printing tool

dwightguth:
This is the llvm backend part of a pair of PRs used to support pretty-printing kore terms with a vastly reduced overhead. Some logic is not implemented yet, but it is sufficient to unparse some simple IMP programs. I will add more features as I go if this proves useful to people.

Second PR is here: https://github.com/kframework/k/pull/1211

### DIFF
--- a/bin/llvm-krun
+++ b/bin/llvm-krun
@@ -3,13 +3,16 @@ set -e
 
 ARGV=()
 params=()
+pretty=()
 output_file="$(mktemp tmp.out.XXXXXXXXXX)"
 input_file="$(mktemp tmp.in.XXXXXXXXXX)"
-trap "rm -rf $input_file $output_file" INT TERM EXIT
+parser_file="$(mktemp tmp.parse.XXXXXXXXXX)"
+trap 'rm -rf "$input_file" "$output_file" "$parser_file"' INT TERM EXIT
 initializer="LblinitGeneratedTopCell{}"
 dir=.
 debug=
 depth=-1
+pretty_print=false
 
 while [[ $# -gt 0 ]]
 do
@@ -19,9 +22,12 @@ do
     name="$2"
     value="$3"
     var_name="params_$name"
+    pretty_name="pretty_$name"
+    sort_name="sort_$name"
     params+=("$name")
     sort="$4"
     type="$5"
+    printf -v "$sort_name" %s "$sort"
     case $type in
       kore)
       printf -v "$var_name" %s "$value"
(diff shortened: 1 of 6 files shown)


lib/ast/AST.cpp
```diff
@@ -285,6 +286,512 @@ sptr<KOREPattern> KORECompositePattern::expandAliases(KOREDefinition *def) {
...
+            indent++;
+            localIndent++;
+            break;
+          case 'd':
+            indent--;
+            localIndent--;
+            break;
+          case 'c':
+            if (data.colors.count(name)) {
+              color(out, data.colors.at(name)[localColor++]);
```
virgil-serbanuta, line 611: I don't understand this. Why is `localColor` incremented here? Could it happen that it is incremented past the end of that vector?

lib/ast/AST.cpp
```diff
@@ -285,6 +286,512 @@ sptr<KOREPattern> KORECompositePattern::expandAliases(KOREDefinition *def) {
...
+            }
+            break;
+          case 'r':
+            if (hasColor) {
+              append(out, "\x1b[0m");
```
virgil-serbanuta, line 616: Could you add a constant for this (e.g. `RESET`)?

tools/kprint/main.cpp
```diff
@@ -0,0 +1,71 @@
...
+  std::map<std::string, std::string> formats;
+  formats["kseq"] = "%1 ~> %2";
+  formats["dotk"] = ".";
+  formats["inj"] = "%1";
+  std::map<std::string, std::string> hooks;
+  std::set<std::string> assocs;
+  std::set<std::string> comms;
+  std::map<std::string, std::vector<std::string>> colors;
+
+  std::ifstream iformats(argv[1] + std::string("/format-att.txt"));
```
virgil-serbanuta, line 36: Who is supposed to write these files? The frontend, the user, or something else? I think I could infer what most of them mean, but, although I could follow what happens with that data, I don't think I actually understand `color-att.txt`.

bin/llvm-krun
```diff
@@ -3,13 +3,16 @@ set -e
```
virgil-serbanuta, line 3: Could you give a command line example for this script?

lib/ast/AST.cpp
dwightguth: That would only happen if the user incorrectly specifies their `format` or `colors` attributes.

tools/kprint/main.cpp
dwightguth: It is generated by the frontend; see the accompanying PR.

virgil-serbanuta: Ok, but why does it make sense to have multiple colors for the same thing, and then to use each of them once?

dwightguth: well, you may want to color different parts of the same production differently. consider for example the following production:
```
syntax Stmt ::= "if" "(" Stmt ")" Block "else" Block
```

Presumably you may want to color the parentheses differently than the keywords.

PR #318, opened 2020-04-22
Even better gdb command support

dwightguth:
adds custom gdb commands to step, start, and attempt to match a particular rule.  The latter will tell you why the match fails as well as telling you all the hooks that are called as part of the matching process with their arguments and result. All of this will be pretty printed using gdb.


runtime/configurationparser/ConfigurationPrinter.cpp
```diff
@@ -216,6 +216,66 @@ extern "C" void* getStderr(void) {
...
+void resetMatchReason(void) {
+  matchLog.clear();
+}
+
+MatchLog *getMatchLog(void) {
+  return &matchLog[0];
+}
+
+void **getMatchFnArgs(MatchLog *log) {
+  return &log->args[0];;;;
```
virgil-serbanuta, line 245: I'm curious: Why so many semicolons?

gdb/kllvm.py
```diff
@@ -630,19 +640,144 @@ def append(self, subject, isVar, sort):
...
+    def invoke(self, arg, from_tty):
+        try:
+            argv = gdb.string_to_argv(arg)
+            gdb.lookup_global_symbol("resetMatchReason").value()()
+            gdb.lookup_global_symbol(argv[0] + '.match').value()(gdb.parse_and_eval(argv[1]))
+            entries = gdb.lookup_global_symbol("getMatchLog").value()()
+            size = int(gdb.lookup_global_symbol("getMatchLogSize").value()())
+            for i in range(size):
+                entry = entries[i]
+                if entry['kind'] == 0:
```
virgil-serbanuta, line 744: The c++ code uses an enum for these values, some python constants with the same name would be nice.

gdb/kllvm.py
```diff
@@ -630,19 +640,144 @@ def append(self, subject, isVar, sort):
...
-        elif t.target().tag == "__mpz_struct":
-            return termPrinter(val, "int", val.type.name)
-        elif t.target().tag == "floating":
-            return termPrinter(val, "floating", val.type.name)
+    if t.code != gdb.TYPE_CODE_PTR:
+        if t.tag:
+            kind = t.tag
+        else:
+            kind = t.name
+        if kind[0:11] == 'immer::list':
```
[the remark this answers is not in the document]
dwightguth: it's a namespace that refers to the name of our data structure library.

runtime/configurationparser/ConfigurationPrinter.cpp
dwightguth: My editor sometimes does this by mistake and I didn't spot this one.

PR #323, opened 2020-04-27
include nine in list of escape digits

dwightguth:
This was an off by one error that causes problems when the format attribute of a production references the ninth nonterminal.

### DIFF
--- a/lib/ast/AST.cpp
+++ b/lib/ast/AST.cpp
@@ -692,7 +692,7 @@ void KORECompositePattern::prettyPrint(std::ostream &out, PrettyPrintData const&
           case '8':
           case '9': {
             std::string buf;
-            for(; i < format.length() && format[i] >= '0' && format[i] < '9'; i++) {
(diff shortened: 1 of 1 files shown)


lib/ast/AST.cpp
```diff
@@ -692,7 +692,7 @@ void KORECompositePattern::prettyPrint(std::ostream &out, PrettyPrintData const&
           case '8':
           case '9': {
             std::string buf;
-            for(; i < format.length() && format[i] >= '0' && format[i] < '9'; i++) {
+            for(; i < format.length() && format[i] >= '0' && format[i] <= '9'; i++) {
```
ehildenb, line 695: Do we only handle productions with arguments up to length 10?

dwightguth: No. This was a bug only with numbers containing the number 9 in their decimal expansion.

dwightguth: The fix should make it work for any arity, although there is another fix that will have to be included in the front end pr

PR #324, opened 2020-05-01
fix non-exhaustiveness bug

dwightguth:
We will not treat a constructor that is an injection as if it makes the
match non-exhaustive if the sort injected into has no productions that
are not injections.

### DIFF
--- a/matching/src/main/scala/org/kframework/backend/llvm/matching/Matrix.scala
+++ b/matching/src/main/scala/org/kframework/backend/llvm/matching/Matrix.scala
@@ -659,7 +659,7 @@ class Matrix private(val symlib: Parser.SymLib, private val rawColumns: IndexedS
       }
     } else {
       val sigma = columns(0).signatureForUsefulness
-      if (r.patterns(0).isWildcard && columns(0).category.hasIncompleteSignature(sigma, columns(0).fringe)) {
+      if (r.patterns(0).isWildcard && columns(0).fringe.sortInfo.category.hasIncompleteSignature(sigma, columns(0).fringe)) {
         val matrixDefault = default(0, sigma)
         val rowDefault = r.default(0, sigma, symlib, fringe, columns)
         matrixDefault.isDefined && rowDefault.isDefined && matrixDefault.get.useful(rowDefault.get)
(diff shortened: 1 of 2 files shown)


ehildenb:
@dwightguth maybe add "Fixes: kframework/k#1221", unless you intend the PR into K that fixes it to be the one that you do that with.

dwightguth:
It's better to let the issue be closed by the front end pr

matching/src/main/scala/org/kframework/backend/llvm/matching/Matrix.scala
```diff
@@ -262,6 +262,7 @@ case class Fringe(val symlib: Parser.SymLib, val sort: Sort, val occurrence: Occ
 
 class SortInfo private(sort: Sort, symlib: Parser.SymLib) {
   val constructors = symlib.constructorsForSort.getOrElse(sort, Seq())
+  lazy val nonEmptyConstructors: Seq[SymbolOrAlias] = constructors.filter(c => c.ctr != "inj" || SortInfo(c.params(0), symlib).category.hasIncompleteSignature(Seq(), true, SortInfo(c.params(0), symlib)))
```
virgil-serbanuta, line 265: Does `nonEmptyConstructors` work properly if you have a chain of injections? E.g., let's say that you have three sorts, `A`, `B` and `C`. Let's say that `C` has three non-inj constructors `C1`, `C2` and `C3`, `B` has only an injection from `C`, and `A` has only an injection from `B` (I hope I got the direction of the injections correctly). Then, If I read the code properly, `C.nonEmptyConstructors` will return `C1`, `C2` and `C3` (because of the `c.ctr != "inj"` condition), `B.nonEmptyConstructors` will return `inj` because of the `hasIncompleteSignature` condition, while `A.nonEmptyConstructors` will return an empty seq.

Is that how it works/should work?

dwightguth: This is not quite correct. `A.nonEmptyConstructors` is not completely empty, it contains `inj{C, A}`. Basically, we remove `inj{B, A}` from the list of constructors that you must match when matching A in order to be exhaustive, because all terms of sort B are `inj{C, B}`, so the only term that could appear under `inj{B, A}` would be `inj{C, B}`, but it cannot appear because the double injections get automatically compacted and you would have simply `inj{C, A}`. We keep that one because sort C has non injection constructors. Thus, if you match all terms of sort C, you also match all terms of sorts B and A, as expected.

PR #330, opened 2020-05-07
New function encoding

dwightguth:
This implements support for the function encoding in kframework/k#1252. It's ready for review, but please don't merge yet as there is still some possibility the encoding will change.


lib/ast/AST.cpp
```diff
@@ -1032,6 +1036,14 @@ KOREPattern *KOREAxiomDeclaration::getRequires() const {
               }
             }
           }
+        } else if (equals->getConstructor()->getName() == "\\and" && equals->getArguments().size() == 2) {
```
virgil-serbanuta, line 1039: Why is the variable called "equals" if it can be an "and"?

dwightguth: historical reasons. I will rename it.

dwightguth:
This is now ready to merge whenever the Haskell backend changes have been made. We are not going to merge yet because there might be further changes to the rule encoding.

PR #331, opened 2020-05-08
rearrange files

dwightguth:
Part of a larger restructuring of K files. Don't review yet.


unittests/runtime-io/io.cpp
```diff
@@ -69,7 +69,7 @@ extern "C" {
 
 int overwriteTestFile() {
   int fd = ::open("test.txt", O_RDWR | O_CREAT | O_TRUNC, S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP | S_IROTH);
-  ::write(fd, "hello world!", 12);
+  BOOST_CHECK_EQUAL(::write(fd, "hello world!", 12), 12);
```
virgil-serbanuta, line 72: Just curious: this is not related to rearranging files, right?

dwightguth: Fixes some warnings that were treated as errors on newer compiler versions.

PR #334, opened 2020-05-13
AddBrackets pass in C++ unparser

dwightguth:
This code is based on `kernel/src/main/java/org/kframework/unparser/AddBrackets.java` in kframework/k. It adds feature parity with respect to brackets to the `kprint` tool which powers unparsing in llvm-krun and gdb.


test/unparse/imp/imp.cfg.kore.out
```diff
@@ -0,0 +1,19 @@
+initGeneratedTopCell[0m ([0m $IO |->[0m on
```
virgil-serbanuta, line 1: What is the `[0m` stuff? terminal escape sequences?

tools/kprint/addBrackets.cpp
```diff
@@ -0,0 +1,202 @@
...
+  if (auto outer = dynamic_cast<KORECompositePattern *>(t.get())) {
+    if (outer->getConstructor()->getName() == "\\dv") {
+      return t;
+    }
+    std::vector<sptr<KOREPattern>> newItems;
+
+    sptr<KORECompositePattern> result = KORECompositePattern::Create(outer->getConstructor());
+    int position = 0;
+    for (auto &inner : outer->getArguments()) {
+      KORECompositePattern *leftCapture = getLeftCapture(previousLeftCapture, outer, position, data);
```
virgil-serbanuta, line 187: I was trying to figure these things out, but it's fairly slow, so I thought it's better to ask.

What do `BARE_LEFT` and `BARE_RIGHT` mean? What does `capture` in `leftCapture` and `rightCapture` mean?

Is it worth adding code comments to explain some of these, or it's just me not having the context/knowledge that other people that work with this code have?

Is the `terminals` attribute documented anywhere? If yes, where? I tried searching for it in the `pending-documentation.md` file, but I didn't find it. If not, should it be?

test/unparse/imp/imp.cfg.kore.out
dwightguth: these are ansi terminal color codes. I include them in the output because it means we can test the colorization code in our CI testing.

tools/kprint/addBrackets.cpp
dwightguth: The terminals attribute is one of the attributes that gets automatically inserted into the kore file by ModuleToKORE in the frontend. It exists because the kore data otherwise does not track the relative position of terminals and nonterminals in the original production, which makes it impossible to correctly compute where to add brackets according to this algorithm. I will document the algorithm in more detail in this file. Mostly it's a straight translation of an algorithm that exists in the K frontend which is in turn a straight translation of an algorithm that exists in the maude codebase, where it definitely wasn't documented because maude's code has basically no documentation.

tools/kprint/main.cpp
```diff
@@ -62,40 +102,142 @@ int main (int argc, char **argv) {
   comms.insert("\\or");
   std::map<std::string, std::vector<std::string>> colors;
 
-  // load information about definition written by k frontend
-  std::ifstream iformats(argv[1] + std::string("/format-att.txt"));
-  std::ifstream ihooks(argv[1] + std::string("/hook-att.txt"));
-  std::ifstream iassocs(argv[1] + std::string("/assoc-att.txt"));
-  std::ifstream icomms(argv[1] + std::string("/comm-att.txt"));
-  std::ifstream icolors(argv[1] + std::string("/color-att.txt"));
+  std::map<std::string, std::set<std::string>> priorities, leftAssoc, rightAssoc;
```
ehildenb, line 105: What are these big tables here? Any way we can load them from the actual information in `domains.k` and `kast.k` instead?

dwightguth: Maybe? it's kinda tricky though. These tables contain the metadata for the stuff that is defined in kast.k and thus doesn't get translated by ModuleToKORE, like kseq, inj, dotk, \or, \and, etc. I'm not sure how I can get this information from the frontend to the backend when there aren't symbols in the definition that I can automatically insert attributes for. I'd like to! I'm just not sure how to go about it. Do you have an idea?

tools/kprint/addBrackets.cpp
```diff
@@ -0,0 +1,359 @@
...
+ * or patterns that appear between two terminals in their parent. Otherwise,
+ * parentheses are needed if priority or associativity forbids the term inside
+ * its parent, or if the term begins or ends with a nonterminal and priority
+ * and associativity do not forbid the term from appearing inside its left
+ * or right capture, respectively, and its left or right capture does not
+ * end or begin, respectively, with a terminal.
+ *
+ * Warning: this algorithm is *unsound*. There is no guarantee that after 
+ * it completes, the resulting term can be unparsed into a string that will
+ * parse unambiguously as the original AST. Indeed, the problem of inserting
```
virgil-serbanuta, line 253: Do you have an example for unsoundness? It's likely not optimal, but I couldn't find a case where (say) it leaves out a parenthesis, or it produces ambiguity.

tools/kprint/addBrackets.cpp
```diff
@@ -0,0 +1,359 @@
...
+    return previousLeftCapture;
+  } else {
+    return outer;
+  }
+}
+
+/**
+ * Compute the right capture of a pattern.
+ *
+ * If you were to unparse the AST of a term without any parentheses as a
```
virgil-serbanuta, line 143: I coudn't find where we check that "P" below has a terminal. Other than that, would the following text be more clear?

 * If you were to unparse the AST of a term without any parentheses as a
 * sequence of terminals, start from the point where a particular pattern P
 * (the one at "position" in outer) ends and proceed right until you
 * reach the first terminal that is part of a pattern Q (directly) such that
 * Q is a parent of P. Then Q would be the right capture of that pattern.
 * Not all terms have a right capture; for example, the right capture of the
 * top term in an AST does not exist.

Also, do we always have terminals between non-terminals? If not, then this description is likely wrong.

tools/kprint/main.cpp
```diff
@@ -62,40 +102,142 @@ int main (int argc, char **argv) {
...
+  leftAssoc["kseq"].insert("kseq");
+  leftAssoc["append"].insert("append");
+  leftAssoc["\\and"].insert("\\and");
+  leftAssoc["\\or"].insert("\\or");
+  leftAssoc["\\rewrites"].insert("\\rewrites");
+  rightAssoc["\\rewrites"].insert("\\rewrites");
+  priorities["\\exists"].insert("\\implies");
```
virgil-serbanuta, line 112: btw, unless we're need this to work with older compilers, you should be able to do
```
priorities["\\exists"] = {\\implies", "\\or", "\\and", "\\equals", "\\floor", ...}
```
(I'm not sure how readable a map initializer list would be, so I only suggested a set one).

tools/kprint/addBrackets.cpp
```diff
@@ -0,0 +1,359 @@
...
+    }
+    bool priority = isPriorityWrong(outer, innerComposite, position, data);
+    if (priority) {
+      return true;
+    }
+    if (data.terminals.at(innerName) == "0") {
+      return false;
+    }
+    Fixity innerFixity = getFixity(innerComposite->getConstructor(), data);
+    if ((innerFixity & BARE_RIGHT) && rightCapture != nullptr) {
```
virgil-serbanuta, line 284: This if (and the one after, but they are the same thing) was somewhat tricky to understand, so if you'd add an explanation of why it works, it would probably help. Suggestion (the details are probably inaccurate):

"
We are not in one of the cases where we know it's safe to skip adding parentheses.
At this point we know that `outer` does not have higher priority than `inner`, so, if the
inversePriority bool is true, then we must have added parentheses somewhere else
(e.g. around `outer`), so we don't need to add some extra ones here. Otherwise we
add some parentheses around `inner` because  we don't want `inner`'s right
operand to be parsed as a direct child of `rightCapture`.
"

tools/kprint/addBrackets.cpp
```diff
@@ -0,0 +1,359 @@
...
+ * parentheses are needed if priority or associativity forbids the term inside
+ * its parent, or if the term begins or ends with a nonterminal and priority
+ * and associativity do not forbid the term from appearing inside its left
+ * or right capture, respectively, and its left or right capture does not
+ * end or begin, respectively, with a terminal.
+ *
+ * Warning: this algorithm is *unsound*. There is no guarantee that after 
+ * it completes, the resulting term can be unparsed into a string that will
+ * parse unambiguously as the original AST. Indeed, the problem of inserting
+ * brackets into an AST in order to ensure this property is NP-hard, and
```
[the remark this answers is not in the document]
dwightguth: Part of the problem is that grammars can contain the same terminals that brackets contain, such as ( and ). Pretty sure the problem of "find a string that parses unambiguously as this AST' is np-hard.

tools/kprint/addBrackets.cpp
```diff
@@ -0,0 +1,359 @@
...
+    Fixity fixity = getFixity(position, outer->getConstructor(), data);
+    if (fixity == EMPTY) {
+      return false;
+    }
+    bool priority = isPriorityWrong(outer, innerComposite, position, data);
+    if (priority) {
+      return true;
+    }
+    if (data.terminals.at(innerName) == "0") {
+      return false;
```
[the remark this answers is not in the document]
dwightguth: it's reached if you have a symbol that has no terminals and one nonterminal.

tools/kprint/addBrackets.cpp
dwightguth: I mean, I believe the obvious example is this:

```
syntax Exp ::= Exp "++" | "++" Exp | Exp Exp | "(" Exp ")" [bracket]

(1 ++) 2
```

This algorithm is not smart enough to insert a parentheses here.

tools/kprint/addBrackets.cpp
dwightguth: I mean, I have no idea how this algorithm works, I just know it does. Somewhere out there on the internet is a master's thesis done by someone in Europe whose algorithm eventually became part of the Maude implementation, from which we took it (and believe me, it was wholly impenetrable to understand in that form). But without access to that paper, I couldn't begin to explain why this particular incantation is the one we use here.

tools/kprint/addBrackets.cpp
virgil-serbanuta: Thanks! Could you add it to the comment?

tools/kprint/addBrackets.cpp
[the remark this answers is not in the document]
virgil-serbanuta: Would that mean that you have an 'inj'? And, if so, would it be filtered out by the `if` in `addBrackets`?

Even if that's the case, I'm not saying that you should remove this if+return, I'm just trying to understand the code better.

tools/kprint/addBrackets.cpp
virgil-serbanuta: Ah, ok. Well, I tried to understand it, so that's why I asked most of these questions...

tools/kprint/addBrackets.cpp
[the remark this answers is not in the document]
dwightguth: No, not all productions with zero terminals and one nonterminal are subsort productions in K.

PR #335, opened 2020-05-14
llvm-krun --dry-run

dwightguth:


runtime/alloc/arena.cpp
```diff
@@ -58,7 +58,10 @@ static unsigned blocks_left = 0;
 static void* megabyte_malloc() {
   if (blocks_left == 0) {
     blocks_left = 15;
-    posix_memalign(&superblock_ptr, BLOCK_SIZE, BLOCK_SIZE * 15);
+    if(int result = posix_memalign(&superblock_ptr, BLOCK_SIZE, BLOCK_SIZE * 15)) {
```
virgil-serbanuta, line 61: perhaps a space after `if`?

PR #338, opened 2020-05-26
add alignment param to alloc

dwightguth:
This is needed for the C semantics so that the native memory allocated for each object is aligned according to that object's alignment requirement.


runtime/meta/ffi.cpp
```diff
@@ -304,14 +304,23 @@ extern "C" {
...
     if (allocatedKItemPtrs.find(kitem) != allocatedKItemPtrs.end()) {
       return allocatedKItemPtrs[kitem];
     }
 
     size_t s = mpz_get_ui(size);
+    size_t a = mpz_get_ui(align);
 
-    string * ret = (string *) calloc(sizeof(string *) + s, 1);
+    string * ret;
+    int result = posix_memalign((void **)&ret, a < sizeof(void *) ? sizeof(void *) : a, sizeof(string *) + s);
```
virgil-serbanuta, line 319: Why not `aligned_alloc`? That's likely to also work on non-posix systems.

unittests/runtime-ffi/ffi.cpp
```diff
@@ -453,40 +454,44 @@ BOOST_AUTO_TEST_CASE(call_variadic) {
 }
 
 BOOST_AUTO_TEST_CASE(alloc) {
-  mpz_t s1;
+  mpz_t s1, s2;
```
virgil-serbanuta, line 457: Would it be hard to also add a test for alignment?

runtime/meta/ffi.cpp
```diff
@@ -304,14 +304,23 @@ extern "C" {
     if (!mpz_fits_ulong_p(size)) {
       throw std::invalid_argument("Size is too large");
     }
+    if (!mpz_fits_ulong_p(align)) {
+      throw std::invalid_argument("Alignment is too large");
+    }
 
     if (allocatedKItemPtrs.find(kitem) != allocatedKItemPtrs.end()) {
       return allocatedKItemPtrs[kitem];
```
virgil-serbanuta, line 312: What would happen when calling this function with the same kitem, but different alignment requirements?:

runtime/meta/ffi.cpp
dwightguth: Well, the llvm backend already depends on posix_memalign, but aligned_alloc requires that size is a multiple of alignment, which we don't necessarily want to assume.

unittests/runtime-ffi/ffi.cpp
dwightguth: I mean, I'm not sure that we could really test something reliably. We could test that the address returned is indeed aligned by a certain amount, but the actual address returned is different every time we run the unit tests, so it seems sketchy to me to make the tests depend on the address like that. Besides, the implementation is allowed to return a value that is aligned by a greater amount than specified by the user, which means that even if the test passed, that doesn't really prove there's not a bug.

runtime/meta/ffi.cpp
dwightguth: You would get a value that is not necessarily aligned by the amount specified; In practice we assume the user won't call the code that way, but if you want, I can add a check that the memory returned is indeed aligned by the amount the user requested the second time around (and have it be an error otherwise).

virgil-serbanuta: I would prefer having an error.

unittests/runtime-ffi/ffi.cpp
virgil-serbanuta: Hm, ok, maybe it's fine to not have a test, but I think that if you allocate 1 byte with a 32k alignment repeatedly without freeing the previous memory (say, 10 times, or 100 times if you really want to make sure), then I'd say that's (really) unlikely to pass unless the hook actually aligns things.

PR #344, opened 2020-06-11
try to revert problematic part of commit

dwightguth:


dwightguth:
It caused compile options to be dropped by the makefile.

PR #348, opened 2020-06-18
hook_FLOAT_rat2float

dwightguth:


runtime/arithmetic/float.cpp
```diff
@@ -443,6 +443,27 @@ bool hook_FLOAT_sign(SortFloat a) {
...
+  if (!mpz_fits_ulong_p(prec)) {
+    throw std::invalid_argument("Precision out of range");
+  }
+  unsigned long uprec = mpz_get_ui(prec);
+  if (!mpz_fits_ulong_p(exp)) {
+    throw std::invalid_argument("Exponent out of range");
+  }
+  unsigned long uexp = mpz_get_ui(exp);
+  floating result[1];
+  mpfr_enter(uprec, uexp, result);
```
virgil-serbanuta, line 456: What does this do? I didn't find this function here:
https://www.mpfr.org/mpfr-current/mpfr.html#Function-and-Type-Index

From its name, it seems that it would be better to have it before line 462, but I can't tell for sure.

dwightguth: it's a function defined higher up in that file. mpfr_enter and mpfr_leave take care of dealing with all the logic associated with emulating IEEE 754 floating point, as well as with the llvm backend representation of floating point terms.

virgil-serbanuta: Ah, ok, thanks. Would it make sense to move this after `mpq_canonicalize`?

I'm fine with merging the PR with or without this change.

PR #350, opened 2020-06-30
Make the llvm backend compile on MacOS

virgil-serbanuta:
`-Wno-reorder-init-list` is needed because there are 4 lines like this:
https://github.com/kframework/llvm-backend/blob/master/runtime/io/io.cpp#L432
Apparently the order of fields for the `flock` struct is not set in stone.

`-Wno-deprecated-register` is needed because of things like this:
```
/Users/virgil/runtime-verification/llvm-backend/build/lib/parser/KOREScanner.cpp:2735:2: error: 
      'register' storage class specifier is deprecated and incompatible with
      C++17 [-Werror,-Wdeprecated-register]
        register yy_state_type yy_current_state;
        ^~~~~~~~~
```


dwightguth:
It seems your environment for building the backend is misconfigured. The llvm backend is designed to be built with the homebrew package of flex, which is sufficiently new that this problem with the register storage class should not occur. You are probably still using the version of flex that comes with Mac OS, which is too old. 

The issue with struct flock is legitimate, but the fix is incorrect; we should not simply disable the warning that the program is ill formed, we should actually fix the program. The simple solution is to simply zero-initialize the struct and then manually set the fields to their appropriate values after initialization.

virgil-serbanuta:
Fixed.

PR #354, opened 2020-07-24
STRING.transcode

dwightguth:


ehildenb:
@dwightguth this looks good to merge to me.

runtime/strings/strings.cpp
```diff
@@ -320,6 +331,21 @@ extern "C" {
     return move_int(result);
   }
 
+  SortString hook_STRING_transcode(SortString input, SortString inputCharset, SortString outputCharset) {
```
virgil-serbanuta, line 334: I think we should have some documentation about what the various hooks are supposed to do.

PR #359, opened 2020-08-20
fix bug involving as patterns and list ranges

dwightguth:


virgil-serbanuta (approved):
I'm assuming that, for `pat #as P`, `listRange` was returning the empty `Seq` until now, and this PR changes it to return the `pat`'s value instead.

However, I'm not fully sure what `listRange` actually does, why does get a Constructor as an argument, and what does the `Occurrence` argument mean.

I'll approve since it's likely to be reasonable.

dwightguth:
basically, if I have a pattern like `(ListItem(_) L ListItem(_)) #as V`, previously, L was not actually getting bound to anything, because the code that would do that is handled by the `listRange` function, and it thought it didn't have any such variables due to how it was previously implemented.

PR #361, opened 2020-09-03
Add \left-assoc and \right-assoc to llvm backend

dwightguth:


runtime/configurationparser/ConfigurationParser.cpp
```diff
@@ -16,65 +17,89 @@ extern "C" {
...
-    return (block *) ((uint64_t)tag << 32 | 1);
-  }
+      std::ostringstream Out;
+      symbol->print(Out);
+      uint32_t tag = getTagForSymbolName(Out.str().c_str());
+      if (isSymbolAFunction(tag) && constructor->getArguments().empty()) {
+        output.push_back(evaluateFunctionSymbol(tag, nullptr));
+        continue;
+      } else if (constructor->getArguments().empty()) {
+        output.push_back((block *) ((uint64_t)tag << 32 | 1));
```
virgil-serbanuta, line 53: I'd add a function/macro whose name explains what kind of encoding is done here.

runtime/configurationparser/ConfigurationParser.cpp
```diff
@@ -16,65 +17,89 @@ extern "C" {
...
-      if (!((uint64_t)child & 1) && layout(child) != 0) {
-        uint32_t tag = tag_hdr(child->h.hdr);
-	if (tag >= first_inj_tag && tag <= last_inj_tag) {
-          return child;
+      if (tag >= first_inj_tag && tag <= last_inj_tag) {
+        uint16_t layout_code = layout_hdr(headerVal.hdr);
+        layout *data = getLayoutData(layout_code);
+        if (data->args[0].cat == SYMBOL_LAYOUT) {
+          block *child = (block *)arguments[0];
+          if (!((uint64_t)child & 1) && layout(child) != 0) {
```
virgil-serbanuta, line 82: What does it mean that the lowest bit of `child` is 1? Is it the same as above? Why not a macro with an explanatory name?

runtime/configurationparser/ConfigurationParser.cpp
```diff
@@ -16,65 +17,89 @@ extern "C" {
...
+            uint32_t tag = tag_hdr(child->h.hdr);
+	    if (tag >= first_inj_tag && tag <= last_inj_tag) {
```
virgil-serbanuta, line 84: This line uses tabs for indentation, which looks wrong in the browser diff. I think the same is true for the next 5 lines.

virgil-serbanuta (approved):

virgil-serbanuta (approved):

PR #363, opened 2020-09-24
Fix llvm backend ffi hook

dwightguth:


runtime/meta/ffi.cpp
```diff
@@ -258,10 +266,22 @@ extern "C" {
...
   }
 
+
   SortInt hook_FFI_address(SortString fn) {
     char * func = getTerminatedString(fn);
-    void * handle = so_lib_handle();
-    void * address = dlsym(handle, func);
+
+    std::string funcStr = func;
+    std::map<std::string, void *> privateSymbols;
```
virgil-serbanuta, line 274: Not sure it's worth changing, but this could be a const static variable. Of course, the `const` would mean that the `if` below should use iterators.

virgil-serbanuta (approved):

runtime/meta/ffi.cpp
dwightguth: I agree it could be, but I couldn't figure out an easy way to initialize a std::map with a specific set of elements, and thus went with this approach. In the unlikely event that this method reports a lot of self time, we can optimize.

virgil-serbanuta: Wouldn't this work?
```
std::map<std::string, void *> privateSymbols = {
    {"atexit", (void *)atexit},
    {"at_quick_exit", (void *)at_quick_exit}
};
```

PR #364, opened 2020-10-02
ffi: do not consider at_quick_exit a private symbol on OSX

sskeirik:
Reason: at_quick_exit is not defined on OSX (see references in kframework/k#1580)

Note: I'm not sure if this PR is enough --- it seems like you want to handle FFI handles to these symbols specially --- but that seems to be irrelevant if the symbol doesn't exist on this platform. Hence this PR.


dwightguth (approved):

dwightguth:
Side note: it's rather pathetic that Apple still doesn't support a feature of C11 in 2020 when they've had over 9 years to implement it, but what can ya do.

PR #365, opened 2020-10-08
KREFLECTION.argv

h0nzZik:
@dwightguth What would be the best way to automatically test this? I tested it only manually, using hardcoded values for the two variables.

### DIFF
--- a/include/runtime/header.h
+++ b/include/runtime/header.h
@@ -230,7 +230,7 @@ extern "C" {
 
   stringbuffer *hook_BUFFER_empty(void);
   stringbuffer *hook_BUFFER_concat(stringbuffer *buf, string *s);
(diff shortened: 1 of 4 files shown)


dwightguth (changes requested):
It seems to me that you need to insert an injection above the strings in argv. I would just test it with unit tests if it were me.

ehildenb:
Just curious, why is this in KREFLECTION and not in something like K-IO, or even a separate module K-ARG?

KREFLECTION seems to be about getting at the internal state of the backend at runtime. argv/argc seems more about input.

dwightguth (approved):
I've approved, but you may want to wait to merge until you've tested it with the c semantics, just in case.

h0nzZik:
@dwightguth I tested it, please merge (I do not have access rights).

dwightguth (approved):

PR #366, opened 2020-10-12
Building with Gcc on Fedora

h0nzZik:
Most of the changes are straightforward. One tricky thing is the change of `hasTerminalAtIdx`, where we might have passed a negative value to it, but changed the parameter to `size_t`, which is unsigned. In that case, the old code:
```
  if (!hasTerminalAtIdx(terminals, terminalPos-1)) {
    result |= BARE_LEFT;
  }
```
would enter the if-block, which is what we must do in the new code:
```
  if (terminalPos == 0 || !hasTerminalAtIdx(terminals, terminalPos-1)) {
    result |= BARE_LEFT;
  }
```

### DIFF
--- a/runtime/alloc/alloc.cpp
+++ b/runtime/alloc/alloc.cpp
@@ -51,25 +51,31 @@ void setKoreMemoryFunctionsForGMP() {
    mp_set_memory_functions(koreAllocMP, koreReallocMP, koreFree);
 }
 
-__attribute__ ((always_inline)) void* koreAlloc(size_t requested) {
+#ifdef __clang__
+#define ALWAYS_INLINE __attribute__ ((always_inline))
+#else
+#define ALWAYS_INLINE
+#endif
+
+ALWAYS_INLINE void* koreAlloc(size_t requested) {
   return arenaAlloc(&youngspace, requested);
 }
 
-__attribute__ ((always_inline)) void* koreAllocToken(size_t requested) {
+ALWAYS_INLINE void* koreAllocToken(size_t requested) {
   size_t size = (requested + 7) & ~7;
   return arenaAlloc(&youngspace, size < 16 ? 16 : size);
 }
 
-__attribute__ ((always_inline)) void* koreAllocOld(size_t requested) {
+ALWAYS_INLINE void* koreAllocOld(size_t requested) {
   return arenaAlloc(&oldspace, requested);
 }
 
-__attribute__ ((always_inline)) void* koreAllocTokenOld(size_t requested) {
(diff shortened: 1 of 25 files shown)


runtime/alloc/alloc.cpp
```diff
@@ -51,25 +51,31 @@ void setKoreMemoryFunctionsForGMP() {
    mp_set_memory_functions(koreAllocMP, koreReallocMP, koreFree);
 }
 
-__attribute__ ((always_inline)) void* koreAlloc(size_t requested) {
+#ifdef __clang__
+#define ALWAYS_INLINE __attribute__ ((always_inline))
+#else
+#define ALWAYS_INLINE
```
dwightguth, line 57: this is a serious problem. If these functions are not inlined, the performance of the entire backend completely bombs to something utterly unacceptable. Gcc supports always inliniing as well, so at the very least you should handle that case.

dwightguth (changes requested):
I don't understand why you changed the data type of `position` in the kprint code to an unsigned type. Please explain, because it makes the code much harder to reason about if you're worried about underflow.

runtime/alloc/alloc.cpp
h0nzZik: Do we have any benchmark to measure it? When I use that attribute for gcc, it says:
```
error: ‘always_inline’ function might not be inlinable [-Werror=attributes]
```
Maybe disable the warning using pragma GCC ?
```
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wattributes"
```

h0nzZik:
The case with `hasTerminalAtIdx` is explained above. Then there is the code in `getRightCapture`, where I removed the possibility of underflow in the if:
```
 if (position == outer->getArguments().size() - 1 && (fixity & BARE_RIGHT))
```
and replaced it with code that neither underflows nor overflows, since `position < outer->getArguments().size()`:
```
if (position + 1 == outer->getArguments().size() && (fixity & BARE_RIGHT)) {
```
. I changed the argument of `getRightCapture` to `size_t` to prevent this error/warning:
```
error: comparison of integer expressions of different signedness: ‘int’ and ‘std::vector<std::shared_ptr<kllvm::KOREPattern> >::size_type’
```

dwightguth:
Just cast the size_t to an int. Then you don't have to worry about an int being interpreted as an unsigned number, but you keep the easy to understand code. Yes, it's a loss of precision, but the user is never going to have a pattern with millions of children and the backend would choke if they did anyway.

ehildenb:
No plans to support gcc for the llvm bacckend.

PR #368, opened 2020-10-14
fix bug involving garbage collection

dwightguth:
This commit fixes a bug that occurs when a young collection cycle starts
in between filling up an old generation memory block, and the allocation
of a fresh block in the oldspace.

### DIFF
--- a/include/runtime/arena.h
+++ b/include/runtime/arena.h
@@ -20,7 +20,7 @@ struct arena {
   static struct arena name = { 0, 0, 0, 0, 0, id }
 
 #define mem_block_start(ptr) \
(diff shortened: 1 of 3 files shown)


dwightguth:
I would love to have a regression test for this, but I have no idea how to reliably trigger the exact sequence of events that leads to the bug, which is:

1. An old generation memory block gets completely filled to the exact byte.
2. A young generation collection occurs in which at least one object is promoted to the old generation, before the new old generation block has been allocated yet.

In practice, the exact same semantics, seemingly deterministic, can sometimes trigger this bug and sometimes not, depending on the exact version of gmp or mpfr you have installed. So I really have no idea how to reliably test for this.

virgil-serbanuta (approved):

PR #371, opened 2020-11-03
Build with Nix

ttuegel:
### TODO:

- [x] Clean up the Clang configuration.
- [x] Build a standalone JAR for `org.kframework.backend.llvm.matching` to speed up integration testing.
- [x] Push to binary cache.


ttuegel:
@dwightguth This is ready for your review. It will push to the binary cache when the build is completed. The only change to the backend itself that I made was building a second JAR for `matching` with all dependencies included. This makes it easier to run the integration tests. If that's not alright with you, we can talk about finding a different way.

ehildenb (approved):

dwightguth:
This looks okay, but has the change to the pom been tested with the k frontend te make sure it doesn't break non nix builds?

ttuegel:
Yes, I tested the frontend build with these changes, and it was unaffected.

PR #381, opened 2020-12-09
Build on Fedora (with Clang)

h0nzZik:


ehildenb:
Reopen if needed, but we're pretty sure this change is in via another PR.

PR #382, opened 2020-12-09
Allocation-heuristic based collection interval

dwightguth:
What this does:

* removes the configurable "collect every N rewrite steps" parameter from the build system and the runtime API
* Tracks the total number of blocks allocated in the youngspace
* Collects after a rewrite step if the youngspace is almost full (ie, there is less than 1MB remaining and also the youngspace is over 95% the size it was after the last collection)
* adds a parameter that configures the initial size of the youngspace

The next steps are going to be to collect during an allocation if we don't have enough free space to allocate something, but this will require stack maps and so it has been deferred for a future PR.

I measured the code on the OpenZeppelin tests and they pass. They take about the same amount of time and use about the same amount of memory for most of the run, but the peak memory usage caps out at 11.7 GB instead of the >60GB required before this change.

I have also tested the code on the KEVM blockchain tests and they pass as well, so I am reasonably certain this doesn't introduce any regressions into the garbage collection process.

### DIFF
--- a/runtime/alloc/arena.cpp
+++ b/runtime/alloc/arena.cpp
@@ -214,6 +214,10 @@ ssize_t ptrDiff(char *ptr1, char *ptr2) {
   }
 }
 
+size_t arenaSize(const struct arena *Arena) {
+  return Arena->num_blocks * (BLOCK_SIZE - sizeof(memory_block_header));
+}
+
 void freeAllMemory() {
   memory_block_header *superblock = (memory_block_header *)first_superblock_ptr;
   while (superblock) {
--- a/runtime/alloc/arena.cpp
+++ b/runtime/alloc/arena.cpp
@@ -30,6 +30,8 @@ void arenaReset(struct arena *Arena) {
   Arena->block_start = 0;
   Arena->block_end = 0;
   Arena->first_collection_block = 0;
+  Arena->num_blocks = 0;
+  Arena->num_collection_blocks = 0;
   Arena->allocation_semispace_id = id;
 }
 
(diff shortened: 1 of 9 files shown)


runtime/alloc/arena.cpp
```diff
@@ -8,16 +8,10 @@
...
-  char* next_block;
-  char* next_superblock;
-  char semispace;
-} memory_block_header;
-
 const size_t BLOCK_SIZE = 1024 * 1024;
 
 #define mem_block_header(ptr) \
-  ((memory_block_header *)(((uintptr_t)(ptr)) & ~(BLOCK_SIZE-1)))
+  ((memory_block_header *)(((uintptr_t)(ptr) - 1) & ~(BLOCK_SIZE-1)))
```
theo25, line 14: I am not sure I get the fix here. Can you explain?

theo25 (approved):
Approved. I had one question, mostly for my understanding rather than to suggest there is an issue.

runtime/alloc/arena.cpp
dwightguth: A pointer pointing to an address that is megabyte aligned might point one past the last object in the block, but it will never point at the start of an object because that space within the block is taken up by the memory block header. However, if you are attempting to compute the block header of a pointer one past the end of the block, the old code would instead give you the block header of the next block sequentially in memory. This causes problems when trying to compare whether two scan pointers are members of the same block or not. Because we just chop off the low order bits, and because the first allocated byte in a block is always 24 bytes after the start of the block, we can make sure that this macro always takes you to the address associated with the block header of the block the pointer points to or is one past the end of by simply subtracting one prior to chopping off the low order bits.

PR #385, opened 2020-12-18
set NOT_YOUNG_OBJECT_BIT on ffi alloc'd Bytes

dwightguth:
This fixes a regression found in the C semantics when building a version of K that includes the llvm backend commit e696645d5a50a7ad06d4fed392504bd50fe34ff4. What happens is that the Bytes object allocated by the FFI module isn't getting the bit set telling GC not to migrate the object, so it migrates the object, leading to corrupt memory when you try to read the bits via its original native address.


theo25 (approved):
Looks fine to me.

PR #389, opened 2021-01-27
Improvements to matching heuristics

dwightguth:
Contains 3 changes:
* a refactoring of the b heuristic
* a bugfix for the r heuristic
* a new heuristic, o, that seems to show some promise when applied to semantics that behave pathologically on the `qba` series of heuristics.


theo25 (approved):
Looks good to me.

PR #394, opened 2021-02-02
Fix substitution filtering

dwightguth:


ttuegel:
This fixed my problem with the Haskell backend tests.

lib/ast/AST.cpp
```diff
@@ -945,8 +945,11 @@ sptr<KOREPattern> KORECompositePattern::filterSubstitution(PrettyPrintData const
           return result->getArguments()[0];
         }
       }
+      return result;
+    } else if (constructor->getName() == "\\or") {
+      return result;
     }
-    return result;
+    return shared_from_this();
```
theo25, line 956: So, if I am reading this correctly, there is a case when you do recurse but then ignore the result returned from the recursion (when after the for loop at line 934, you don't enter the `if` or the `else if` blocks). This seems off but if it is intended, let me know and I will approve.

theo25 (changes requested):
I left a comment for you.

lib/ast/AST.cpp
dwightguth: I will rewrite this so that we skip the recursion in this case.

theo25 (approved):
Looks good to me.

PR #397, opened 2021-02-25
Fix stack overflow when destructing the parsed KORE

h0nzZik:
When executing one cFS OSAL test case with the new c-semantics, the default destructor of `KOREPattern` was causing a stack overflow right after the translated program in KORE was parsed. I propose an alternative destruction scheme for `KOREPattern` - one which does not do recursive calls and thus does not overflow.

This was tested only with the c-semantics. @dwightguth do you have any idea how and where to write a test case for parsing which would be a part of this repository?


dwightguth:
Looks like tests are failing.

h0nzZik:
@dwightguth That is the Arch in Docker issue.

dwightguth (approved):

PR #398, opened 2021-03-03
Faster parsing - cache calls to `getTagForSymbolName`

h0nzZik:
Using std::map as the cache was sufficient. I wanted to try some prefix/radix tree, but with this change, all the parsing time is spent elsewhere, so I didn't try it.

### DIFF
--- a/lib/codegen/EmitConfigParser.cpp
+++ b/lib/codegen/EmitConfigParser.cpp
@@ -47,7 +47,7 @@ llvm::LLVMContext &Ctx = module->getContext();
 static void emitGetTagForSymbolName(KOREDefinition *definition, llvm::Module *module) {
   llvm::LLVMContext &Ctx = module->getContext();
   auto type = llvm::FunctionType::get(llvm::Type::getInt32Ty(Ctx), {llvm::Type::getInt8PtrTy(Ctx)}, false);
-  auto func = getOrInsertFunction(module, "getTagForSymbolName",
+  auto func = getOrInsertFunction(module, "getTagForSymbolNameInternal",
       type);
   auto CurrentBlock = llvm::BasicBlock::Create(Ctx, "");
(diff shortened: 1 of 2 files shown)


dwightguth (commented):
If this works and solves the performance issue, I guess I'm not opposed per se. But this seems like a much more inefficient solution than what I thought you were going to do, which was to just at compilation time, build a std::unordered_map that is used as the implementation of getTagForSymbolName rather than the if/else chain that uses strcmp. Then getTagForSymbolName would be amortized O(1) in the number of symbols in the definition rather than O(n). Then you wouldn't need to cache the results of the function because the function itself would be as fast as what you're currently using for the cache.

Is there a particular reason you didn't go for that approach?

h0nzZik:
I didn't try any compile-time approach because it seemed to me to be more complex. I wanted to try the simplest solution that might reduce the problem - and this caching made it disappear.

dwightguth:
Alright. I would only suggest one change to this pr, then. The `getTagForSymbolName` symbol should refer to the cached version, not the uncached version.

h0nzZik:
@dwightguth Would renaming the generated `getTagForSymbolName` into `_getTagForSymbolName` and having a global cache work? I assume there is no threading involved.

dwightguth:
I mean, not with that name, no. C/C++ has undefined behavior if a global variable or function starts with an underscore followed by a lowercase letter. But in general, renaming the function ought to work. I would make the cache thread-local though just to be on the safe side. There is no threading within the interpreter itself, but if you have a global cache and someone tries to embed an interpreter into a threaded application, then you're going to end up racing on accesses to the cache. In general we want to support cases like that, so making the cache thread-local is a good idea. But it should just be a matter of adding the `thread_local` specifier to the variable.

dwightguth:
I will take this over since @h0nzZik is no longer working on RV-Match.

theo25 (approved):
Approved.

PR #399, opened 2021-03-03
fix bug in LLVMKompile2.cmake

dwightguth:
This fixes an issue reported by @malturki where if you are using LLVMKompile2.cmake to build your interpreter, and you modify a .k file and nothing else, it doesn't rerun the linking step, leading to an out of date interpreter even though make thinks everything is up to date.

### DIFF
--- a/cmake/LLVMKompile2.cmake
+++ b/cmake/LLVMKompile2.cmake
@@ -20,6 +20,7 @@ add_custom_command(
 add_custom_target(definition
 	DEPENDS "${KOMPILED_DIR}/definition.o")
 add_dependencies(${TARGET_NAME} definition)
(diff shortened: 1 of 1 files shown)


theo25 (approved):
I approved. I got a little confused as to why this was needed but I think I figured it out. It's so that the build system redoes the linking when definition.o is updated, which was an immediate dependency before, right?

dwightguth:
Yes. Basically, because we are manually adding definition.o to the link command line rather than explictly adding it as a translation unit within the target, it didn't know that it needed to relink the binary when we rebuilt the object file. There's probably a better solution, but this one seems to have the incrementality we want, so I'm not hugely worried about it.

PR #403, opened 2021-03-19
Nix: Find clang at runtime automatically

ttuegel:
This pull request patches `llvm-kompile-clang` during the Nix build so that it finds the right compiler at runtime without help from the user.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -55,6 +55,7 @@ let
   llvm-backend = callPackage ./nix/llvm-backend.nix {
     inherit llvmPackages src;
     inherit (ttuegel) cleanSourceWith;
+    host.clang = clang;
   };
 
   mavenix = import sources."mavenix" { inherit pkgs; };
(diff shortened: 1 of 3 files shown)


default.nix
```diff
@@ -55,6 +55,7 @@ let
   llvm-backend = callPackage ./nix/llvm-backend.nix {
     inherit llvmPackages src;
     inherit (ttuegel) cleanSourceWith;
+    host.clang = clang;
```
andreiburdusa, line 58: What is `host`?

andreiburdusa (commented):
Should line 41 from `default.nix` also be updated not to mention 1clang`?

default.nix
ttuegel: The usual terminology in compilers is:

- `build` is the machine that is building the compiler
- `host` is the machine where the compiler will run
- `target` is the machine where the programs _built_ by the compiler will run (you usually only see this with cross-compiling)

This is the `host` Clang because it will run when the LLVM backend runs, but it won't be used to build the backend itself.

andreiburdusa (approved):

PR #406, opened 2021-04-30
add ifdef guard for sret change

dwightguth:
This fixes a regression where the llvm backend would not build on older versions of LLVM. Since Debian Buster is still the latest Debian Stable and buster-backports only has llvm 8 as its newest version of llvm, we still want to support llvm 8 for the time being. Hence, this change.


theo25 (approved):
Looks fine to me.

PR #407, opened 2021-04-30
Add flag to build Nix package with LTO

ttuegel:
This allows us to build the Nix package with LTO, and populates the cache.


andreiburdusa (approved):

ttuegel:
@andreiburdusa Could you take care of adding a similar flag to the frontend at kframework/k, and passing the flag through to the LLVM backend?

andreiburdusa:
> @andreiburdusa Could you take care of adding a similar flag to the frontend at kframework/k, and passing the flag through to the LLVM backend?

Sure, I'll try

PR #408, opened 2021-05-14
Search in llvm backend

dwightguth:
Doesn't yet support search patterns or --search-all, etc. But the rudiments are there, and I have tested against lambda++ and C.

It also doesn't support the iterated pattern matching optimization, but it will still fall back gracefully on the unoptimized version if you enable it.


theo25 (commented):
I think this looks fine.

theo25 (approved):

PR #409, opened 2021-05-19
dedupe state list during search

dwightguth:
This is an optimization of the search algorithm from the previous PR: we only insert a state into the state list once if it otherwise appears during search multiple times. We are still not guaranteed to process each state exactly once, but it does cut down significantly on the number of states processed in some use cases, such as when modeling threads in K.

### DIFF
--- a/runtime/util/search.cpp
+++ b/runtime/util/search.cpp
@@ -47,11 +47,14 @@ std::unordered_set<block *, HashBlock, KEq> take_search_steps(int64_t depth, blo
   }
 
   std::unordered_set<block *, HashBlock, KEq> results;
+  std::unordered_set<block *, HashBlock, KEq> states_set;
   states.clear();
+  states_set.insert(subject);
   states.push_back(subject);
   while(!states.empty() && depth != 0) {
     state = states.front();
     states.pop_front();
+    states_set.erase(state);
     if (depth > 0) depth--;
     uint64_t count;
     block **stepResults = take_search_step(state, &count);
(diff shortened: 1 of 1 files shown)


runtime/util/search.cpp
```diff
@@ -47,19 +47,25 @@ std::unordered_set<block *, HashBlock, KEq> take_search_steps(int64_t depth, blo
...
 
   std::unordered_set<block *, HashBlock, KEq> results;
+  std::unordered_set<block *, HashBlock, KEq> states_set;
   states.clear();
+  states_set.insert(subject);
   states.push_back(subject);
   while(!states.empty() && depth != 0) {
     state = states.front();
     states.pop_front();
+    states_set.erase(state);
```
theo25, line 57: Maybe I am wrong here, but why couldn't  we just not erase states from this set. If I understand correctly, the way it's implemented right now, we avoid exploring from the same state if it is already in the queue for future exploration. If we never erase states from the `states_set` data structure, we will also be avoiding re-exploring states that have been inserted and in the queue in the past, thus have already been explored.

theo25 (commented):
I think this is fine. I left a comment for a potential further optimization. If I am wrong, let me know and I will approve.

runtime/util/search.cpp
dwightguth: That is a good point.

theo25 (approved):

runtime/util/search.cpp
dwightguth: We chose not to do this because it causes a memory leak.

PR #411, opened 2021-06-09
don't print spurious error if we pass substitution filtering to kprint

dwightguth:


tools/kprint/main.cpp
```diff
@@ -9,7 +9,7 @@ using namespace kllvm::parser;
 sptr<KOREPattern> addBrackets(sptr<KOREPattern>, PrettyPrintData const&);
 
 int main (int argc, char **argv) {
-  if (argc != 3 && argc != 4) {
+  if (argc != 3 && argc != 4 && argv != 5) {
```
theo25: This seems like a typo, right? Maybe `argc != 5` ?

theo25: Never mind, already fixed

theo25 (approved):

gtrepta (approved):

PR #414, opened 2021-07-09
Fix two issues relating to the color and colors attributes in kprint

dwightguth:
This contains fixes for two bugs that @SchmErik discovered when taking the K tutorial. The first happened because we were not trimming whitespace from each part of the string after splitting it on the comma character, and the second happened because we were passing an incorrect length to std::string::substr. These should both work now.


SchmErik (approved):
I've built this and tried the lesson 09 tutorial exercises. This pull request builds and fixes the issues that I've been seeing. Thanks!

gtrepta (approved):
LGTM

PR #415, opened 2021-07-21
Enable building llvm-backend on MacOS with Apple Silicon (M1) CPU

spencerhaoxiao:
Currently, llvm-backend is assumed to build on MacOS running on Intel CPU with home-brew, and in the various build scripts it assumes the default installation directory of home-brew is `/usr/local`. However, home-brew for Apple Silicon (M1) is installed at `/opt/homebrew` (see [home-brew install.sh](https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh) Line 35). 

We use cmake environment variables to detect whether the host is running on MacOS with M1 CPU, and then set the variable `BREW_HOME` to `/opt/homebrew` and  to `/usr/local` if it is MacOS with Intel CPU. Then we use the custom variable to set the various include and lib paths.


CMakeLists.txt
```diff
@@ -29,11 +29,17 @@ set(CMAKE_EXE_LINKER_FLAGS_RELEASE "${CMAKE_EXE_LINKER_FLAGS_RELEASE} -flto")
...
+set(BREW_HOME "/opt/homebrew")
+else()	
 set(BACKEND_TARGET_TRIPLE "x86_64-apple-darwin")
+set(BREW_HOME "/usr/local")
+endif()
 if(NOT USE_NIX)
-include_directories(AFTER SYSTEM /usr/local/include)
-link_directories(AFTER /usr/local/lib)
-set(ENV{PKG_CONFIG_PATH} "/usr/local/opt/libffi/lib/pkgconfig")
+include_directories(AFTER SYSTEM /opt/homebrew/include)
```
dwightguth, line 40: Looks like you forgot to use BREW_HOME here

bin/llvm-kompile-clang
```diff
@@ -58,9 +58,9 @@ else
   MAINFILES=""
 fi
 
-if [[ "$OSTYPE" == "darwin"* ]]; then
+if [[ "$OSTYPE" == "Darwin"* ]]; then
   libraries="-liconv -lncurses"
-  flags="-L/usr/local/opt/libffi/lib -L/usr/local/lib -Wl,-u,_sort_table"
+  flags="-L/opt/homebrew/opt/libffi/lib -L/opt/homebrew/lib -Wl,-u,_sort_table"
```
dwightguth, line 63: And here

dwightguth (changes requested):
Looks like tests are failing on our mac ci. I see a couple lines that might explain why that is that I've highlighted, but I don't know for sure if they're the problem.

CMakeLists.txt
spencerhaoxiao: Thanks for spotting this, it is fixed now.

bin/llvm-kompile-clang
spencerhaoxiao: fixed

bin/llvm-kompile-clang
```diff
@@ -60,7 +60,7 @@ fi
 
 if [[ "$OSTYPE" == "Darwin"* ]]; then
   libraries="-liconv -lncurses"
-  flags="-L/opt/homebrew/opt/libffi/lib -L/opt/homebrew/lib -Wl,-u,_sort_table"
+  flags="-L${BREW_HOME}/opt/libffi/lib -L${BREW_HOME}/lib -Wl,-u,_sort_table"
```
dwightguth, line 63: this needs to be @BREW_HOME@

dwightguth (changes requested):

bin/llvm-kompile-clang
spencerhaoxiao: Thanks. Done.

PR #418, opened 2021-08-04
Parser: update rules to reflect KORE syntax change for ensures

### LINKED ISSUE #417: Support for new rule format
The LLVM backend should be updated so as to accept a new format for rules: https://github.com/kframework/k/issues/1945

There is already a PR up in the frontend set up to emit the new format: https://github.com/kframework/k/pull/2061
It just has to also be supported in the LLVM backend.

SchmErik:
There was an [issue](kframework/k#1945) where the the rule
```
rule LEFT => RIGHT requires REQUIRES ensures ENSURES
```
is translated to
```
axiom \implies(REQUIRES, \and(\equals(LEFT, RIGHT), ENSURES))
```
so that if ENSURES is \bottom and REQUIRES is anything else, then the axiom is axiom \bottom which is unsound. Actually, we want to say
```
axiom \implies(REQUIRES, \equals(LEFT, \and(RIGHT, ENSURES)))
```

There is a [pull request](kframework/k#2061) that changes the K frontend to emit the correct KORE syntax. This commit implements the change in the KORE syntax for the llvm-backend parsers.

Fixes #417

Signed-off-by: Erik Kaneda <erik.kaneda@runtimeverification.com>

### DIFF
--- a/lib/ast/AST.cpp
+++ b/lib/ast/AST.cpp
@@ -1159,6 +1159,12 @@ KOREPattern *KOREAxiomDeclaration::getRightHandSide() const {
               return andPattern2->getArguments()[1].get();
             }
           }
+        } else if (andPattern->getConstructor()->getName() == "\\equals" && andPattern->getArguments().size() == 2) {
+          if (auto andPattern2 = dynamic_cast<KORECompositePattern *>(andPattern->getArguments()[1].get())) {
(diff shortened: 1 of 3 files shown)


SchmErik:
@dwightguth this is ready for review

dwightguth (approved):

PR #419, opened 2021-08-10
Static term allocation

gtrepta:


lib/codegen/CreateStaticTerm.cpp
```diff
@@ -0,0 +1,318 @@
...
+  if (!globalVar->hasInitializer()) {
+    std::vector<llvm::Constant *> blockVals;
+
+    llvm::StructType *BlockHeaderType
+        = getTypeByName(Module, BLOCKHEADER_STRUCT);
+    uint64_t headerVal = symbol->getTag();
+    uint64_t sizeInBytes = llvm::DataLayout(Module).getTypeAllocSize(BlockType);
+    assert(sizeInBytes % 8 == 0);
+    headerVal |= (sizeInBytes / 8) << 32;
+    headerVal |= (uint64_t)symbol->getLayout() << LAYOUT_OFFSET;
```
dwightguth, line 44: Can we please refactor this out into its own function rather than duplicating the logic in `getBlockHeader`?

dwightguth (changes requested):
Overall looks good, but I had a couple requests

dwightguth (approved):

PR #425, opened 2021-09-14
Add hook for kompiled directory metadata

Baltoli:
This PR implements the feature suggested in [this issue](https://github.com/kframework/k/issues/1889) - that the LLVM backend should be able to provide a hooked function that returns the current `*-kompiled` directory.

The necessary information is already passed through from the command line to the LLVM backend in debug mode, so the implementation just reads the baked-in symbol and constructs a boxed K string object to return. The hook is tested with a unit test in the backend using a mocked symbol.

On the frontend side, I'll implement a matching PR that adds an appropriate function symbol to the `K-REFLECTION` module, along with an integration test.

### DIFF
--- a/lib/codegen/CreateTerm.cpp
+++ b/lib/codegen/CreateTerm.cpp
@@ -1,20 +1,20 @@
 #include "kllvm/codegen/CreateTerm.h"
-#include "kllvm/codegen/Util.h"
 #include "kllvm/codegen/Debug.h"
+#include "kllvm/codegen/Util.h"
 
 #include <gmp.h>
 #include <iomanip>
 #include <iostream>
 
-#include "llvm/IRReader/IRReader.h"
+#include "runtime/header.h" //for macros
 #include "llvm/IR/BasicBlock.h"
 #include "llvm/IR/Constants.h"
 #include "llvm/IR/DerivedTypes.h"
 #include "llvm/IR/Instructions.h"
+#include "llvm/IRReader/IRReader.h"
 #include "llvm/Support/MemoryBuffer.h"
 #include "llvm/Support/SourceMgr.h"
 #include "llvm/Support/raw_ostream.h"
-#include "runtime/header.h" //for macros
 
 namespace kllvm {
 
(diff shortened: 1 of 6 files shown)


lib/codegen/CreateTerm.cpp
```diff
@@ -651,6 +656,21 @@ llvm::Value *CreateTerm::createHook(KORECompositePattern *hookAtt, KOREComposite
...
+        dirString->getValueType(), dirString, {zero, zero},
+        "hook_KREFLECTION_kompiledDir", CurrentBlock);
+
+    if (auto arrayTy = llvm::dyn_cast<llvm::ArrayType>(dirString->getValueType())) {
+      auto len = llvm::ConstantInt::get(llvm::Type::getInt64Ty(Ctx), arrayTy->getNumElements() - 1);
+      return createFunctionCall("makeString", {SortCategory::Symbol, 0}, {ptr, len}, false, false);
+    } else {
+      assert(false && "Kompiled directory global is not a string");
+      abort();
+    }
```
Baltoli, line 673: Is there a better / more standard way of handling errors than this? This is a global value we create ourselves earlier in compilation, so I can't see a non-exceptional path on which it wouldn't be a string.

lib/codegen/CreateTerm.cpp
```diff
@@ -651,6 +656,21 @@ llvm::Value *CreateTerm::createHook(KORECompositePattern *hookAtt, KOREComposite
     llvm::Value *first = (*this)(pattern->getArguments()[0].get()).first;
     llvm::Value *second = (*this)(pattern->getArguments()[1].get()).first;
     return llvm::BinaryOperator::Create(llvm::Instruction::URem, first, second, "hook_MINT_urem", CurrentBlock);
+  } else if (name == "KREFLECTION.kompiledDir") {
```
dwightguth, line 659: This is not typically how we implement most hooks. The ones here are implemented this way out of necessity, but in general, we prefer to use the libraries in the `runtime` directory. You are looking to define a function named `hook_KREFLECTION_kompiledDir`

dwightguth (changes requested):
The code for making sure you have the correct path baked into the binary unconditionally is fine, but we typically implement most hooks using named c++ functions that are extern "C", rather than via code generation like this.

unittests/runtime-io/io.cpp
```diff
@@ -470,6 +473,11 @@ BOOST_AUTO_TEST_CASE(time) {
   BOOST_CHECK(tt >= 1573756117);
 }
 
+BOOST_AUTO_TEST_CASE(kompiledDir) {
+  auto dir = hook_KREFLECTION_kompiledDir();
+  BOOST_CHECK_EQUAL(0, strcmp(dir->data, kompiled_directory));
```
dwightguth, line 478: Let's use memcmp here because dir->data isn't necessarily null terminated

Baltoli: The LLVM code that constructs the symbol (see [docs here](https://llvm.org/doxygen/classllvm_1_1ConstantDataArray.html#a3edef3fa47c611d3d10606591213e57b) guarantees that it will be a null-terminated C string.

If we don't want to rely on that behaviour, I will need to also go back and bake in a `kompiled_directory_length` global so that we know how many bytes to `memcmp` against.

dwightguth (approved):
I'm going to approve, but you should also add a check that the length of the string returned by the hook is equal to the length of the kompiled_directory global, because otherwise, one could technically be a substring of the other and the check would still pass.

PR #426, opened 2021-09-20
Use statepoints to compute stack roots when invoking koreCollect

dwightguth:
This PR makes use of the LLVM garbage collection API in order to compute the gc roots on the stack when calling `koreCollect`.

It operates via the following pieces:

* We extract out "opaque" functions for address space casts and inttoptr and ptrtoint casts, so that the optimizer is not aware of them. This is needed due to limitations in the implementation of garbage collection in LLVM
* We modify the address space of all the garbage collected pointers in the code generation phase to use address space 1
* We invoke the RewriteStatepointsForGC llvm pass on every compilation so that we can generate the statepoint calls themselves
* We create an llvm pass that runs after RewriteStatepointsForGC that emits a global variable that tracks the sort category of every live pointer at every statepoint, so that we know how to relocate it
* We modify koreAlloc functions so that we have typed versions that return specific types, in order to better track the sort category of gc roots
* We add code to parse the stack map when the program starts
* We use libunwind in order to walk the stack and generate a list of all the gc roots during collection
* We remove the parameters to `koreCollect` since we do not need them to track roots anymore
* We switch to using `tailcc` instead of `fastcc` so as to simplify tail call optimization

Generating roots from global variables remains the same as before.

This is rather big, so I'd like multiple sets of eyes on it. Note that I have not yet added new calls to `koreCollect` at the start of other functions in the llvm backend (like `apply_rule` and `side_condition` and `eval`). I decided to do that in a follow up PR since this is already so large.

Ready for review.


lib/codegen/CreateTerm.cpp
```diff
@@ -199,15 +205,53 @@ llvm::Value *getBlockHeader(llvm::Module *Module, KOREDefinition *definition, co
...
+llvm::Value *addrspaceCast1to0(llvm::Value *val, llvm::BasicBlock *block) {
+  return addrspaceCast(val, block, 1, 0);
+}
+
+llvm::Value *ptrToInt(llvm::Value *val, llvm::BasicBlock *block) {
+  bool hasUnnamed = false;
+  std::string name = "ptrtoint_i64." + getMangledTypeStr(val->getType(), hasUnnamed);
+  auto ptr = llvm::CallInst::Create(getOrInsertFunction(block->getModule(), name, llvm::Type::getInt64Ty(block->getModule()->getContext()), val->getType()), {val}, "", block);
+  return ptr;
 }
```
Baltoli, line 229: Why are these implemented by concatenating strings to get a function name - i.e. why don't we construct an `[AddrSpaceCastInst](https://llvm.org/doxygen/classllvm_1_1AddrSpaceCastInst.html)` (c.f. pointers to integers) directly?

Happy with this if there's a good reason for it, but getting access to LLVM intrinsics through string building seems a bit off to me.

lib/codegen/CreateTerm.cpp
```diff
@@ -199,15 +205,53 @@ llvm::Value *getBlockHeader(llvm::Module *Module, KOREDefinition *definition, co
...
+  auto cast = new llvm::BitCastInst(malloc, llvm::PointerType::get(AllocType, 1), "", block);
+  return cast;
+}
+
+llvm::Value *allocateTermNoReloc(llvm::Type *AllocType, llvm::BasicBlock *block, const char *allocFn) {
+  return allocateTermNoReloc(AllocType, llvm::ConstantExpr::getSizeOf(AllocType), block, allocFn);
+}
+
+llvm::Value *allocateTermNoReloc(llvm::Type *AllocType, llvm::Value *Len, llvm::BasicBlock *block, const char *allocFn) {
+  auto malloc = llvm::CallInst::Create(koreHeapAlloc(allocFn, block->getModule()), llvm::ConstantExpr::getPtrToInt(llvm::ConstantExpr::getGetElementPtr(AllocType, llvm::ConstantPointerNull::get(llvm::PointerType::get(AllocType, 1)), llvm::ConstantInt::get(llvm::Type::getInt32Ty(block->getContext()), 1)), llvm::Type::getInt64Ty(block->getContext())), "malloccall", block);
```
Baltoli, line 249: Only a stylistic issue, but I find this quite difficult to read as the line is so long.

lib/codegen/Decision.cpp
```diff
@@ -520,12 +523,12 @@ void abortWhenStuck(llvm::BasicBlock *CurrentBlock, llvm::Module *Module, KORESy
   symbol = d->getAllSymbols().at(Out.str());
   auto BlockType = getBlockType(Module, d, symbol);
   llvm::Value *Ptr;
-  auto BlockPtr = llvm::PointerType::getUnqual(getTypeByName(Module, BLOCK_STRUCT));
+  auto BlockPtr = llvm::PointerType::get(getTypeByName(Module, BLOCK_STRUCT), 1);
   if (symbol->getArguments().empty()) {
-    Ptr = llvm::ConstantExpr::getIntToPtr(llvm::ConstantInt::get(llvm::Type::getInt64Ty(Ctx), ((uint64_t)symbol->getTag() << 32 | 1)), getValueType({SortCategory::Symbol, 0}, Module));
+    Ptr = llvm::CallInst::Create(getOrInsertFunction(Module, "inttoptr_i64.p1s_blocks", getValueType({SortCategory::Symbol, 0}, Module), llvm::Type::getInt64Ty(Ctx)), {llvm::ConstantInt::get(llvm::Type::getInt64Ty(Ctx), (((uint64_t)symbol->getTag()) << 32) | 1)}, "", CurrentBlock);
```
Baltoli, line 528: Why doesn't this fit the int-to-pointer conversion functions defined previously?

lib/codegen/Util.cpp
```diff
@@ -1,13 +1,81 @@
...
+    case Type::FP128TyID:     Result += "f128";     break;
+    case Type::PPC_FP128TyID: Result += "ppcf128";  break;
+    case Type::X86_MMXTyID:   Result += "x86mmx";   break;
+    case Type::IntegerTyID:
+      Result += "i" + std::to_string(cast<IntegerType>(Ty)->getBitWidth());
+      break;
+    }
+  }
+  return Result;
+}
```
Baltoli, line 61: I presume this is at least part of the reason for the comment I left above regarding string-concatenated intrinsics - what are the limitations in the LLVM GC API that mean we need to do this?

lib/codegen/Util.cpp
```diff
@@ -1,13 +1,81 @@
...
+
+llvm::Function* koreHeapAlloc(ValueType Cat, std::string name, llvm::Module *module) {
+  llvm::Type *size_type = llvm::Type::getInt64Ty(module->getContext());
+  llvm::Type *Ty = getParamType(Cat, module);
+  auto allocType = llvm::FunctionType::get(Ty, llvm::ArrayRef<llvm::Type*>(size_type), false);
+  bool hasUnnamed = false;
+  auto result = getOrInsertFunction(module, name + "_" + getMangledTypeStr(Ty, hasUnnamed), allocType);
```
Baltoli, line 68: Stylistic again - do we have a set of shared code style guidelines for C++? Here we're mixing `auto` and non-`auto` declarations, for example. There isn't a clang-format style set, so perhaps it's worth having that to keep things consistent.

Baltoli (commented):
Mostly questions for the sake of my own learning as much as anything else

lib/codegen/CreateTerm.cpp
dwightguth: I'll add a comment explaining this.

lib/codegen/Decision.cpp
dwightguth: I'm not sure what you mean.

lib/codegen/Util.cpp
dwightguth: We don't really. I haven't used clang-format before, so if you have suggestions about how to set things up so that stylistic things of this nature can be checked automatically, I'm willing to listen, as long as it doesn't involve requiring us to duplicate the types of variables in both the declaration and initializer.

Baltoli: I can push a PR demonstrating the use of clang-format; it's responsible for purely aesthetic things, and integrates nicely with pretty much every editor and is very low-maintenance once it's set up.

clang-tidy is a bit more involved and probably suits an occasional pass through the codebase using it to see if anything is obviously wrong / could be modernised. Happy to demonstrate that as well.

lib/codegen/Decision.cpp
Baltoli: Ah sorry - I saw `llvm::Value *ptrToInt(llvm::Value *val, llvm::BasicBlock *block);` defined above and mistakenly assumed that you had also implemented `intToPtr` as a similar helper.

Baltoli (approved):

lib/codegen/Decision.cpp
dwightguth: It was easier to just inline this one because it is the only occurrence in the code base and always has the same types.

lib/codegen/CreateTerm.cpp
```diff
@@ -502,7 +517,7 @@ llvm::Value *CreateTerm::createHook(KORECompositePattern *hookAtt, KOREComposite
     if (nwords == 0) {
       return createToken({SortCategory::Int, 0}, "0");
     }
-    auto Ptr = allocateTerm(llvm::Type::getInt64Ty(Ctx), llvm::ConstantInt::get(llvm::Type::getInt64Ty(Ctx), nwords * 8), CurrentBlock, "koreAllocAlwaysGC");
+    auto Ptr = allocateTermNoReloc(llvm::Type::getInt64Ty(Ctx), llvm::ConstantInt::get(llvm::Type::getInt64Ty(Ctx), nwords * 8), CurrentBlock);
```
theo25, line 543: Similarly, I did not check that no `allocateTerm` refactorings were missed.

lib/codegen/Decision.cpp
```diff
@@ -601,106 +601,12 @@ std::pair<std::vector<llvm::Value *>, llvm::BasicBlock *> stepFunctionHeader(uns
...
-    auto loaded = new llvm::LoadInst(ptr->getType()->getPointerElementType(), ptr, "", collect);
-    auto phi = llvm::PHINode::Create(loaded->getType(), 2, "phi", merge);
-    phi->addIncoming(loaded, collect);
-    phi->addIncoming(roots[i++], checkCollect);
-    phis.push_back(phi);
-  }
   llvm::BranchInst::Create(merge, collect);
-  i = 0;
-  unsigned rootIdx = 0;
   std::vector<llvm::Value *> results;
```
theo25, line 613: It seems like this line can be deleted.

CMakeLists.txt
```diff
@@ -117,14 +117,21 @@ find_program(OPT opt-12)
...
 find_program(LLC llc-12)
 find_program(LLC llc-11)
 find_program(LLC llc-10)
 find_program(LLC llc)
 if(${OPT} STREQUAL "OPT-NOTFOUND")
   message(FATAL_ERROR "Could not find an opt binary. Is llvm installed on your PATH?")
 endif()
-if(${LLC} STREQUAL "OPT-NOTFOUND")
+if(${LLC} STREQUAL "LLC-NOTFOUND")
+  message(FATAL_ERROR "Could not find an llvm binary. Is llvm installed on your PATH?")
```
theo25, line 137: maybe the message can be more specific: `Could not find an llvm binary` -> `Could not find an llc binary`

CMakeLists.txt
```diff
@@ -117,14 +117,21 @@ find_program(OPT opt-12)
...
+endif()
+if(${LLVM_LINK} STREQUAL "LLVM_LINK-NOTFOUND")
   message(FATAL_ERROR "Could not find an llvm binary. Is llvm installed on your PATH?")
```
theo25, line 140: Similar here: `Could not find an llvm binary` -> `Could not find an llvm-link binary`

lib/llvm/MarkTailCallsAsGCLeaf.cpp
```diff
@@ -0,0 +1,56 @@
...
+      return dirty;
+    }
+  };
+}
+
+char MarkTailCallsAsGCLeaf::ID = 0;
+
+static RegisterPass<MarkTailCallsAsGCLeaf> X("mark-tail-calls-as-gc-leaf", "Mark tailcc tail calls with gc-leaf-function attribute",
+                             false /* Only looks at CFG */,
+                             false /* Analysis Pass */);
```
theo25, line 56: This pass also seems like it should not be an analysis pass since it modifies the LLVM IR code.

theo25 (approved):
I left a handful of comments, but otherwise this looks good to me.

lib/codegen/CreateTerm.cpp
```diff
@@ -222,15 +222,30 @@ llvm::Value *ptrToInt(llvm::Value *val, llvm::BasicBlock *block) {
...
+  }
+  auto cast = new llvm::BitCastInst(malloc, llvm::PointerType::get(AllocType, 1), "", block);
+  return cast;
+}
+
+llvm::Value *allocateTermNoReloc(llvm::Type *AllocType, llvm::BasicBlock *block, const char *allocFn) {
+  return allocateTermNoReloc(AllocType, llvm::ConstantExpr::getSizeOf(AllocType), block, allocFn);
+}
+
+llvm::Value *allocateTermNoReloc(llvm::Type *AllocType, llvm::Value *Len, llvm::BasicBlock *block, const char *allocFn) {
```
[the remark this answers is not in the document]
dwightguth: This was actually a copy-paste bug. I will fix it.

theo25 (approved):

lib/llvm/EmitGCLayoutInfo.cpp
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  bool runOnModule(Module &M) override {
+    unsigned int nextID = 0;
+    unsigned int numRelocations = 0;
+    std::map<std::pair<unsigned int, unsigned int>, ValueType> cats;
+    auto &Ctx = M.getContext();
+    for (Function &F : M.functions()) {
+      for (BasicBlock &BB : F.getBasicBlockList()) {
+        for (Instruction &I : BB.getInstList()) {
```
Baltoli, line 35: A very minor issue, but you can iterate directly over Modules / Functions / BasicBlocks rather than calling these functions.

```suggestion
    for (Function &F : M) {
      for (BasicBlock &BB : F) {
        for (Instruction &I : BB) {
```

Baltoli (approved):
One minor stylistic issue but otherwise LGTM

PR #428, opened 2021-09-20
Cleanup CMake llvm code

dwightguth:
Ups the minimum required LLVM version to 10 and does some spring-cleaning type changes to the CMake build system where llvm is concerned. Should bring us mostly in alignment with how LLVM documentation recommends you structure your out-of-source CMake build.


theo25 (approved):
Looks good to me.

PR #430, opened 2021-09-24
Apply clang-format style to project

Baltoli:
This PR introduces a global clang-format style, and applies it to the llvm-backend code.

Of the commits in this PR, only https://github.com/kframework/llvm-backend/pull/430/commits/316754483f188151b7ee46f8fda0fb9193841402 is non-trivial.

### Why automatically format?

Adopting a consistent style for our C++ code means that we're all on the same page when formatting, and automating it with clang-format means that we don't have to think about formatting or style when writing the code.

Pretty much every big C++ project will use clang-format to keep things consistent; it's a very standard tool to use, and is well-supported and maintained by the LLVM project.

### Using clang-format

To install clang-format:
```console
brew install clang-format    # macOS / Homebrew
apt install clang-format     # Ubuntu / APT
```

There are several ways to use clang-format:
* **Command line:** you can run it on a single file (in place) with `clang-format -i $FILE`; doing this manually is time consuming for an entire project.
* **Editor support:** every mainstream editor will have a plugin for clang-format that either exposes a formatting shortcut for the current file, or formats automatically on save. This is the best way to use it.
* **Git commit hooks:** it's possible to run clang-format automatically on commit, but this tends to lead to dirty commits and isn't a good way to work.
* **In CI:** usually this is a read-only check on push to make sure that new code is formatted correctly. I can add this to CI if we adopt a style.

Some editor plugins:
* [Vim](https://github.com/rhysd/vim-clang-format)
* [CLion](https://www.jetbrains.com/help/clion/clangformat-as-alternative-formatter.html)
* [Emacs](https://github.com/SavchenkoValeriy/emacs-clang-format-plus)
* [VS Code](https://code.visualstudio.com/docs/cpp/cpp-ide)
* [Sublime Text](https://packagecontrol.io/packages/Clang%20Format) 

### Code style

The tool picks up a style defined in a `.clang-format` file at the root of the project; the style is defined in terms of key-value pairs that usually modify a root style (for example, the style I've defined here is based on the WebKit style).

There are [lots of options here](https://clang.llvm.org/docs/ClangFormatStyleOptions.html) that control how the formatting is applied; I'm more than happy to bikeshed any of the rules I've introduced in this PR.

### DIFF
--- a/include/kllvm/codegen/CreateTerm.h
+++ b/include/kllvm/codegen/CreateTerm.h
@@ -19,62 +19,82 @@ private:
   llvm::LLVMContext &Ctx;
   bool isAnywhereOwise;
 
-  llvm::Value *createHook(KORECompositePattern *hookAtt, KORECompositePattern *pattern);
-  llvm::Value *createFunctionCall(std::string name, KORECompositePattern *pattern, bool sret, bool fastcc);
-  llvm::Value *notInjectionCase(KORECompositePattern *constructor, llvm::Value *val);
+  llvm::Value *
+  createHook(KORECompositePattern *hookAtt, KORECompositePattern *pattern);
+  llvm::Value *createFunctionCall(
+      std::string name, KORECompositePattern *pattern, bool sret, bool fastcc);
+  llvm::Value *
+  notInjectionCase(KORECompositePattern *constructor, llvm::Value *val);
+
 public:
   CreateTerm(
-    llvm::StringMap<llvm::Value *> &Substitution,
-    KOREDefinition *Definition,
-    llvm::BasicBlock *EntryBlock,
-    llvm::Module *Module,
-    bool isAnywhereOwise) :
-      Substitution(Substitution),
-      Definition(Definition),
-      CurrentBlock(EntryBlock),
-      Module(Module),
-      Ctx(Module->getContext()),
-      isAnywhereOwise(isAnywhereOwise) {}
+      llvm::StringMap<llvm::Value *> &Substitution, KOREDefinition *Definition,
+      llvm::BasicBlock *EntryBlock, llvm::Module *Module, bool isAnywhereOwise)
+      : Substitution(Substitution)
+      , Definition(Definition)
+      , CurrentBlock(EntryBlock)
+      , Module(Module)
+      , Ctx(Module->getContext())
+      , isAnywhereOwise(isAnywhereOwise) { }
 
-/* adds code to the specified basic block in the specified module which constructs
-   an llvm value corresponding to the specified KORE RHS pattern and substitution in the
-   specified definition, and returns the value itself, along with a boolean indicating whether the resulting term could be an injection. */
+  /* adds code to the specified basic block in the specified module which
+     constructs an llvm value corresponding to the specified KORE RHS pattern
+     and substitution in the
+     specified definition, and returns the value itself, along with a boolean
+     indicating whether the resulting term could be an injection. */
   std::pair<llvm::Value *, bool> operator()(KOREPattern *pattern);
   llvm::Value *createToken(ValueType sort, std::string contents);
-  /* creates a call instructin calling a particular llvm function, abstracting certain abi and calling convention details. 
-   * name: the nmae of the function to call in llvm
-   * returnCat: the value category of the return type of the function
-   * args: the arguments to pass to the functgion
-   * sret: if true, this is a function that returns a struct constant via the C abi, ie, 
-   * the function actually returns void and the return value is via a pointe. Note that this 
-   * can be set to true even if the function does not return a struct, in which case its value
-   * is ignored.
-   * load: if the function returns a struct via sret, then if load is true, we load the value 
-   * fastcc: true if we should use the fastcc calling convention
-   * returned from the function before returning it. */
-  llvm::Value *createFunctionCall(std::string name, ValueType returnCat, const std::vector<llvm::Value *> &args, bool sret, bool fastcc);
+  /* creates a call instructin calling a particular llvm function, abstracting
+   * certain abi and calling convention details. name: the nmae of the function
+   * to call in llvm returnCat: the value category of the return type of the
+   * function args: the arguments to pass to the functgion sret: if true, this
+   * is a function that returns a struct constant via the C abi, ie, the
+   * function actually returns void and the return value is via a pointe. Note
+   * that this can be set to true even if the function does not return a struct,
+   * in which case its value is ignored. load: if the function returns a struct
+   * via sret, then if load is true, we load the value fastcc: true if we should
+   * use the fastcc calling convention returned from the function before
+   * returning it. */
+  llvm::Value *createFunctionCall(
+      std::string name, ValueType returnCat,
+      const std::vector<llvm::Value *> &args, bool sret, bool fastcc);
 
   llvm::BasicBlock *getCurrentBlock() const { return CurrentBlock; }
 };
 
-/* Creates a new llvm::Module with the predefined declarations common to all llvm modules
-   in the llvm backend. */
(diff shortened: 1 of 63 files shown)


dwightguth:
It's a little hard to review this because it's difficult to come up with a list of every different type of formatting change that was made to make sure none of them seem problematic to me. Thus far I see a whole lot of changes to whitespace, and some reordered imports. Can you tell me whether there are any other non-whitespace formatting changes that clang-format did here, and if so, what they are?

Baltoli:
Yes, I skimmed the changes earlier so I have some idea of what the major categories are; I'll leave comments in the diff

include/kllvm/codegen/CreateTerm.h
```diff
@@ -19,71 +19,95 @@ class CreateTerm {
...
+   * certain abi and calling convention details. name: the nmae of the function
+   * to call in llvm returnCat: the value category of the return type of the
+   * function args: the arguments to pass to the functgion sret: if true, this
+   * is a function that returns a struct constant via the C abi, ie, the
+   * function actually returns void and the return value is via a pointe. Note
+   * that this can be set to true even if the function does not return a struct,
+   * in which case its value is ignored. load: if the function returns a struct
+   * via sret, then if load is true, we load the value fastcc: true if we should
+   * use the fastcc calling convention returned from the function before
+   * returning it. */
```
Baltoli, line 57: It likes to reflow comments; I can either disable this behaviour (so comments will get ignored completely), or introduce line breaks between your list items so they get identified as distinct items.

include/kllvm/codegen/CreateTerm.h
```diff
@@ -19,71 +19,95 @@ class CreateTerm {
...
-llvm::Value *allocateTerm(llvm::Type *AllocType, llvm::BasicBlock *block, const char *allocFn = "koreAlloc");
-llvm::Value *allocateTerm(llvm::Type *AllocType, llvm::Value *Len, llvm::BasicBlock *block, const char *allocFn = "koreAlloc");
-}
+llvm::Value *allocateTerm(
+    llvm::Type *AllocType, llvm::BasicBlock *block,
+    const char *allocFn = "koreAlloc");
+llvm::Value *allocateTerm(
+    llvm::Type *AllocType, llvm::Value *Len, llvm::BasicBlock *block,
+    const char *allocFn = "koreAlloc");
+} // namespace kllvm
```
Baltoli, line 111: Declarations and definitions will get wrapped to respect the maximum line length. It's possible to specify one line per parameter if that's preferable

include/kllvm/codegen/Util.h
```diff
@@ -5,17 +5,18 @@
...
 
+// If Value is an instance of llvm::Function, cast and return. Otherwise, print
+// errors and abort.
+llvm::Function *castToFunctionOrAbort(llvm::Value *value);
 
 // getOrInsertFunction on module, aborting on failure
-template<class...Ts>
-static llvm::Function* getOrInsertFunction(llvm::Module *module, Ts... Args) {
+template <class... Ts>
+static llvm::Function *getOrInsertFunction(llvm::Module *module, Ts... Args) {
```
Baltoli, line 19: Whitespace and pointer alignment (as you identified)

include/kllvm/parser/location.h
```diff
@@ -19,55 +19,45 @@ class position {
     }
   }
 
-  void columns(int count = 1) {
-    column = add_(column, count, 1);
-  }
+  void columns(int count = 1) { column = add_(column, count, 1); }
```
Baltoli, line 22: Short functions in a class definition get put on one line. Configurable

include/runtime/arena.h
```diff
@@ -1,34 +1,37 @@
...
 // 127.
-#define REGISTER_ARENA(name, id) \
-  static struct arena name = { .allocation_semispace_id = id }
+#define REGISTER_ARENA(name, id)                                               \
+  static struct arena name = {.allocation_semispace_id = id}
 
-#define mem_block_start(ptr) \
-  ((char *)(((uintptr_t)(ptr) - 1) & ~(BLOCK_SIZE-1)))
+#define mem_block_start(ptr)                                                   \
+  ((char *)(((uintptr_t)(ptr)-1) & ~(BLOCK_SIZE - 1)))
```
Baltoli, line 34: It will keep macro line continuations right-aligned

include/runtime/header.h
```diff
@@ -21,112 +21,107 @@
...
+#define reset_gc(s)                                                            \
+  ((s)->h.hdr = (s)->h.hdr & ~(NOT_YOUNG_OBJECT_BIT | AGE_MASK | FWD_PTR_BIT))
+#define struct_base(struct_type, member_name, member_addr)                     \
+  ((struct_type *)((char *)(member_addr)-offsetof(struct_type, member_name)))
 #define leaf_block(tag) ((block *)((((uint64_t)(tag)) << 32) | 1))
 #define variable_block(tag) ((block *)((((uint64_t)(tag)) << 32) | 3))
 #define is_leaf_block(block) (((uintptr_t)block) & 1)
 #define is_variable_block(block) ((((uintptr_t)block) & 3) == 3)
 
 extern "C" {
```
Baltoli, line 43: Block alignment; the default is not to indent namespace nor extern contents

lib/ast/AST.cpp
```diff
@@ -132,34 +136,46 @@ ValueType KORECompositeSort::getCategory(KOREDefinition *definition) {
...
+  else if (name == "FLOAT.Float")
+    category = SortCategory::Float;
+  else if (name == "BUFFER.StringBuffer")
+    category = SortCategory::StringBuffer;
+  else if (name == "BOOL.Bool")
+    category = SortCategory::Bool;
+  else if (name == "KVAR.KVar")
+    category = SortCategory::Variable;
+  // we expect the "hook" of a MInt to be of the form "MINT.MInt N" for some
+  // bitwidth N
```
Baltoli, line 173: This was one of the only places I thought you might strongly prefer the pre-formatted version. There's an option in the next release of clang-format that would get you the one-line version; for now I can mark this block as not for formatting, or leave it as it is.

lib/ast/AST.cpp
```diff
@@ -918,18 +915,25 @@ sptr<KOREPattern> KORECompositePattern::filterSubstitution(PrettyPrintData const
       indent = oldIndent;
       atNewLine = oldAtNewLine;
       std::string name = ss.str();
-      if (vars.count(var->getName()) && (name[0] == '_' || (name.size() > 1 && (name[0] == '@' || name[0] == '!' || name[0] == '?') && name[1] == '_'))) {
+      if (vars.count(var->getName())
+          && (name[0] == '_'
+              || (name.size() > 1
+                  && (name[0] == '@' || name[0] == '!' || name[0] == '?')
+                  && name[1] == '_'))) {
```
Baltoli, line 922: Extra indentation to reflect the nesting of this condition

runtime/alloc/alloc.cpp
```diff
@@ -1,13 +1,13 @@
...
 #include <stdio.h>
 #include <stdlib.h>
-#include <stdint.h>
-#include <stdbool.h>
+#include <string.h>
 
 #include "runtime/alloc.h"
-#include "runtime/header.h"
 #include "runtime/arena.h"
+#include "runtime/header.h"
```
Baltoli, line 11: Includes are sorted and grouped

dwightguth (approved):
I'm approving this. I don't feel particularly strongly about any of the changes that just reorganize line breaks and whitespace, and the change to sort the includes is reasonable to me. I'm going to merge this now so that we can make sure to apply formatting to future PRs. Please go ahead and work as soon as you can on setting up the CI check.

PR #431, opened 2021-09-24
Enable collection during allocation

dwightguth:
We add a call to koreCollect in `arenaAlloc` if we are close to running out of free space, and only allocate a new block if we're still out of space after the collection. We also modify the behavior of the alwaysgcspace so that it is cleared after each rewrite step.

This ought to ensure that there is never a long period of time where collection cannot occur, since collection can occur during any allocation with the exception of some allocations from inside hooks. It also makes it much less likely that the application will allocate more blocks than are strictly needed, thus reducing memory usage. Additionally, since we are no longer periodically making an expensive check about whether or not to perform a collection, there is a slight improvement in runtime performance.

Limitations: we still only collect things allocated via koreAllocAlwaysGC after each rewrite step. This needs to be improved in future PRs.

### DIFF
--- a/lib/codegen/Decision.cpp
+++ b/lib/codegen/Decision.cpp
@@ -643,14 +643,7 @@ void makeEvalOrAnywhereFunction(
   initChoiceBuffer(
       dt, module, block, stuck, fail, &choiceBuffer, &choiceDepth, &jump);
 
-  auto koreCollect = getOrInsertFunction(
-      module, "tryKoreCollect",
-      llvm::FunctionType::get(
-          llvm::Type::getVoidTy(module->getContext()),
-          {llvm::Type::getInt1Ty(module->getContext())}, false));
-  llvm::CallInst::Create(
-      koreCollect, {llvm::ConstantInt::getFalse(module->getContext())}, "",
-      block);
+  insertCallToGC(block, false);
 
   int i = 0;
   Decision codegen(
@@ -800,15 +793,7 @@ llvm::BasicBlock *stepFunctionHeader(
       module->getContext(), "step", block->getParent());
   llvm::BranchInst::Create(stuck, merge, isFinished, block);
 
-  auto koreCollect = getOrInsertFunction(
-      module, "tryKoreCollect",
-      llvm::FunctionType::get(
-          llvm::Type::getVoidTy(module->getContext()),
-          {llvm::Type::getInt1Ty(module->getContext())}, false));
-  auto call = llvm::CallInst::Create(
-      koreCollect, {llvm::ConstantInt::getTrue(module->getContext())}, "",
-      merge);
-  setDebugLoc(call);
+  insertCallToGC(merge, true);
   return merge;
 }
 
--- a/lib/codegen/Decision.cpp
+++ b/lib/codegen/Decision.cpp
@@ -882,10 +882,7 @@ void makeStepFunction(
     auto result = new llvm::LoadInst(bufType, resultBuffer, "", stuck);
     llvm::ReturnInst::Create(module->getContext(), result, stuck);
   } else {
-    auto phi = llvm::PHINode::Create(val->getType(), 2, "phi_1", stuck);
-    phi->addIncoming(val, block);
-    phi->addIncoming(val, pre_stuck);
-    llvm::ReturnInst::Create(module->getContext(), phi, stuck);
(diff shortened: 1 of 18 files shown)


runtime/collect/StackMapParser.cpp
```diff
@@ -39,26 +39,35 @@ void parseStackMap() {
       void *ip = (void *)(func.FunctionAddress + InstructionOffset);
       int32_t NumDeopts = *(int32_t *)(stackMapRecord + 24 + 2 * 12);
       uint16_t RelocationOffset = 0;
-      for (uint16_t j = 4 + NumDeopts; j < NumLocations; j += 2) {
-        uint8_t type = *(uint8_t *)(stackMapRecord + 16 + j * 12);
-        if (type == 5) {
+      for (uint16_t j = 3 + NumDeopts; j < NumLocations; j += 2) {
```
Baltoli, line 42: What part of the schema is this looking at, and what's the difference between offset 4 and offset 3? Some further comments tying this parser to [the schema](https://llvm.org/docs/StackMaps.html#stack-map-format) would be nice - I have no doubt it's correct, but it's difficult to follow the byte arithmetic to make sure it's correct.

runtime/collect/StackMapParser.cpp
```diff
@@ -39,26 +39,35 @@ void parseStackMap() {
...
           // thus do not need to be relocated.
           RelocationOffset++;
           continue;
         }
-        if (type != 3) {
+        if (base_type != 3) {
           abort();
         }
-        int32_t offset = *(int32_t *)(stackMapRecord + 24 + j * 12);
+        uint8_t derived_type = *(uint8_t *)(stackMapRecord + 16 + (j + 1) * 12);
```
Baltoli, line 55: For example, as above this kind of thing is difficult to grok - a comment mapping the pointer computation to a path into the schema would be good (e.g. for the sake of demonstration on a contrived example not related to this particular line, something like `StkMapRecord[i].Location[j].DwarfRegNum` would clarify exactly what bytes we're expecting to access.

There is a bit of an additional burden involved with keeping this in sync, I guess - how often is this code likely to be modified? If it's only once in a while, then probably worth the effort of keeping the comments.

Baltoli (changes requested):

runtime/collect/StackMapParser.cpp
dwightguth: Good call about the comments. I'll go ahead and go through this function and add them.

lib/codegen/Decision.cpp
```diff
@@ -888,23 +887,19 @@ void makeStepFunction(
...
   if (search) {
     auto result = new llvm::LoadInst(bufType, resultBuffer, "", stuck);
     llvm::ReturnInst::Create(module->getContext(), result, stuck);
   } else {
-    auto phi
-        = llvm::PHINode::Create(collectedVal->getType(), 2, "phi_1", stuck);
+    auto phi = llvm::PHINode::Create(val->getType(), 2, "phi_1", stuck);
     phi->addIncoming(val, block);
-    phi->addIncoming(collectedVal, pre_stuck);
+    phi->addIncoming(val, pre_stuck);
```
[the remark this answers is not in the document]
dwightguth: Not really. It was just a lazy update to the function. I have taken this phi node out and it seems to still pass.

Baltoli (approved):
Great - the comments on the stack map parser are really helpful for understanding what's going on.

theo25 (approved):
This looks good to me.

theo25 (approved):

lib/llvm/EmitGCLayoutInfo.cpp
```diff
@@ -64,7 +64,7 @@ struct EmitGCLayoutInfo : public ModulePass {
               if (!Ty->isStructTy()) {
                 error(Ty);
               }
-              auto StructTy = llvm::cast<StructType>(Ty);
+              auto StructTy = cast<StructType>(Ty);
               if (!StructTy->hasName()) {
                 error(StructTy);
               }
```
Baltoli, line 70: This whole stanza could be a bit more succinct:

```c++
auto StructTy = dyn_cast<StructType>(Ty);
if (!StructTy || !StructTy->hasName()) {
  error(StructTy);
}
```

Baltoli (approved):
One small change, but otherwise looks good to me.

PR #443, opened 2021-10-20
Initial build support for AS

Baltoli:
This PR implements the build system changes needed to get the project compiling on ARM macOS. There are two main parts to the change:
* Set the Homebrew root directory by asking `brew --prefix` where it's located, rather than hard-coding `/usr/local`.[^1]
* Configure LLVM source code to get the target triple at build time, rather than hard-coding it.

[^1]: There's also a _very_ niche benefit to this - that we can support custom brew installations not in the usual place.


Baltoli:
@dwightguth I've automated the hard-coded parts of the build as you suggested; the triple now comes from `llvm-config` and the datalayout is scraped from clang.

dwightguth (approved):
Excellent!

PR #449, opened 2021-11-01
Bump MPFR version

Baltoli:
This updates to follow the deployment of a new version of MPFR; see https://github.com/kframework/mpfr-java/pull/16

### DIFF
--- a/matching/pom.xml
+++ b/matching/pom.xml
@@ -45,7 +45,20 @@
     <dependency>
       <groupId>org.kframework.mpfr_java</groupId>
       <artifactId>mpfr_java</artifactId>
-      <version>1.3</version>
+      <version>0.2</version>
+    </dependency>
+
+    <dependency>
+      <groupId>org.kframework.mpfr_java</groupId>
+      <artifactId>mpfr_java</artifactId>
+      <version>0.2</version>
+      <classifier>linux64</classifier>
+    </dependency>
+    <dependency>
+      <groupId>org.kframework.mpfr_java</groupId>
+      <artifactId>mpfr_java</artifactId>
+      <version>0.2</version>
+      <classifier>osx</classifier>
     </dependency>
   </dependencies>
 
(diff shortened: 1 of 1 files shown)


matching/pom.xml
```diff
@@ -133,4 +114,96 @@
...
+      <dependencies>
+        <dependency>
+          <groupId>org.kframework.mpfr_java</groupId>
+          <artifactId>mpfr_java</artifactId>
+          <version>1.3</version>
+          <classifier>linux64</classifier>
+        </dependency>
+      </dependencies>
+
+      <properties>
```
dwightguth, line 138: I'm pretty sure you don't need these because it should inherit them from the parent pom.

dwightguth (changes requested):
It looks like your attempt to make maven only depend on the host os's version of the dependency has screwed up the nix lock file. Can you explain what prompted you to make that change?

matching/pom.xml
Baltoli: The "host OS only change" is in response to a problem with the way `hawtjni` packages the native code. It allows you to attach whatever classifier you like to the JAR filename, but requires that the _internal_ structure of the JAR uses its naming structure.

This produces JARs that look something like:
```
osx-arm.jar
|-- META-INF
  |-- osx64
    |-- native
      |-- lib.dylib [arm library]

osx-x86.jar
|-- META-INF
  |-- osx64
    |-- native
      |-- lib.dylib [x86 library]
```

When the two dependencies are _both_ specified, the two copies of `lib.dylib` collide, and we end up potentially selecting the wrong version. If you depend on only one at a time, the problem goes away.

The options I considered to fix the problem were:
1. Modify the `hawtjni` packaging code (impractical, more trouble than it's worth)
2. Modify _our_ packaging code (i.e. the copying of native libraries from JARs. I suspect this might also be a hassle).
3. Select only one native dependency at Maven time. This also appears to be an issue, as you identify.

Re. the parent POM, my understanding is that `matching` gets built even if the backend isn't part of the K distribution repo. The dependency was there beforehand, and I've just reconfigured it.

Let me know what you reckon the best way to resolve these issues is.

PR #454, opened 2021-11-05
Pick up the correct environment variable

Baltoli:
Matches https://github.com/kframework/homebrew-k/pull/10 in the Homebrew formula.

My previous change made a silly error and didn't pick up the correct prefix variable; this change should fix that and get the actual homebrew root (`/usr/local` rather than `/usr/local/Cellar/kframework/5.1.246`, say).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,7 +30,7 @@ set(CMAKE_EXE_LINKER_FLAGS_RELWITHDEBINFO "${CMAKE_EXE_LINKER_FLAGS_RELEASE} -fl
 
 if(APPLE)
   if(NOT USE_NIX)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -31,7 +31,7 @@ set(CMAKE_EXE_LINKER_FLAGS_RELWITHDEBINFO "${CMAKE_EXE_LINKER_FLAGS_RELEASE} -fl
 if(APPLE)
   if(NOT USE_NIX)
     if(DEFINED ENV{PREFIX})
```
dwightguth, line 33: shouldn't this be modified too?

Baltoli: Good catch!

dwightguth (approved):

PR #456, opened 2021-11-05
Update to current draft of upstream immer PR

dwightguth:
Definitely not ready for review as the upstream PR may still change dramatically. Just testing.


dwightguth:
I'm going to close this because the upstream pr has further changes that need to be made and I am no longer working on this. I'll let @theo25 open his own pr whenever he gets around to picking up the immer tasks.

PR #458, opened 2021-11-09
Static Terms with new GC

gtrepta:


dwightguth:
I'd really actually like to see this branch have a merge commit rather than cherry picking the commit from master. That will keep the history much cleaner once this branch is merged back into master.

gtrepta:
@dwightguth It's a merge commit now.

dwightguth (approved):
I'm approving. Please don't merge, I'm going to do that on the command line in order to preserve the history.

PR #459, opened 2021-11-10
Fix for crash on Apple Silicon

Baltoli:
This PR cherry-picks a change from the `new-gc` branch by hand to fix a crash with LLVM `--search` on Apple Silicon.

I'm not exactly sure what the cause of this bug is, but I suspect that it's some kind of subtle issue with the interaction between generated LLVM and the C code in the runtime library. The ["packed traversals" fix](https://github.com/kframework/llvm-backend/pull/446) from a while ago was a similar-looking issue.

Pulling in this change (that reduces the amount and complexity of generated code used) fixes the problem, and given that we were planning to make this change anyway I'm inclined to say that it's the most sensible solution.

I can investigate in more detail, but I'm not sure that's a worthwhile use of time given that this is a) only on Apple Silicon and b) fixed neatly by this PR.

The original change by @dwightguth is [here](https://github.com/kframework/llvm-backend/pull/441/commits/338d55fed3eb7c8ca19c43f8d916acabf69a33f8); this isn't a direct cherry-pick as the original commit is on a squashed branch.

Fixes https://github.com/kframework/k/issues/2311

### DIFF
--- a/runtime/util/search.cpp
+++ b/runtime/util/search.cpp
@@ -10,8 +10,7 @@ static std::vector<block *> stepResults;
 
 extern "C" {
 
-void addSearchResult(
-    block *result, block ***bufPtr, uint64_t *count, uint64_t *capacity) {
+void addSearchResult(block *result) {
   stepResults.push_back(result);
 }
 
--- a/runtime/util/search.cpp
+++ b/runtime/util/search.cpp
@@ -6,22 +6,16 @@
 #include "runtime/collect.h"
 #include "runtime/header.h"
 
+static std::vector<block *> stepResults;
+
 extern "C" {
 
 void addSearchResult(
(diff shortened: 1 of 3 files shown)


runtime/util/search.cpp
```diff
@@ -6,22 +6,16 @@
 #include "runtime/collect.h"
 #include "runtime/header.h"
 
+static std::vector<block *> stepResults;
+
 extern "C" {
 
 void addSearchResult(
     block *result, block ***bufPtr, uint64_t *count, uint64_t *capacity) {
```
dwightguth, line 14: I dunno if this is because the new_gc branch forgot to do this, or the cherry pick was incomplete, but the type of this function should match the declaration in Decision.cpp

dwightguth (changes requested):

dwightguth (approved):

PR #460, opened 2021-11-16
Cleanup cmake and nix

jonringer:
Eventually I would like to flakify this repository. However, thought the scope of that would be a little too big. So I'm creating this PR which should be a smaller refactor of the existing code. But sill be much more compatible with a flake as most of the complexity is delegated to overlays, rather than having a lot of logic in default.nix

cc @ttuegel

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -133,16 +133,8 @@ install(
   DESTINATION lib/kllvm/llvm/main
 )
 
-find_program(OPT opt-13)
-find_program(OPT opt-12)
-find_program(OPT opt-11)
-find_program(OPT opt-10)
-find_program(OPT opt)
-find_program(LLC llc-13)
-find_program(LLC llc-12)
-find_program(LLC llc-11)
-find_program(LLC llc-10)
-find_program(LLC llc)
+find_program(OPT NAMES opt-13 opt-12 opt-11 opt-10 opt)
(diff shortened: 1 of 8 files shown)


dwightguth:
I am not a Nix expert and Tom Tuegel is no longer at Runtime Verification, but IOG is the primary consumer of our nix expressions, and the changes to cmake seem fine to me, so if these changes to the nix files are acceptable to me, I have no objections. I would like to have some confirmation that the modified code is compatible with downstream nix expressions before this is merged though. Did you test the k frontend and iele?

jonringer:
not sure why the nix:maven step is paused, but this produces the same derivations that were available before. (this shouldn't have changed anything that other repositories expect.)

Not sure who else can test nix changes.

dwightguth:
Try pushing to the branch again. It's possible because that job runs on pushes and not prs that it didn't interact intelligently with the GitHub logic for requiring approval to run workflows from forks.

dwightguth:
It seems that this didn't fix the issue about the maven check. I did some digging and it seems that workflow doesn't support prs from forks anyway. We probably need to address this somehow eventually, but for now, I'm going to approve this and merge it despite the yellow check, once the other checks pass, because that check wouldn't do anything anyway to a pr that doesn't touch maven at all.

dwightguth (approved):

Issue #461, opened 2021-11-26
Segfault in search mode


vasil-sd:
llvm-backend commit: 72cf698be6baef5cdd01aa9239e32e9be89ec6a
k-framework commit: 97358b03296b6f2f6575c432906d6c1fd8043160
clang --version:
```
clang version 13.0.0
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
```

simple-untyped.k (minimized as much as I could):
```
module SIMPLE-UNTYPED-SYNTAX
  imports DOMAINS-SYNTAX

  syntax Id ::= "main" [token]

  syntax Stmt ::= "var" Exp ";"
                | "var" Exps ";" [macro-rec, avoid]
                | "function" Id "(" Ids ")" Block

  syntax Exp ::= Int | Bool | String | Id
               | "(" Exp ")"             [bracket]
               | "++" Exp
               > Exp "[" Exps "]"        [strict]
               > Exp "(" Exps ")"        [strict]
               | "-" Exp                 [strict]
               | "sizeOf" "(" Exp ")"    [strict]
               | "read" "(" ")"
               > left:
                 Exp "*" Exp             [strict, left]
               | Exp "/" Exp             [strict, left]
               | Exp "%" Exp             [strict, left]
               > left:
                 Exp "+" Exp             [strict, left]
               | Exp "-" Exp             [strict, left]
               > non-assoc:
                 Exp "<" Exp             [strict, non-assoc]
               | Exp "<=" Exp            [strict, non-assoc]
               | Exp ">" Exp             [strict, non-assoc]
               | Exp ">=" Exp            [strict, non-assoc]
               | Exp "==" Exp            [strict, non-assoc]
               | Exp "!=" Exp            [strict, non-assoc]
               > "!" Exp                 [strict]
               > left:
                 Exp "&&" Exp            [strict(1), left]
               | Exp "||" Exp            [strict(1), left]
               > "spawn" Block
               > Exp "=" Exp             [strict(2), right]

  syntax Ids  ::= List{Id,","}           [klabel(exps), prefer]
  syntax Exps ::= List{Exp,","}          [klabel(exps), strict]  // automatically hybrid now
  syntax Exps ::= Ids                    [prefer]
  syntax Val
  syntax Vals ::= List{Val,","}          [klabel(exps)]
  syntax Bottom
  syntax Bottoms ::= List{Bottom,","}    [klabel(exps)]
  syntax Ids ::= Bottoms
[...]

PR #462, opened 2021-11-30
Add pass manager flag to llvm-kompile script

Baltoli:
This flag opts into the old pass manager infrastructure on newer versions of LLVM (>= 13). See the discussion [here](https://groups.google.com/g/llvm-dev/c/kQYV9dCAfSg) for context.

### DIFF
--- a/bin/llvm-kompile
+++ b/bin/llvm-kompile
@@ -49,7 +49,7 @@ if $compile; then
   "$(dirname "$0")"/llvm-kompile-codegen "$definition" "$dt_dir"/dt.yaml "$dt_dir" $debug > "$mod"
 
   # optimization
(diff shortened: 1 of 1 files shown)


dwightguth:
I'd prefer we just make it so that it also works with the new pass manager; that's likely to be a lot more future proof than this change. Can you tell me what your motivation for this is, though? The new_gc branch is definitely not ready for public use yet.

Baltoli:
Fair enough - I wanted to see if `new_gc` would fix some of Vasil's test cases on #461, and ran into this issue. I'd pushed the branch to save him the bother of fixing it himself were he on LLVM 13, so thought I'd put in the PR in case it was useful. Happy not to merge in favour of future-proofing our passes, though

PR #463, opened 2021-12-01
Collect `states_set` alongside `states` in search

Baltoli:
As discussed, this is the fix for #461 that promotes the `states_set` variable to a global so that it can be collected in the same way as `states`.


dwightguth (approved):
This looks fine as far as it goes, but do we need to do the same thing for the result set?

Baltoli:
Oh yeah of course - I missed this code:
```c++
if (stepResults.size() == 0) {
  results.insert(state);
} else {
```

and read wrongly that the results were only collected at the end. I'll add that change now

dwightguth (approved):

PR #465, opened 2021-12-02
Make unit testing more idiomatic

Baltoli:
CMake has built-in support for unit test frameworks; this PR rewrites the unit testing build script to use this support more idiomatically. The changes are:
* Instead of excluding unit-test targets from the default build target, the inclusion of the unit tests subdirectory is gated behind the CMake configuration variable `BUILD_TESTS`. If the option is enabled, `make` will build the unit tests as well as the backend itself.
* Rather than manually tracking dependencies between a phony `run-unittests` target and the individual executables, use `add_test` to get CMake to do it for us.
* Update the CI script to use the new invocations.


Baltoli:
@dwightguth You can see a unit test failure branched from this PR in #468

PR #466, opened 2021-12-02
Convert integration tests to use `lit`

Baltoli:
These changes mirror the ones made to the C semantics:
* Add LLVM's `lit` package to the Dockerfile so that it's installed and available for the CI runner. This uses the same steps as the C semantics to install `lit`.
* Convert the tests in `defn` and `unparse` to be run as individual `lit` cases, rather than from root Makefiles.
* Small changes to the CI runner script to deal with files being generated into different directories.
* Add an example `envrc.sh` as we have done for other projects; this handles setting your paths appropriately.


Baltoli:
A note on converting some of the tests in `defn` - they appear not to finish rewriting (exit code 113). They still produce output on `stdout` that matches their reference output; my assumption is that this behaviour is intentional, but I wanted to check that was the case. The solution is adding `%allow-pipefail` to allow the interpreter execution to fail.

Baltoli:
This is blocked until I can bump the version of `lit` that nixpgs references: https://github.com/NixOS/nixpkgs/blob/a0dbe47318bbab7559ffbfa7c4872a517833409f/pkgs/development/tools/misc/lit/default.nix

Baltoli:
@dwightguth This is ready for review now that I've fixed the Nix issues from last week.

ciscript
```diff
@@ -37,18 +37,14 @@ cd ../build
...
-cp ../test/Makefile ./TestMakefile
+TEST_DIR=../test
 
-make -f TestMakefile -j`nproc`
+lit $TEST_DIR
 
-diff io_output1.txt ../test/ioTest1.output
-diff `ls | grep io_log` ../test/ioTest3.output
+cd $TEST_DIR/defn
+diff io_output1.txt ../ioTest1.output
```
dwightguth, line 45: These checks can probably be inlined into the respective test now.

test/lit.cfg.py
```diff
@@ -0,0 +1,57 @@
...
+config.name = 'llvm-backend'
+config.test_source_root = os.path.join(ROOT_PATH, "test")
+
+config.test_format = formats.ShTest(True)
+
+config.suffixes = ['.kore']
+
+config.excludes = []
+
+if os.getenv('LIT_USE_NIX'):
```
dwightguth, line 16: Can we get a comment here about what this is doing?

test/unparse/exp/lit.local.cfg.py
```diff
@@ -0,0 +1,4 @@
+config.excludes = [
```
dwightguth, line 1: Should we move these into the parent directory somehow since it's the same across each directory?

dwightguth (changes requested):
Overall this looks good, but I'd like to see a few minor changes

ciscript
Baltoli: Will do.

test/lit.cfg.py
Baltoli: Will comment it - the problem is basically `lit` not inheriting the environment when it's launched, which of course breaks nix derivations.

Baltoli:
@dwightguth The manual section of `ciscript` is now inlined (with cleanup) into the run scripts for `test-io` and `test28` respectively.

dwightguth (approved):
Great work, thanks!

PR #470, opened 2021-12-07
Deduplicate search results after search result filtering

dwightguth:
We automatically remove conjuncts from substitutions when running kprint that are simply substitutions for variables that were marked as anonymous by the user. However, previously we were not deduplicating the list of search results after doing so, even though this might result in multiple identical disjuncts. The test I added is from lamda++ lesson 5 in the pl-tutorial in the frontend, and tests that the disjuncts are successfully deduped.


Baltoli:
@dwightguth It looks like this breaks some tests - I don't know if it's just a question of needing to propagate the deduplication to those outputs, or if there's more to fix. Will review properly when the CI is fixed

lib/ast/AST.cpp
```diff
@@ -902,6 +902,31 @@ std::map<std::string, int> KORECompositePattern::gatherVarCounts(void) {
...
+  std::vector<sptr<KOREPattern>> items, dedupedItems;
+  flatten(this, "\\or", items);
+  std::set<std::string> printed;
+  for (sptr<KOREPattern> item : items) {
+    std::ostringstream Out;
+    item->print(Out);
+    if (printed.insert(Out.str()).second) {
+      dedupedItems.push_back(item);
+    }
+  }
```
Baltoli, line 918: It may well be the best way to do it, but I'm curious why we don't have some kind of structural equality on these patterns other than the one induced by printing them.

Baltoli (approved):

lib/ast/AST.cpp
dwightguth: Just was never implemented. This doesn't seem to be particularly performance intensive and it was a convenient way to implement it, so I went ahead with it. I can try to implement an operator== if you'd rather?

PR #474, opened 2021-12-16
Fix bugs involving bytes literals

dwightguth:
This PR fixes a couple different issues that stand in the way of supporting bytes literals in the llvm backend.

Namely:

1. Fixes a bug where decision tree generation emits mojibake into the decision tree if binary data is included in a bytes literal on the lhs. This is not a full fix for the issue; the same issue occurs with strings, but in order to fix the issue with strings, we would have to support unicode in the llvm backend, which we do not yet do.
2. Fixes a bug where binary data serialized by snakeyaml was incorrectly deserialized by libyaml. We continue to use the old deserialization when generating names for occurrences, because occurrences can appear in basic block labels and binary data can contain null bytes, but null bytes cannot occur in basic block labels. however, we now correctly emit the correct actual string data, so it should be fine.


.github/workflows/clang-format-check.yml
```diff
@@ -10,4 +10,4 @@ jobs:
       uses: jidicula/clang-format-action@v4.0.0
       with:
         clang-format-version: '12'
-        exclude-regex: '(build|config|deps)'
+        exclude-regex: '(build|config|deps|base64)'
```
Baltoli, line 13: Might the `base64` library live in the `deps/` tree, instead of directly in the `include/` path? It would be cleaner to keep all third-party code in the same place. It would also be pretty easy to wrap the library as its own CMake target, which would make the build system code a little cleaner. I'm happy to write these changes up as a PR to this branch if you like.

Baltoli (changes requested):
The code itself looks fine to me, but I have a comment regarding the build system setup with the base64 library.

.github/workflows/clang-format-check.yml
dwightguth: Yeah, if you think that's appropriate, go ahead. This seemed like the simplest way to me, but if you want to go through that process, go ahead and I'll merge it into this branch.

Baltoli (approved):

PR #475, opened 2022-01-10
Implement check for heap blocks

Baltoli:
This implements a macro to check for heap (vs., for example, static terms) blocks. By doing so, we can then avoid trying to GC non-heap-allocated blocks.

This appears to fix the issue identified in https://github.com/runtimeverification/c-semantics/issues/472

The solution here was proposed by @dwightguth [here](https://runtimeverification.slack.com/archives/C5GE8MBC6/p1639589383188700).


Baltoli:
I've tested this successfully against:
* The C semantics test suite, where the original problem was first picked up. I can no longer reproduce the intermittent segfaults from https://github.com/runtimeverification/c-semantics/issues/472

include/runtime/header.h
```diff
@@ -39,6 +39,9 @@
 #define variable_block(tag) ((block *)((((uint64_t)(tag)) << 32) | 3))
 #define is_leaf_block(block) (((uintptr_t)block) & 1)
 #define is_variable_block(block) ((((uintptr_t)block) & 3) == 3)
+#define is_heap_block(s)                                                       \
+  ((((s)->h.hdr) & AGE_MASK) || (!(((s)->h.hdr) & NOT_YOUNG_OBJECT_BIT))       \
```
dwightguth, line 43: the checks for AGE_MASK and FWD_PTR_BIT should be redundant, I believe.

Baltoli: Great, thanks - I wasn't sure of the exact semantics of these bits. I'll correct the check to reflect this.

dwightguth:
I think I may have led you astray; I apologize. While it is true that all non-heap blocks must have the NOT_YOUNG_OBJECT_BIT set, some heap blocks will have it set as well, so this code is too aggressive now. In order to write a condition which is only true of non-heap-blocks, the NOT_YOUNG_OBJECT_BIT must be set and the AGE_MASK must be not set

dwightguth (approved):

PR #482, opened 2022-02-11
Inlined pretty printer in .debug_gdb_scripts section

gtrepta:
The `.debug_gdb_scripts` section has been hardcoding an absolute path to a pretty printing python script which gets determined at compile time using the install prefix. This creates issues when installing with the `DESTDIR` variable which a lot of semantics repos use. This PR inlines the script into the section to fix that issue.


dwightguth (approved):
Looks good. I assume you tested by starting gdb and pretty printing some terms. If so, go ahead and merge.

PR #484, opened 2022-02-16
Initial binary KORE format

Baltoli:
This PR represents the initial work towards a binary representation for KORE terms.

The changes are:
* Design and implement a concatenative binary representation for KORE terms; `docs/binary_kore.md` contains a description of the schema. The `serializer` and `deserializer` components implement an emitter and parser for this format.
* Implement a tool (`kore-convert`) that converts between textual and binary KORE representations.
* Adapt the existing test suite to provide tests that verify the serializer / deserializer implementations (by ensuring that conversion to binary and back behaves identically to parsing, then printing a KORE term).
* Test composition of binary terms.
* Allows compiled interpreters to accept binary KORE terms as input, and has the test suite run each test in `defn` with both textual and the equivalent binary input.

I'd like to get reviews on this PR before spending too much time moving forward with the next steps:
* ~~Allow the backend to accept binary terms as input by modifying the configuration parser.~~
* Allow the backend to produce binary terms directly from a runtime configuration.
* Tool support + integration with `llvm-krun` etc.

There are also opportunities for compressing these terms that I haven't yet implemented (so as to keep the size of this PR down):
* Term sharing (possibly expensive at emission-time based on the KAST serializer; might need LRU cache)
* Indexing based on a definition (may need a different tool interface)


lib/ast/AST.cpp
```diff
@@ -1829,6 +1831,72 @@ void KOREDefinition::print(std::ostream &Out, unsigned indent) const {
...
+
+void KORECompositePattern::serialize_to(serializer &s) const {
+  for (auto const &arg : arguments) {
+    arg->serialize_to(s);
+  }
+
+  constructor->serialize_to(s);
+
+  s.emit(header_byte<KORECompositePattern>);
+  s.emit(int16_t(arguments.size()));
```
theo25, line 1850: I would like a graceful failure in case we have a number of arguments that cannot fit in an `int16_t`.

lib/ast/AST.cpp
```diff
@@ -1829,6 +1831,72 @@ void KOREDefinition::print(std::ostream &Out, unsigned indent) const {
...
+  s.emit_string(name);
+}
+
+void KORECompositeSort::serialize_to(serializer &s) const {
+  for (auto const &arg : arguments) {
+    arg->serialize_to(s);
+  }
+
+  s.emit(header_byte<KORECompositeSort>);
+  s.emit(int16_t(arguments.size()));
```
theo25, line 1869: Similarly here.

lib/ast/AST.cpp
```diff
@@ -1829,6 +1831,72 @@ void KOREDefinition::print(std::ostream &Out, unsigned indent) const {
...
+  }
+
+  if (sort) {
+    sort->serialize_to(s);
+  }
+
+  s.emit(header_byte<KORESymbol>);
+
+  s.emit(int16_t(arguments.size()));
+  s.emit(int16_t(formalArguments.size()));
```
theo25, line 1889: Similarly here.

theo25 (commented):
I left a handful of comments for you to review.

Baltoli:
Some representative file sizes for large KORE terms extracted from the C linking semantics:

| file | text | binary | ratio |
| --- | --- | --- | --- |
| 0 | 111515863 | 50954329 | 0.457 |
| 1 | 261349946 | 121194326 | 0.464 |
| 2 | 86917040 | 40598393 | 0.467 |
| 3 | 204318254 | 93657437 | 0.458 |
| 4 | 52232598 | 22932327 | 0.439 |
| 5 | 96386778 | 44017782 | 0.457 |
| 6 | 34833151 | 16103558 | 0.462 |
| 7 | 12066690 | 5618967 | 0.466 |
| 8[^1] | 620512295 | - | - |
| 9 | 322499023 | 147555533 | 0.458 |

[^1]: I wasn't able to parse this text file on my machine due to OOM.

Baltoli:
Timings[^1] for converting the same files between text and binary, all quoted in milliseconds:

| file | parse text | output binary | parse binary | output text | parse speedup |
| :-- | --: | --: | --: | --: | --: |
| 0 | 5074 | 2878 | 2952 | 1068 | 1.719 |
| 1 | 11625 | 6910 | 7033 | 2297 | 1.653 |
| 2 | 3926 | 2323 | 2362 | 833 | 1.662 |
| 3 | 9065 | 5311 | 5430 | 1868 | 1.669 |
| 4 | 2265 | 1318 | 1376 | 502 | 1.646 |
| 5 | 4320 | 2520 | 2557 | 910 | 1.689 |
| 6 | 1576 | 917 | 943 | 325 | 1.671 |
| 7 | 571 | 326 | 328 | 109 | 1.741 |
| 8[^2] | - | - | - | - | - |
| 9 | 14322 | 9121 | 8488 | 3014 | 1.687 |
| geomean | | | | | **1.682** |

[^1]: Obtained somewhat unscientifically; there isn't a huge amount of variation in the timings to warrant a rigorous examination.
[^2]: As above, I can't parse this one

Baltoli:
Memory usage[^1] (in kb) for text-binary and binary-text conversions:

| file | text-binary | binary-text |
| :-- | --: | --: |
| 0 | 972400 | 943528 |
| 1 | 2257144 | 2225244 |
| 2 | 788080 | 749296 |
| 3 | 1785548 | 1729908 |
| 4 | 452464 | 431868 |
| 5 | 850896 | 815848 |
| 6 | 312992 | 303108 |
| 7 | 118312 | 111156 |
| 8 | - | - |
| 9 | 2866420 | 2734244 |

As expected, the size of the AST itself seems to be the dominant factor in parsing the text / binary KORE formats.

[^1]: Maximum resident set size as reported by `/usr/bin/time -v`

Baltoli:
Some memory usage benchmarking with the new change to allow the interpreter to accept a binary file as input:

| Step | Size / Memory Usage (kb) |
| :-- | --: |
| text file size | 2747940 |
| binary file size | 1255216 |
| text to binary conversion | 24722424 |
| execute from text | 25486340 |
| execute from binary | 8500332 |

This is a factor of 3 reduction in peak memory usage with the binary format loaded directly into the backend's runtime structures.

lib/binary/serializer.cpp
```diff
@@ -24,22 +27,32 @@ void serializer::emit(std::byte b) {
 
 void serializer::emit_string(std::string const &s) {
   if (intern_table_.find(s) == intern_table_.end()) {
-    emit(direct_string_prefix_);
-
-    intern_table_[s] = next_idx_;
+    emit_direct_string(s);
+  } else {
+    int64_t previous = intern_table_.at(s);
+    int64_t diff = (next_idx_ + 5) - previous;
```
theo25, line 40: I think I am missing something. Could you explain the reason for changing `next_idx_ + 4` to `next_idx_ + 5`?

lib/binary/serializer.cpp
```diff
@@ -1,10 +1,17 @@
...
 
+#include <arpa/inet.h>
 #include <cassert>
 #include <limits>
 
 namespace kllvm {
 
+namespace detail {
+bool is_big_endian() {
+  return htonl(365) == 365;
```
theo25, line 11: I am not very familiar with the `arpa/inet.h` header but it seems to be OS specific. An easier way to determine endianess is something like this:
```
inline bool is_big_endian() {
  uint32_t i = 1;
  uint8_t *c = (uint8_t*)&i;
  if (*c)
        return false;
    else
        return true;
}
```
Another advantage of this code is that a modern compiler like GCC and LLVM is fairly certain to inline the code and optimize away the branch.

theo25 (commented):
I had a question and also a suggestion for you. Otherwise this looks good to me.

lib/binary/serializer.cpp
Baltoli: The reason is that we want to compute the diff _before_ the prefix (`0x01` or `0x02`) has been emitted to the buffer, so we're a byte further back in the buffer.

Baltoli: I've added a comment to this code explaining where the 5-byte offset comes from.

lib/binary/serializer.cpp
Baltoli: That makes sense - thanks!

lib/binary/serializer.cpp
theo25: That sounds correct, thanks for explaining!

theo25 (approved):
This looks good to me!

PR #485, opened 2022-02-17
Generalise argument parsing in interpreters

Baltoli:
Currently, compiled interpreters set the `@statistics` global option if `argc != 4` (i.e. there is a 4th positional argument supplied to the interpreter).

At some point we'll need to add a flag to interpreters that causes them to emit binary output rather than textual KORE; doing so will need a slightly more sophisticated argument parsing implementation than the above.

This PR does the initial generalisation work towards this, and should hopefully highlight any potential issues before we add additional flags.

The proposed new behaviour is to silently ignore any unknown flags; this will break anyone using the statistics option by passing a flag other than `--statistics` (which `krun` does when invoking the interpreter), so it's worth testing this PR carefully.

Adding new flags to the interpreter once this PR is merged is simply a case of adding new blocks to `parse_flags`.

### DIFF
--- a/runtime/main/main.ll
+++ b/runtime/main/main.ll
@@ -12,8 +12,47 @@ declare void @finish_rewriting(%block*, i1) #0
 
 declare void @initStaticObjects()
 
+@statistics.flag = private constant [13 x i8] c"--statistics\00"
+
 @output_file = external global i8*
 @statistics = external global i1
+
+declare i32 @strcmp(i8* %a, i8* %b)
+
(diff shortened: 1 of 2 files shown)


Baltoli:
I've built K locally with this change, and the test suite passes.

Baltoli:
This change should also apply to the `search` executables as appropriate.

theo25 (approved):
This looks good to me. Feel free to merge after you make sure this doesn't break any project that uses the LLVM backend. I guess you could let people know about the change on slack and if no objections are raised we can merge this.

PR #491, opened 2022-02-25
Don't `cat` huge files in `llvm-krun`

Baltoli:
This is a partial fix for the C semantics linking problems. Instead of storing file contents directly in a variable, we change the logic to read from the file on disk instead.

~~Testing this will probably need to be done in the frontend so that a definition is available for `llvm-krun`; I'll do so in a separate PR before this can be merged. Informally, the C semantics test suite (which makes heavy use of `llvm-krun`) passes when built with this change.~~

This can be reviewed now, but should only be merged once https://github.com/runtimeverification/k/pull/2459 has been merged; any breakages should be caught by the backend submodule update.

### DIFF
--- a/bin/llvm-krun
+++ b/bin/llvm-krun
@@ -172,7 +172,15 @@ for name in "${params[@]}"; do
     KItem)
     ;;
     *)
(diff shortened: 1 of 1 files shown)


theo25 (approved):
This looks good to me

PR #496, opened 2022-03-11
Generate binary KORE as interpreter output

Baltoli:
This is a follow-up PR to #484. The main changes it makes are as follows:

* Add an additional flag (`--binary-output`) to the command-line interface of compiled interpreters.
* Adapt the existing (textual) runtime configuration printer to generate binary KORE if that flag is passed.
* Ensure that the test suite produces binary KORE for each output, and converts it back to text for verification.

Once this is merged, I'll be able to adapt `llvm-krun` and the C semantics build scripts to produce binary KORE, entirely obviating the need for the textual format.


Baltoli:
There is one issue that needs to be resolved before this can be merged - see message [here](https://runtimeverification.slack.com/archives/CBQAELQMR/p1647002484227009) on slack.

Baltoli:
Depends on #497

Baltoli:
This is tested by the existing test suite:
https://github.com/runtimeverification/llvm-backend/blob/49aaca2a81d7a4a9076d9c0f4ddc3d82c4b4e7a3/test/lit.cfg.py#L44-L48

Each test definition runs the following variants:
* Text in, text out
* Convert text -> binary, run on binary input
* Run on text input, produce binary output, convert back to text

PR #497, opened 2022-03-14
Variable-size lengths

Baltoli:
This PR implements the variable-size lengths change as discussed last week.

All fixed-size length and arity fields in the binary KORE format have been replaced with a simple representation where the highest bit in a byte is a continuation flag, and the lower 7 bits are chunks of the length itself.

This representation is a bit more complex to encode and decode, but should save space compared to the previous version.

### DIFF
--- a/include/kllvm/binary/deserializer.h
+++ b/include/kllvm/binary/deserializer.h
@@ -44,12 +44,35 @@ T read(It &ptr, It end) {
 }
 
 template <typename It>
+uint64_t read_length(It &ptr, It end) {
+  uint64_t ret = 0;
+  auto should_continue = true;
+  auto steps = 0;
+
+  while (should_continue) {
+    assert(ptr != end && "Invalid variable-length field");
+
+    auto chunk = peek(ptr);
+    auto cont_bit = std::byte(0x80);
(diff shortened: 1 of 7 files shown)


include/kllvm/binary/deserializer.h
```diff
@@ -43,13 +43,36 @@ T read(It &ptr, It end) {
   return val;
 }
 
+template <typename It>
+uint64_t read_length(It &ptr, It end) {
+  uint64_t ret = 0;
+  auto should_continue = true;
+  auto steps = 0;
+
+  while (should_continue) {
```
dwightguth, line 52: we should probably explicitly error out after the 9th byte, otherwise we might potentially read the entire file as a length field on malformed input, which could lead to the deserialization process hanging.

dwightguth (changes requested):
in addition to the comment I had, it would be nice if the deserializer could still read version 1.0.0 terms, for compatibility purposes. Shouldn't be too hard to pass the version flag to read_length along with the number of bits it used to read.

theo25 (approved):
I don't have any further comments, so I will approve, but please address @dwightguth 's comments.

dwightguth (approved):

PR #498, opened 2022-03-14
Use binary KORE in tools

Baltoli:
~~Depends on #496; the diff in this PR won't look sensible until that one is merged.~~

## Summary

This PR adds support for the binary KORE format version `1.1.0` to `llvm-krun` and related tools. A list of the major changes is as follows:

* Fix a bug in `llvm-krun` where input textual KORE files would be modified if passed as a `korefile` argument.
* Add `binaryfile` mode to load configuration parameters directly from serialized binary files.
* Add `--binary-input` flag to construct the initial interpreter configuration as a binary file rather than textual KORE.
* Add `--binary-output` flag to produce binary KORE as the output of `llvm-krun`. This is useful when composing multiple execution steps in a pipeline, as the C semantics do.
* Implement binary KORE input support for `kprint` and `kore-expand-macros`.
* Add `kore-arity` tool to generate variable-length byte sequences at the command line; doing so is easier and safer than echoing / printf-ing raw bytes.
* Add `kore-strip` tool to remove header and term arity bytes from binary KORE files.
* Add `kore-header` tool to generate a KORE header + version without an associated serialized term.

The new paths in `llvm-krun` basically end up replacing textual interpolation of KORE terms (`ctor{}($var)`) with concatenation of binary files; these need their headers and arities maintained carefully to keep the resulting terms valid. The tools listed above provide convenient interfaces to do so without too much shell code.

## Testing

The bulk of this PR is related to the `llvm-krun` script, which is tested in the K frontend rather than as part of this repository. Because of this, I'll follow up this PR with a frontend PR that expands the existing test structure to cover the new binary features.

I've built and tested the `llvm-krun` changes in this PR against the K frontend, and they don't break the existing tests. The submodule update will catch any unforeseen breakages to existing code paths.


theo25 (approved):
This looks good to me. I just left a minor comment.

PR #501, opened 2022-03-16
Run shellcheck on backend scripts

Baltoli:
This PR introduces a new CI step to verify that the shell scripts in `bin/` conform to best practices using [`shellcheck`](shellcheck). If we're making changes to these scripts in the future, it's useful to know if we're likely to fall into any common pitfalls.

The changes made to the scripts are broadly small; in a few places I've replaced word-expanded strings with proper bash arrays.

[shellcheck]: https://www.shellcheck.net/


dwightguth (approved):
Looks good but I had one suggestion for cleanup. Go ahead and merge once you make that change

Issue #493, opened 2022-03-02
[Bug] [krun] - Backend crashed due to invalid string slice


yanliu18:
K Version
---------

```
K version:    v5.2.95-0-gaa5eee670-dirty
Build date:   Tue Mar 01 00:15:37 AEDT 2022
```

Description
-----------
Compilation terminated with error
```
libc++abi: terminating with uncaught exception of type std::invalid_argument: Invalid string slice
[Error] krun: Backend crashed during rewriting with exit code 134
```
Input Files
-----------
K definition file ``` string-split.k```
refer to exercise 1 of[ this tutorial](https://github.com/runtimeverification/k/tree/master/k-distribution/k-tutorial/1_basic/10_strings).

```
module STRING-SPLIT
    imports STRING
    imports INT

    syntax String ::= pick(String, String) [function]

    rule pick(R, "") => R
    rule pick(R:String, S:String) => pick(R +String substrString(S, 0, 1),
            substrString(S, findChar(S, ".", 0) +Int 1, lengthString(S)))
endmodule
```

Reproduction Steps
------------------

```
kompile string-split.k
krun -cPGM='pick("", "This.Is.A.Good.Day.")'
```

outputs:
```
libc++abi: terminating with uncaught exception of type std::invalid_argument: Invalid string slice
/Users/emma/Documents/tools/k/k-distribution/target/release/k/bin/../lib/kframework/k-util.sh: line 74: 14936 Abort trap: 6           "$@"
[Error] krun: Backend crashed during rewriting with exit code 134
```
Expected Behavior
-----------------

output "TIAGD"

yanliu18:
@Baltoli Thanks for the explanation. I realised that I've created non-determinism here (messed up with pattern matching in Haskell).

@ehildenb noted. Though, maybe there could be a better error message or warning.

PR #505, opened 2022-03-16
better error messages for substr hooks

theo25:
This PR improves the error messages for the `substrString` hook in order to address #493


theo25:
I wanted to have something that simulates the `std::format` of C++20. I guess I can make the code simpler, but I don't see how we can print detailed error messages without some kind of string formatting

theo25:
Would it help, if I add some unit tests that test the string formatting code?

runtime/strings/bytes.cpp
```diff
@@ -8,6 +9,18 @@
...
 #include "runtime/header.h"
 
+template <typename... Args>
+std::string string_format(const std::string &format, Args... args) {
+  int size_s = std::snprintf(nullptr, 0, format.c_str(), args...) + 1;
+  if (size_s <= 0) {
+    throw std::runtime_error("Error during formatting.");
+  }
+  size_t size = static_cast<size_t>(size_s);
+  std::unique_ptr<char[]> buf(new char[size]);
```
Baltoli, line 19: Very minor point but I tend to prefer `make_unique` here

Baltoli (approved):
One small point but otherwise looks good!

theo25:
I have changed the implementation to use `std::string` and string concatenation to generate the error messages in a less complicated manner.

runtime/strings/bytes.cpp
```diff
@@ -122,10 +122,22 @@ SortBytes hook_BYTES_substr(SortBytes input, SortInt start, SortInt end) {
   uint64_t ustart = get_ui(start);
   uint64_t uend = get_ui(end);
   if (uend < ustart) {
-    throw std::invalid_argument("Invalid string slice");
+    throw std::invalid_argument(
+        std::string("Invalid string slice for string \"")
+        + std::string(input->data) + std::string("\": Requested start index ")
```
Baltoli, line 127: I know we're trying to keep the code simple, but what if the string data is huge? Won't this produce a potentially enormous error message?

Baltoli (changes requested):

runtime/strings/bytes.cpp
theo25: That's fair, I removed the string from the error message. It should still be more informative than before.

Baltoli (approved):

PR #511, opened 2022-04-12
run kore-convert if needed in llvm-krun --dry-run

dwightguth:
If you run `llvm-krun --dry-run --binary-input` or `llvm-krun --dry-run --binary-output`, it will not respect the presence or absence of the `--binary-output` flag because it is copying the input file verbatim to the output. In this case, we want the script to run `kore-convert` on the input file before providing it back to the user. This PR implements this change.

### DIFF
--- a/bin/llvm-krun
+++ b/bin/llvm-krun
@@ -359,7 +359,7 @@ if $dryRun; then
     if [ -n "$verbose" ]; then
       set -x
     fi
-    "$(dirname "$0")/kore-convert" "$expanded_input_file"
+    "$(dirname "$0")/kore-convert" -F "$expanded_input_file"
     )
   elif [ -n "$real_output_file" ]; then
     cat "$expanded_input_file" > "$real_output_file"
--- a/bin/llvm-krun
+++ b/bin/llvm-krun
@@ -347,7 +347,21 @@ else
 fi
 
 if $dryRun; then
-  if [ -n "$real_output_file" ]; then
+  if [ $binary_input != $binary_output ] && [ -n "$real_output_file" ]; then
+    (
+    if [ -n "$verbose" ]; then
+      set -x
+    fi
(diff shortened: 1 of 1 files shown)


bin/llvm-krun
```diff
@@ -347,7 +347,21 @@ else
 fi
 
 if $dryRun; then
-  if [ -n "$real_output_file" ]; then
+  if [ $binary_input != $binary_output ] && [ -n "$real_output_file" ]; then
+    (
+    if [ -n "$verbose" ]; then
+      set -x
+    fi
+    "$(dirname "$0")/kore-convert" "$expanded_input_file" > "$real_output_file"
```
Baltoli, line 355:
```suggestion
    "$(dirname "$0")/kore-convert" "$expanded_input_file" -o "$real_output_file"
```

This should use `-o` rather than a redirect; `kore-convert` doesn't have a check for whether `stdout` is a TTY yet so binary output will break this. This is a limitation of the tool API, really

bin/llvm-krun
```diff
@@ -347,7 +347,21 @@ else
...
+    )
+  elif [ $binary_input != $binary_output ]; then
+    (
+    if [ -n "$verbose" ]; then
+      set -x
+    fi
+    "$(dirname "$0")/kore-convert" "$expanded_input_file"
```
Baltoli, line 362:
```suggestion
    "$(dirname "$0")/kore-convert" -F "$expanded_input_file"
```

If the user specifies binary output but no output file specifically, I presume we do just want to dump binary output to `stdout` here under the assumption that they know what they're doing and will redirect it?

Baltoli (changes requested):
There are a couple of minor changes owing to limitations of the `kore-convert` interface, rather than your changes!

Baltoli (approved):

PR #513, opened 2022-04-25
Add backend `#trace` hook

Baltoli:
This hook is a quality-of-life debugging feature that allows K terms to be printed to standard error in _surface syntax_, rather than as KORE.

Doing so requires that the bulk of `kprint`'s functionality be moved into a library that can be invoked at runtime. In #425, we added the kompiled directory path as a global symbol in the backed, which allows us to invoke `kprint` without a CLI.

The frontend documentation for this hook will mention that calling it is potentially expensive, and should only really be used as a debugging aid.

Draft PR that should be considered alongside the associated frontend support.


lib/printer/printer.cpp
```diff
@@ -0,0 +1,261 @@
...
+}
+
+namespace kllvm {
+
+using namespace parser;
+
+std::ostream &printKORE(
+    std::ostream &os, std::string const &definitionPath,
+    std::string const &patternPath, bool hasColor, bool filterSubst) {
+  std::map<std::string, std::string> formats;
```
dwightguth, line 30: If this code is going to be invoked in a running interpreter instead of a standalone executable, we need to make some modifications here. In particular, we should only build the information about the definition the first time we call this function, and then we should reuse it all in subsequent calls.

dwightguth: Please also do a leak check with valgrind once you have finished refactoring the method. I know for a fact this function as it's currently written leaks the memory that is part of the definition it parses, and I want to make sure we completely address this so that multiple calls to this function won't gradually accumulate memory leaks.

Baltoli:
I have refactored the printing logic substantially so that each call to `printKore` (i.e. each call to the new hook function) only recalculates information that depends on the _pattern_, rather than the definition.

I also ran a leak check as you suggested, using the following definition:
```k
module TEST
  imports INT
  imports K-IO

  syntax Foo ::= foo(Int)
  
  rule foo(0) => .
  rule foo(X) =>
    #let _::K = #trace(foo(X)) #in foo(X -Int 1) [owise]
endmodule
```

Running the program `foo(100000)` under Valgrind, I see:
```
==1660264== 
==1660264== HEAP SUMMARY:
==1660264==     in use at exit: 62,928,640 bytes in 100,001 blocks
==1660264==   total heap usage: 195,331,393 allocs, 195,231,392 frees, 15,676,848,618 bytes allocated
==1660264== 
==1660264== LEAK SUMMARY:
==1660264==    definitely lost: 0 bytes in 0 blocks
==1660264==    indirectly lost: 0 bytes in 0 blocks
==1660264==      possibly lost: 0 bytes in 0 blocks
==1660264==    still reachable: 62,928,640 bytes in 100,001 blocks
==1660264==         suppressed: 0 bytes in 0 blocks
==1660264== Rerun with --leak-check=full to see details of leaked memory
==1660264== 
==1660264== For lists of detected and suppressed errors, rerun with: -s
==1660264== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0)
```

lib/printer/printer.cpp
```diff
@@ -4,142 +4,244 @@
...
 
-static std::string trim(const std::string &s) {
+std::string trim(const std::string &s) {
   return rtrim(ltrim(s));
 }
 
-namespace kllvm {
+std::map<std::string, std::string> getFormats() {
+  static std::map<std::string, std::string> formats;
+  static bool once = false;
```
dwightguth, line 30: same comment here about once; we can flip the values and it becomes a bit easier to read. Not going to comment every time this appears in the PR after this though.

dwightguth (approved):

dwightguth:
Looks really good. I approved, but please try to address my minor comments before merging.

PR #515, opened 2022-05-04
build decision trees in parallel

dwightguth:


Baltoli (approved):
This passes CI, and the change itself seems logical so I'm approving.

radumereuta:
@dwightguth si there a measurable performance difference you were able to find?
I tested it with the c-semantics and can't find anything relevant.

PR #516, opened 2022-05-06
Make error messages friendlier

Baltoli:
This PR makes some very small changes to the error messages we print from the hooked map functions; the messages make it slightly easier for users to understand where a crash is coming from.

Fixes https://github.com/runtimeverification/k/issues/2565

### DIFF
--- a/runtime/collections/maps.cpp
+++ b/runtime/collections/maps.cpp
@@ -28,7 +28,7 @@ map hook_MAP_concat(SortMap m1, SortMap m2) {
   for (auto iter = from->begin(); iter != from->end(); ++iter) {
     auto entry = *iter;
     if (to.find(entry.first)) {
-      throw std::invalid_argument("Duplicate keys");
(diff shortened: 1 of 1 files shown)


runtime/collections/maps.cpp
```diff
@@ -45,7 +45,7 @@ SortKItem hook_MAP_lookup_null(SortMap m, SortKItem key) {
 SortKItem hook_MAP_lookup(SortMap m, SortKItem key) {
   auto res = hook_MAP_lookup_null(m, key);
   if (!res) {
-    throw std::invalid_argument("key not found");
+    throw std::invalid_argument("Key not found for map lookup");
```
radumereuta, line 48: Would it be possible to pretty print the key?
If you have a large definition with multiple places where something like this could happen, it would help narrow it down even faster.
I'm not sure if it would be feasible if the key is very large though. We might need to limit the pretty printed term size.

theo25 (approved):
This looks good to me.

PR #521, opened 2022-05-25
add formal documentation for decision tree algorithm

dwightguth:
I would really appreciate anyone interested in developing on the llvm backend who is not already familiar with this material to review this PR carefully with an eye towards its suitability to newcomers. The goal is that it should be able to fully explain how matching is compiled on all the cases we cover, without needing to refer to the original codebase.

### DIFF
--- a/matching/README.md
+++ b/matching/README.md
@@ -176,8 +176,6 @@ optionally a boolean condition Cond, we compute the initial matrix as follows:
 P starts with one column and n rows. P_i1 = LHS[i]. o_1 = []. s_1 = s.
 C_i = (a_i = i, S_i = empty, Cond_i = Cond[i], R_i = []).
 
-Note that while 
-
 The rows of P are sorted in increasing priority order, where each set of rows
 with the same priority forms a group.
 
--- a/matching/README.md
+++ b/matching/README.md
@@ -1,8 +1,8 @@
 # LLVM Backend Pattern Matching
 
 The LLVM backend pattern matching algorithm is heavily based on the paper
-"Compiling Pattern Matching to Good Decision Trees" by Luc Maranget. We will
-assume for the purposes of this documentation that the reader has read and
+"Compiling Pattern Matching to Good Decision Trees" by Luc Maranget [1]. We
+will assume for the purposes of this documentation that the reader has read and
 understands that paper in detail, and focus instead primarily on the
 differences between that paper and the final algorithm used by the LLVM
 backend. We will, however, exhaustively cover the preliminary concepts as they
@@ -58,9 +58,18 @@ syntax Pattern ::= Variable // variable patterns
                  | "Map" [(Pattern, Pattern)] Variable // map patterns with a variable
 ```
 
+Variables come in two varieties: anonymous variables, which do not get
+added to the substitution, and named variables, which do get added to the
+substitution. Named variables, when matched, are bound to a particular
+occurrence, which corresponds to a term at runtime. Anonymous variables match
+any pattern, like a named variable, but are not bound to anything.
+
 Note that one restriction not captured by the above grammar is that List, Set,
 and Map patterns with a Variable must have at least one element Pattern.
-Otherwise, they would simply be a variable pattern.
+Otherwise, they would simply be a variable pattern. The reason that `Set [] V`
+is equal to `V` is that it can appear only in a context that expects a set
+sort, which means that `List`, `Map`, or regular constructors cannot appear.
+The same, obviously, is true respectively for lists and maps.
 
 ### Actions
 
@@ -80,21 +89,37 @@ syntax OccurrenceItem ::= Int | "size" | "key" | "value"
                         | "rem" | "iter" | fresh(Int) | pat(Pattern)
 ```
 
+Integers are used to refer to indices within lists and regular terms. `size` is
+used to refer to the output of functions called to compute the size of a
+collection. `key`, `value`, and `rem` are used in the context of map/set
+matching in order to refer to the key and value of maps, the element of sets,
+and the remainder after an element that has been matched has been removed from
+the set. `iter` is used to refer to an iterator for a map or a set. `fresh` is
+used to refer to the output of a side condition on a rule. `pat` is used to
+identify *which* key or value a particular set or map occurrence is in relation
+to.
+
 ### Matrices
 
 A matrix represents an n-row by m-column grid of patterns, plus a *fringe* and
 a *clause list*. The fringe consists of m occurrences and m sorts, one for each
-column. The clause list consists of n clauses, one for each row. A clause C
+column. We call these occurrences for convenience o_1, ..., o_m and the sorts
+s_1, ..., s_m. What actual occurrence corresponds to o_1 depends on what
+occurrence is in the first column of the matrix and will change as the matrix
+transforms. The clause list consists of n clauses, one for each row. A clause C
 consists of an action a, a substitution S (mapping variables to occurrences and
 sorts), a boolean expression Cond (representing the side condition), and a list
-R of type [(Occurrence, Sort, Int, Int)], used to compile list patterns.
+R of type [(Occurrence, Sort, Int, Int)], used to compile list patterns. We
+call R the set of `list ranges` bound by a particular row. This corresponds
+to each variable `L` in a pattern of the form `List ps; L; qs`. We use this
+information to ensure that the correct substitution for L is bound.
 Conceptually, rows of a matrix are sorted into groups, which represent a set of
 rows of equal priority. The notion of priority generalizes the concept from
 Maranget's paper that rows at the top of the matrix apply first if they can,
 and lower rows only apply if the upper rows cannot. Instead, any row in the
 topmost group can apply if it matches, and rows in lower groups only apply if
 no row in an upper group applies. This allows us to improve pattern matching
-efficiency since K does not, for the most part, case what order rules are tried
+efficiency since K does not, for the most part, care what order rules are tried
 in.
 
 Matrices are first simplified to remove or-patterns by replacing each row
@@ -108,8 +133,8 @@ process and whose leaves identify which rewrite rule, if any, applies to the
 term being matched. It has the following BNF-style syntax:
 
 ```
-syntax DT ::= "Failure"
-            | Success(Int, [(Occurrence, Sort)])
+syntax DT ::= "Failure" // Fail in Maranget's paper
+            | Success(Int, [(Occurrence, Sort)]) // Leaf in Maranget's paper
             | Switch(Occurrence, Sort, [(Constructor, DT)], DT?)
             | CheckNull(Occurrence, Sort, DT, DT)
             | Function(Occurrence, Function, [(Occurrence, Sort)], Sort, DT)
@@ -117,11 +142,25 @@ syntax DT ::= "Failure"
             | IterHasNext(Occurrence, Sort, DT, DT)
 ```
 
+We will describe how each decision tree node is used later, but essentially,
+Failure corresponds to the leaf case where no rule is matched, Success
+corresponds to the leaf case where one particular rule is matched, Switch
+corresponds to testing the constructor of a particular pattern, CheckNull
+corresponds to testing whether a particular value is None, Function corresponds
+to calling a particular function and binding the return value into the
+substitution, `Pattern` corresponds to constructing a particular pattern and
+binding it to an occurrence, and `IterHasNext` is very similar to `CheckNull`
+except that it will be used to implement the behavior that backtracks to the
+next element in a map or set in the case where we are iterating through a
+collection testing whether each element will make matching succeed.
+
 ### Heuristics
 
 The only heuristics regularly used by the LLVM backend are `q`, `b`, `a`, and
 the `L` pseudo-heuristic. We will discuss in a later section how these
-heuristics are adapted to the new features of the algorithm.
+heuristics are adapted to the new features of the algorithm. We also discuss
+how we extend the notion of heuristics to deal with the problem of choosing
+in what order to decompose the keys of maps and sets.
 
 ## Compilation
 
@@ -137,6 +176,8 @@ optionally a boolean condition Cond, we compute the initial matrix as follows:
 P starts with one column and n rows. P_i1 = LHS[i]. o_1 = []. s_1 = s.
 C_i = (a_i = i, S_i = empty, Cond_i = Cond[i], R_i = []).
 
+Note that while 
+
 The rows of P are sorted in increasing priority order, where each set of rows
 with the same priority forms a group.
 
@@ -144,7 +185,12 @@ with the same priority forms a group.
 
 `Compile` is a function that takes a Matrix and returns a Decision Tree.
 
-Described below is pseudocode for the function:
+Described below is pseudocode for the function. This essentially corresponds
+exactly to the `CC` function in Maranget's paper, except that rows are grouped
+into groups with equal priority rather than totally ordered, and we explicitly
+handle variables and side conditions. the `decomposeBestCol` function
+essentially corresponds exactly to Maranget's algorithm in the case where the
+sort of the chosen column is a regular sort.
 
 ```
 def compile(Matrix m): DT
@@ -173,10 +219,10 @@ def getLeaf(Row i, DT dt): DT
   val atomicLeaf = Success(a_i, vars)
   if Cond_i is not None:
     val f = a function that evaluates Cond_i
-    val sc = Function(Fresh(a_i), f, vars, Bool, dt)
+    val sc = Function(fresh(a_i), f, vars, Bool, atomicLeaf)
   else:
     val sc = atomicLeaf
-  return foldRanges(R_i, dt)
+  return foldRanges(R_i, sc)
 
 def decomposeBestCol(Matrix m): DT
   choose a column j that is best according to the heuristics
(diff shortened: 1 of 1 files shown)


matching/README.md
```diff
@@ -0,0 +1,991 @@
...
+    return None
+
+def getLeaf(Row i, DT dt): DT
+  val newVars = empty
+  for j = 1 to m:
+    for each variable X in P_ij:
+      newVars[X] = o_j
+  val allVars = newVars U S_i
+  val vars = sort the values of allVars by their keys alphabetically
+  val atomicLeaf = Success(a_i, vars)
```
Baltoli, line 173: How are the fringe entries `o_j` initialised? On line 137 you set up the first entry `o_1 = []`, but don't mention further cases. What makes `o_1` different such that it's initialised explicitly and each further `o_i` is implicitly pulled in from the term being matched?

Here, if `P_i = [x(A, B) y(C) D]` then we'll end up with:
```
newVars[A] = o_1
newVars[B] = o_1
newVars[C] = o_2
newVars[D] = o_3
```
I don't follow why we'd want `A` and `B` to map to the same occurrence. Wouldn't we actually want to know that `A` maps to the occurrence `o_1 . 1` (i.e. the first child of `o_1`):
```
newVars[A] = o_1 . 1
newVars[B] = o_1 . 2
newVars[C] = o_2 . 1
newVars[D] = o_3
```

matching/README.md
```diff
@@ -0,0 +1,991 @@
...
+  if Cond_i is not None:
+    val f = a function that evaluates Cond_i
+    val sc = Function(Fresh(a_i), f, vars, Bool, dt)
+  else:
+    val sc = atomicLeaf
+  return foldRanges(R_i, dt)
```
Baltoli, line 179: I don't understand the intended semantics of the side condition parts of the decision tree structure. The variable `sc` isn't used at all here, and the `atomicLeaf` isn't used in the part where we construct the `Function` term.

matching/README.md
```diff
@@ -0,0 +1,991 @@
...
+whose child pattern is a wildcard pattern.
+
+`foldRanges`, `decomposeList`, `decomposeMap`, and `decomposeSet` are described
+as part of the sections on list, map, and set matching below.
+
+### Signature function
+
+`Signature` is a function that takes a column j of a matrix and a sort and
+returns the set of constructors that must be considered when pattern matching
+examines the term correpsonding to o_j.
```
Baltoli, line 212:
```suggestion
examines the term corresponding to o_j.
```

matching/README.md
```diff
@@ -0,0 +1,991 @@
...
+* specialize(c, i, X) = 
+  [P_i1, ..., P_i(j-1), Y_1, ..., Y_a, P_i(j+1), ..., P_im] -> C
+  where Y_1, ..., Y_a are fresh anonymous variables that do not appear in the
+  substitution S, and C is the clause of row i except S_i is updated with the
+  binding X |-> (o_j, s_j) if X is not an anonymous variable.
+* specialize(c, i, c(p_1, ..., p_a)) =
+  [P_i1, ..., P_i(j-1), p_1, ..., p_a, P_i(j+1), ..., P_im] -> C_i
+* specialize(c, i, c'(p_1, ..., p_a')) where c != c' = None
+* specialize(c, i, p as X) = specialize(c, i, p) except S_i is updated with the
+  binding X |-> (o_j, s_j)
```
Baltoli, line 267: Does this differ from Maranget's specialisation in anything other than the tracking of variable bindings?

matching/README.md
```diff
@@ -0,0 +1,991 @@
...
+P_ij.
+
+We define default first on rows.
+
+* default(i, X) = [P_i1, ..., P_i(j-1), P_i(j+1), ..., P_im] -> C
+  where C is the clause of row i except S_i is updated with the binding
+  X |-> (o_j, s_j)
+* default(i, c(p_1, ..., p_a)) = None
+* default(i, p as X) = default(i, p) except S_i is updated with the binding
+  X |-> (o_j, s_j)
```
Baltoli, line 312: Same question as above wrt. bindings / Maranget.

matching/README.md
```diff
@@ -0,0 +1,991 @@
...
+an exception (on failure) or calls a function which returns the right-hand side
+of rule A substituted with S. However, we define it this way formally in order
+to allow implementations to choose the details of exactly what machine code
+to generate.
+
+The function `evaluate` is defined statefully using an intermediate state
+representing the substitution S which initially maps [] to the initial pattern,
+and a stack of (DT, S) pairs that is initially empty. In order to
+implement set/map matching, S is extended to allow us to map occurrences to an
+iterator over a map or a set, or to None.
```
Baltoli, line 362: What does it mean for us to map an occurrence to an iterator? I can visualise a simple substitution but it's less clear to me what the semantics of this are. Why don't lists participate in iterator mapping - is it to do with the ACU property?

matching/README.md
```diff
@@ -0,0 +1,991 @@
...
+
+Roughly speaking, `decomposeList` behaves similarly to the default Switch case
+except it switches over the runtime length of the list and makes sure to add
+bindings for each list element being matched on to the substitution. Note that
+in the case of the default, we know that the list must be greater in length
+than any list pattern in the column being decomposed, so we only must decompose
+maxHead elements from the front of the list and maxTail elements from the end,
+since we can guarantee any elements in between will never be referenced by any
+submatrix. This is the reason why we treat the default operation as equivalent
+to specializing on `maxHead + maxTail` elements.
```
Baltoli, line 510: This kind of explanation is nice to have up-front, even if in an abbreviated form.

matching/README.md
```diff
@@ -0,0 +1,991 @@
...
+
+As a result, K has two types of matching on map patterns. The first, when the
+key of the map element being matched on is bound, is called a *lookup*. In the
+event that the pattern is not bound when matching occurs, we call this a
+*map choice*. Essentially, the intuition behind a map choice is that we try
+each element of the map in turn, proceeding as if it is the correct one, until
+matching fails, at which point we backtrack to the next map element. In the
+worst case, if matching a rule requires n map choices on a map of size m, then
+the complexity of this algorithm is `O(m^n)`. This is, of course, worse than
+the complexity of linear AC matching, which is `O(s*t^3)` where s is the size
```
Baltoli, line 575: Do we have nice citations for AC matching? If we're claiming this is "of course" we should cite the AC complexity

matching/README.md
```diff
@@ -0,0 +1,991 @@
...
+in p have a binding in S_i.
+
+### Map Patterns
+
+K does not implement full AC matching on maps or sets. Instead, it attempts to
+order the matching process in such a manner that wherever possible, the
+substitution for all variables in the keys of map patterns are bound before
+matching on the map itself. This allows us to, in the vast majority of cases,
+compile matching on maps to map membership and lookup checks, which are vastly
+more efficient than AC matching.
```
Baltoli, line 565: For a technical document this is fine here, but for a paper we'll probably want to shift it closer to the start to understand the motivation for extending Maranget with Lists / Sets / Maps.

Baltoli: By the way - do you have an intuitive example of what happens if you _don't_ specialise your implementation for collections, and instead just implement them as algebraic types (i.e. use their "as-if" version)? You'd end up with horrible-to-write code, and wouldn't be able to abstract over collections of different size, but what would the difference in the generated trees look like?

matching/README.md
```diff
@@ -0,0 +1,991 @@
...
+performs choices on a single varaible, which can be performed in `O(n)` time.
+Cases where multiple choices are required usually indicate that the semantics
+is badly optimized.
+
+Note that, technically speaking, the algorithm does not care whether it
+performs lookups or choices, and it does not care which keys are decomposed in
+which order. Thus, we first describe the semantics of matching
+nondeterministically with the assumption that *some* key will be chosen, and
+describe the process of deciding /which/ key to choose in the section on
+heuristics.
```
Baltoli, line 590: This is a clear extension over Maranget as well - mention closer to the start that heuristics are extended to cover collection patterns.

matching/README.md
```diff
@@ -0,0 +1,991 @@
...
+  if row i is not in the topmost group of m
+* specialize(Choice, i, Map []) = None
+
+Note that in the above, if ps |-> qs = [], then Map ps |-> qs M is just a variable pattern M.
+
+Intuitively, specializing on Empty creates no new columns. Specializing on
+Key(k) creates 3 new columns corresponding to the value bound to key k, the
+remaining map pattern after removing key k, and the original map itself. The
+third category is required because we may test for membership of a key that
+is only present in some rows of the matrix, and thus, for rows that may or maya
```
Baltoli, line 641:
```suggestion
is only present in some rows of the matrix, and thus, for rows that may or may not
```

matching/README.md
```diff
@@ -0,0 +1,991 @@
...
+we need to redefine the q, b, a, and L heuristics used by K in light of the
+differences between the Maranget paper and our own design, as well as defining
+the `bestKey` function.
+
+First, we start from the assumption we want to avoid set and map choices
+whenever possible, as these are very expensive. Thus, we define the concept of
+a *valid* or *invalid* column. A column over list or regular sorts is always
+valid. A column over map or set sorts is valid if all of its map patterns are
+of the form Map [], or if there exists a key k in a map pattern in the column
+such that k is bound.
```
Baltoli, line 797: Valid here captures the idea of columns that don't require us to perform an expensive choice operation; there might be a better way to name it but I can't think of one off the top of my head.

matching/README.md
```diff
@@ -0,0 +1,991 @@
...
+
+```
+def specializeBy(Matrix m, [Term] rhs): (Matrix, [Term])
+  if m has no rows:
+    return (m, rhs)
+  if every term t in rhs is a wildcard pattern or is not a pattern:
+    return (m, rhs)
+  pick the best column j of m such that rhs[j] is a non-wildcard pattern
+  val c = the constructor of rhs[j]
+  return specializeBy(specialize(m, j, c), specialize(rhs, j, c))
```
Baltoli, line 974: So the iteration process will keep going until it reaches an empty matrix (terminal pattern?) or the RHS is completely wildcard patterns.

Did we do any experiments on this process? Is it ever worth applying the optimisation partially to trade off between code size and runtime performance?

Baltoli (commented):
I've left a number of comments - largely these follow my thought process following the document in comparison to Maranget, as well as some thoughts on how this technical detail could be transferred into a paper version that highlights the important contributions and motivation.

matching/README.md
```diff
@@ -0,0 +1,991 @@
...
+            | Function(Occurrence, Function, [(Occurrence, Sort)], Sort, DT)
+            | Pattern(Occurrence, Sort, Pattern, DT)
+            | IterHasNext(Occurrence, Sort, DT, DT)
+```
+
+### Heuristics
+
+The only heuristics regularly used by the LLVM backend are `q`, `b`, `a`, and
+the `L` pseudo-heuristic. We will discuss in a later section how these
+heuristics are adapted to the new features of the algorithm.
```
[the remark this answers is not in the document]
dwightguth: Yes, we tried a bunch of different heuristics and these proved to be a good compromise between decision tree complexity and compilation time.

matching/README.md
```diff
@@ -0,0 +1,991 @@
...
+      None -> decomposeBestCol(m)
+      BestRow(i) -> getLeaf(i, compile(m with row i removed))
+
+def bestRow(Matrix m): Int
+  if m has no columns:
+    return the first row of m
+  if there is a row in the topmost group of m where all columns are wildcard patterns:
+    return that row
+  else:
+    return None
```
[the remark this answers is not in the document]
dwightguth: Similar, but a little different. In particular we introduce the notion of groups of rows here.

matching/README.md
```diff
@@ -0,0 +1,991 @@
...
+    val f = a function that evaluates Cond_i
+    val sc = Function(Fresh(a_i), f, vars, Bool, dt)
+  else:
+    val sc = atomicLeaf
+  return foldRanges(R_i, dt)
+
+def decomposeBestCol(Matrix m): DT
+  choose a column j that is best according to the heuristics
+  if column j has a sort that is a regular sort:
+    return Switch(o_j, s_j, compiledCases(m, j), compiledDefault(m, j))
```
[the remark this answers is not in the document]
dwightguth: yes, essentially.

matching/README.md
dwightguth: I added the following text to the section `Matrices` in the preliminaries and I hope it makes the notation clearer:

```The fringe consists of m occurrences and m sorts, one for each
column. We call these occurrences for convenience o_1, ..., o_m and the sorts
s_1, ..., s_m. What actual occurrence corresponds to o_1 depends on what
occurrence is in the first column of the matrix and will change as the matrix
transforms```

File: matching/README.md (hunk 2 above)
Comment (dwightguth) [reply matching/README.md:179]: whoops, you are correct. It's a typo.

File: matching/README.md (hunk 4 above)
Comment (dwightguth) [reply matching/README.md:267]: No.

File: matching/README.md (hunk 5 above)
Comment (dwightguth) [reply matching/README.md:312]: Again, no.

File: matching/README.md (hunk 6 above)
Comment (dwightguth) [reply matching/README.md:362]: We don't need to use iterators to implement list matching. Essentially, you can think of occurrences as being the variables that exist at runtime. Mapping an occurrence to an iterator just means that instead of storing a term, we store an iterator.

File: matching/README.md [hunk 17]
```diff
@@ -0,0 +1,991 @@
...
+    for each key k in validKeys:
+      compute the score of column j according to key k
+    return the key with the highest score
+```
+
+Now, we have to redefine each heuristic q, b, a, and L. To summarize, heuristic
+q measures how many rows at the top of a column are not wildcard patterns, b
+measures the number of constructors in the signature of the column, and a
+measures the number of columns specializing on the column would create.
+Finally, L simply chooses the leftmost column with the shortest occurrence.
```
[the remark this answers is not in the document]
dwightguth: some of both. We evaluated some K definitions against each of the heuristics that Maranget mentioned as being high-quality, and found that this one worked the best among those which did not significantly increase compilation time.

matching/README.md
```diff
@@ -0,0 +1,991 @@
...
+occurs can actually be very neatly understood using the concept of priority
+inversion. We can think of the columns as tasks with priorities equal to their
+scores, and dependencies such that each invalid column depends on those columns
+which contain variables present in the keys of the first column in the same row
+(ie, decomposing the second column and the columns created by doing so repeatedly
+would eventually add the bindings to S which make the first column valid). It
+then becomes natural to think of the case when a higher priority column depends
+on a lower priority column as a priority inversion. The natural solution to
+priority inversion problems becomes to assign the effective priority of the low
+priority column equal to the highest priority of all the columns that depend on
```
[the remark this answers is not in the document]
dwightguth: I went ahead and wrote pseudocode to describe how to compute the effective priority of a column.

matching/README.md
dwightguth: I mean, you wouldn't really get much of a space savings by only doing it partially, because you would still have the problem that you have a bunch of different decision trees, one for each rule, that need to be turned into functions.

But we do have some heuristics that say, don't bother generating a decision tree for a particular rule if it isn't much better than the original decision tree. I will add some text here to explain how we do that.

matching/README.md
Baltoli: That makes sense - thanks

matching/README.md
```diff
@@ -58,9 +58,18 @@ syntax Pattern ::= Variable // variable patterns
...
+occurrence, which corresponds to a term at runtime. Anonymous variables match
+any pattern, like a named variable, but are not bound to anything.
+
 Note that one restriction not captured by the above grammar is that List, Set,
 and Map patterns with a Variable must have at least one element Pattern.
-Otherwise, they would simply be a variable pattern.
+Otherwise, they would simply be a variable pattern. The reason that `Set [] V`
+is equal to `V` is that it can appear only in a context that expects a set
+sort, which means that `List`, `Map`, or regular constructors cannot appear.
+The same, obviously, is true respectively for lists and maps.
```
Baltoli, line 72: This makes sense; I hadn't appreciated the sortedness constraint on where the constructors appear.

matching/README.md
```diff
@@ -137,14 +176,21 @@ optionally a boolean condition Cond, we compute the initial matrix as follows:
 P starts with one column and n rows. P_i1 = LHS[i]. o_1 = []. s_1 = s.
 C_i = (a_i = i, S_i = empty, Cond_i = Cond[i], R_i = []).
 
+Note that while 
+
```
Baltoli, line 180: Trailing sentence?

matching/README.md
```diff
@@ -989,3 +1067,33 @@ Key(k). Care would have to be taken that if the signature of a column in m was
...
+a number between 0 and 1. We then compare the number of rows in the specialized
+matrix to the number of rows in the original matrix, as a fraction. If the
+fraction is larger than the threshold, we say that the matrix does not
+specialize well for this rule, and we simply discard that matrix and the
+resulting decision tree and fall back on constructing the original RHS of the
+rule and using the original decision tree at runtime. We have found that a
+threshold of 1/2 works quite well for us, but this can be increased in order
+to improve the execution speed somewhat, or decreased to decrease the code
+size somewhat. Of course, in either direction you will eventually see
+diminishing returns.
```
Baltoli, line 1089: This is interesting - thanks!

Baltoli (approved):
One very minor change, but this goes a long way to clearing up the parts I didn't understand or had questions on.

matching/README.md
dwightguth: I think this was a fragment I started writing relating to occurrences and what o_1 meant, but then I ended up adding the relevant text to the section on matrices instead.

PR #522, opened 2022-05-27
Nix flake improvements

goodlyrottenapple:
This is part of the nix rengineering effort https://github.com/runtimeverification/k/issues/2607
As we are introducing nix flakes in other parts of the k framework, there are some tweaks that need to be made to the current flake expression. Namely:

* the use of `ttuegel.cleanGitSubtree` is being deprecated in the flake, as nix complains about not having an SHA256 for the fetch command thes functions use. This is done by pulling the clean functions out of the overlay and calling the `ttuegel` functions in `default.nix` to preserve the current behaviour and using the `lib.cleanSource` together with `nix-gitignore.gitignoreSourcePure` for the flake expression. 
* the llvm tools used are being bumped to LLVM 12
* add both dev and release packages, built via `nix build .#llvm-backend` and `nix build .#llvm-backend-release` respectively
* set `llvm-backend` as default package (built when running `nix build .`)
* use serokell's nixfmt on all the nix files

Missing:
* CI tests as a nix expression?
* `llvm-backend-release` does not build on the M1, but is not critical


calin1304:
LGTM

Baltoli (approved):

PR #526, opened 2022-06-07
Strip injections from terms to be `#trace`d

Baltoli:
While fixing up the frontend support for this hook, I realised that terms like:
```
inj{SortInt{}, (null)}(\dv{SortInt{}}("1234"))
```
were ending up in the KORE handed to the pretty-printer. To handle this correctly, we need to strip the injection from the term before printing it.

### DIFF
--- a/runtime/io/logTerm.cpp
+++ b/runtime/io/logTerm.cpp
@@ -13,6 +13,7 @@ extern "C" {
 extern char kompiled_directory;
 
 block *hook_IO_log(string *path, string *msg);
+int getTag(block *term);
 
 SortKItem hook_IO_logTerm(SortString path, SortKItem term) {
   string *msg = printConfigurationToString(term);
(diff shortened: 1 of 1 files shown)


dwightguth:
It may be better to simply provide the sort to the function that does the unparsing, then you won't have a `null` and the pretty printer will be able to handle it. The sort ought to be known statically since you know the signature of the hook.

Baltoli:
Ah, that's a better idea. I did think this seemed like a bit of a hack

Baltoli:
@dwightguth this is ready for another look

dwightguth (commented):
Can we get a test for this?

Baltoli:
There's an integration test in the frontend that tests the built in sort printing - I can add a unit test for this as well if you like

dwightguth (approved):

PR #530, opened 2022-06-09
better error message when tag for symbol is not found

theo25:
This PR addresses the issue https://github.com/runtimeverification/k/issues/2618. We use the tag `2^32-2` as a special value signifying that no tag was found. Then, during runtime we check that tag lookups do not return the error value. This way, we can have a more informative assertion failure message.


runtime/util/ConfigurationParser.cpp
```diff
@@ -33,6 +33,10 @@ uint32_t getTagForSymbolName(const char *name) {
     return lb->second;
   }
   uint32_t const tag = getTagForSymbolNameInternal(s.c_str());
+  assert(
```
dwightguth, line 36: let's not use assert for this, as it will not trigger if NDEBUG is set, as it generally is when we are building release builds of semantics.

Baltoli: Agreed - it would be nice to print the symbol name as well, given that it's available.

theo25: Agreed. I just pushed a change that prints the error message to `stderr` and aborts.

runtime/util/ConfigurationParser.cpp
```diff
@@ -33,6 +33,11 @@ uint32_t getTagForSymbolName(const char *name) {
     return lb->second;
   }
   uint32_t const tag = getTagForSymbolNameInternal(s.c_str());
+  if (tag == ERROR_TAG) {
+    std::cerr << "No tag found for symbol " << name << ". Maybe attempted to "
+              << "evaluate a symbol with no rules?\n";
+    abort();
+  }
```
Baltoli, line 40: How many places is this function called from? It might be better to delegate the reporting to _callers_ of this function; that way we can be more precise in the error message we print.

If there are lots of call sites and it would be a big change to do that, then I'm happy with this.

theo25: It is called in two places in the configuration parser, in the functions that construct the initial configuration from text and binary KORE. It is also called quite some times in various runtime libraries (json, bytes, io, arithmetic, ffi). In all of these callsites, the symbol that is passed as an argument is a hardcoded symbol like `dotk{}`, some version of `inj{{.., ..}`, etc.. As such, I don't see much utility from specialized error messages.

Baltoli: Great, this works absolutely fine for me in that case.

Baltoli (approved):

theo25:
@dwightguth do you have any further comments? Or can I merge this?

dwightguth (approved):

Issue #531, opened 2022-06-13
Fix binary KORE printing of `amb`


Baltoli:
This got missed out of #529, but is the exact same problem as that PR.

PR #533, opened 2022-06-16
Binary KORE serialization of amb nodes

gtrepta:
fixes: #531


Baltoli (approved):
Looks good to me given that there's a separate issue to fix this more generally in the future. Thanks @gtrepta

PR #538, opened 2022-07-05
Enable CI for macOS / arm64

Baltoli:
This PR makes several changes to the build and test configurations for the LLVM backend in order to make our CI processes compatible with M1 macOS machines; this is an important first step towards full K support on that platform.
* Build the LLVM backend Nix flake as well as the standard derivation. This makes sure that we're testing the flake build _before_ the upstream K repo uses it.
* Sets up Nix installation properly for our self-hosted M1 machines; previously, we were blocked on figuring out how to install Nix on each run (thanks @F-WRunTime)
* Add an option to completely disable LTO in the backend build, even in Release configurations. This isn't ideal, but there is a [6 year old issue](https://github.com/NixOS/nixpkgs/issues/19098) in Nix that prevents us from using Clang LTO on ARM macOS. Until that's fixed, I don't have a better solution to getting Nix + M1 running here. If performance is a critical issue for M1 users, they can simply build outwith Nix.


ehildenb:
Until our M1 is set up better, we probably can't make progress here.

F-WRunTime:
Apologies Idk how this slipped through the cracks. I will look into this.

F-WRunTime:
Okay @Baltoli Made some progress on this so it isn't continually blocked going forward. I still have some work to do to get the ephemeral runners. But I did get a successful build. 
I have one more commit I'm pushing after this. Which ideally will get the ubuntu stuff tagged in here as well. 
Please let me know your thoughts.

F-WRunTime:
This was a big piece and took 1hr 15m to finish. 
https://github.com/runtimeverification/llvm-backend/runs/8079650499?check_suite_focus=true

F-WRunTime (approved):

F-WRunTime:
> 

Okay thanks, taking a look

PR #545, opened 2022-08-10
Fix GDB on Ubuntu Jammy

Baltoli:
Fixes https://github.com/runtimeverification/k/issues/2637.

The actual change here is pretty small - it adds some extra global flags to the debug metadata we emit to match the behaviour of clang when emitting DWARF v5 files.


dwightguth (approved):
It looks good, but did you test that this doesn't break the dwarf v4 code path?

Baltoli:
Yes, the dwarf 4 codepath is fine with the new change. The LLVM metadata is identical for `clang-12` (dwarf 4) and `clang-14` (dwarf 5) C codegen, it's just the dwarf generation that changes with `clang++-14`.

PR #546, opened 2022-08-10
Structure code for future LLDB support

Baltoli:
This PR sets up the infrastructure for future K support of LLDB (https://github.com/runtimeverification/k/issues/2650), but doesn't yet implement the K-specific debugging features. Future PRs can follow on implementing individual debugger support features by hooking into the `dispatch_k` function.

The stubbed LLDB plugin as implemented here just attaches to the interpreter, sets a breakpoint before `main`, then drops into the default LLDB REPL. Any K commands (`start`, `step` etc.) just produce a warning.

I'll follow this up with a corresponding frontend PR (similar to https://github.com/runtimeverification/k/pull/2646) that launches `klldb.py` with the appropriate arguments when `--debugger` is passed to `krun` on macOS, and updates the documentation accordingly.


dwightguth (approved):
This looks fine as far as it goes, so I'm approving, but if we name the lldb script klldb.py, should we rename the gdb script kgdb.py?

Baltoli:
Good point - I'll do that then merge if you're happy.

dwightguth:
Yes, that's fine. That's why I approved.

PR #547, opened 2022-08-12
[WIP] pybind11 investigation

Baltoli:
This PR is tracking my work on investigating [`pybind11`](https://github.com/pybind/pybind11) as a way of packaging the LLVM backend as a Python library.


Baltoli:
Now fully subsumed by the production versions of these changes.

PR #549, opened 2022-08-25
Set up basic Pybind integration

Baltoli:
This PR is a first effort to build a production-ready version of the prototype pybind11 integration in #547; that PR will be superseded when this one is merged.

The code in this module is intentionally incomplete; it is the minimum amount of code required to build a python module exposing some functionality from the LLVM backend to python. Several more PRs will be required to fully implement the backend support we have been discussing.

Nevertheless, this PR is worth reviewing and merging on its own currently. It:
* Sets up a pybind submodule for the LLVM backend
* Builds the two configurations of library that we need for backend integration (shared + static for llvm-kompile)
* Ensures that the new binding code builds under CI
* Exposes a feature to Python that could be integrated into Pyk (KORE sorts)

Further, once the submodule update corresponding to this PR goes through, we can make a corresponding frontend / pyk PR that consumes a small feature from the bindings.

Further changes will be able to build on the code here with smaller, more self-contained PRs once the infrastructure is ready.


ehildenb:
Nice and small :) Excited for this! Will leave it to C++ experts to decide if this is good code.

tothtamas28 (approved):
Tried this from the Python interpreter, works nicely.

tothtamas28:
> does the structure of the binding code make it clear to you what's going to get produced on the python side?

Yes, I find it quite readable.

CMakeLists.txt
```diff
@@ -4,6 +4,10 @@ list(APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake")
 include(LLVMKompilePrelude)
 project (KLLVM CXX C)
 
+# Required to build the Python bindings as a static library that is then linked
+# into a shared library with the interpreter library generated by llvm-kompile.
+set(CMAKE_POSITION_INDEPENDENT_CODE ON)
```
dwightguth, line 9: Do we need to do this globally or can we only do it for the targets that need it?

theo25:
This looks to me! I agree with Dwight's question though, let's make sure we get this right.

Baltoli:
@dwightguth @theo25 The `-fPIC` flag needs to be enabled for every library that the Python module eventually links against (as it's a shared library). The easiest way to achieve that is to enable it globally; otherwise you end up with a lot of repeated CMake code to set it for each library in the transitive closure of the module's dependencies.

If performance is an issue, what we'd need to do is build two copies of each backend library - one destined for static linking, and the other for dynamic. This would need some bigger boilerplate changes to our CMake code that I'd probably want to do in a separate PR.

Baltoli:
I ran a very quick benchmark on the C semantics with `-fPIC` enabled globally vs. the existing backend; the following code was executed 10 times and the wall clock execution times measured with `/usr/bin/time`:
```c
int main() {
  for (int i = 0; i < 1000000; ++i) {
  }

  return 0;
}
```

| Version | Time | Std. Dev. | Slowdown |
| :-- | --: | --: | --: | 
| Unmodified | 30.91 | 0.51 | - |
| `-fPIC` enabled | 31.18 | 0.57 | 1.087x |

I'd be happy to accept a sub-1% slowdown in the backend for the substantially smaller CMake code that the global flag enables, but if others disagree I can implement the two-library solution in a separate PR.

dwightguth:
I'm not hugely worried about performance. My main concern is that I know sometimes position independent code can interact with other compiler features in ways that might cause weird breakages. The tests for the backend are passing though, so I'm willing to give this a try and as long as nothing breaks, it should be fine. If anything downstream does break though, we may have to visit the possibility of having separate build targets to rebuild the dependencies you need with fPIC.

dwightguth (approved):

PR #550, opened 2022-08-30
The stages for building and testing different OSes are serialized

theo25:
This PR ensures that no more than one executors will be assigned to single CI job. It does that by making sure that the stages that build and test on different OSes (which require different agents) are not run in parallel but rather sequentially hence being able to reuse the same executor.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -10,57 +10,55 @@ pipeline {
     }
     stage('Build and test') {
       options { timeout(time: 30, unit: 'MINUTES') }
-      parallel {
-        stage('Arch Linux') {
-          agent {
-            dockerfile {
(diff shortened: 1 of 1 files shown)


dwightguth (dismissed):
This looks fine as far as it goes, but I wonder if we should next try to parallelize each of the configuration targets within an OS in a follow-up PR.

dwightguth:
Looks like something went wrong parsing the updated Jenkins file. I'm going to dismiss my review and I'll take another look when you get it passing. Shouldn't be too complex of a fix though I should think. Not entirely sure what is wrong, but hopefully you can figure it out.

dwightguth (approved):

PR #551, opened 2022-08-31
Set up build script for Python

Baltoli:
In order to get Pyk to import the backend's Python bindings, we need to be able to build a wheel from the shared library we're already building.

This PR directly adopts the [pybind example code](https://github.com/pybind/cmake_example) as a way to do this; the intention is for Pyk to then have a stanza in its Makefile that builds the wheel such that Poetry knows about it.

The PR also adds a line to `ciscript` that builds the python wheel under CI.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,3 +48,169 @@ test/int/*.interpreter
 
 *.dSYM
 .envrc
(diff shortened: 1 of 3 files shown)


Baltoli:
I actually don't think this is the right approach, after some more experiments.

PR #554, opened 2022-09-02
Build LLVM backend Python bindings in flake

Baltoli:
This PR makes sure that we include `pybind11` as a pseudo-submodule in the same way as we do for immer and rapidjson. It also adds the flake build as a step in the nix testing workflow.


goodlyrottenapple (approved):
LGTM

Issue #544, opened 2022-08-10
Improve `llvm-kompile` interface


Baltoli:
Currently, the `llvm-kompile` command doesn't support lots of the helpful options that its upstream (`kompile`) and downstream (`llvm-kompile-clang`) commands do. For example, it would be useful to have at least:
* A verbose mode
* Saving of temporary files

We should implement this change in a backwards-compatible way: the existing positional argument interface to the script shouldn't change, and any new options should only appear after the mandatory signature has been consumed.

This should also probably encompass the building of the Type 2 (see #555) binding module as an option; the frontend will then need a matching option.

Current status of the progressive refactor required to implement this change:
* [x] Add new script alongside existing one: #558
* [x] Implement python mode for new script: #559 
* [x] Switch K over to using the new script: https://github.com/runtimeverification/k/pull/2871
* [x] Remove old script from other repos:
  * [x] https://github.com/runtimeverification/avm-semantics/pull/120
  * [x] https://github.com/runtimeverification/llvm-backend/pull/571
  * [x] https://github.com/runtimeverification/c-semantics/pull/713
* [x] Rename `llvm-kompilex` to `llvm-kompile`: *not started*

Baltoli:
Current status of the progressive refactor required to implement this change:
* Add new script alongside existing one: #558
* Switch K over to using the new script: https://github.com/runtimeverification/k/pull/2871
* Remove old script from K: *not started*
* Remove old script from other repos: *not started*
* Rename `llvm-kompilex` to `llvm-kompile`: *not started*

PR #558, opened 2022-09-09
Improve `llvm-kompile` interface (part 1)

Baltoli:
`llvm-kompile` isn't a friendly script to use; it can't pass through options like `-v` and `-save-temps` to its callees, and its documentation is out of date.

This PR makes a start to improving its interface, by duplicating its functionality into a new `llvm-kompilex` script that has nicer option parsing and documentation. We need to have both scripts around for a time so that the frontend can be tested with the new script, after which point the old one can be deleted (cf. the `kprovex` migration). An audit of downstream repos will be required here.

This PR will also make it easier to implement the python binding module compilation (seen partially in https://github.com/runtimeverification/llvm-backend/pull/547)

The reason for the two-step change is that we need to separate the clang flags (e.g. `-g`, `-O0`) from the script flags (`-v` etc.). We do this with a `--` separator, which creates a small incompatibility in the script syntax.

Partial fix for #544


bin/llvm-kompilex
```diff
@@ -0,0 +1,129 @@
...
+      verbose=true
+      kompile_clang_flags+=("-v")
+      shift
+      ;;
+    --save-temps)
+      save_temps=true	    
+      kompile_clang_flags+=("-save-temps")
+      shift
+      ;;
+    --)
```
dwightguth, line 64: I feel like in order to follow standard practices for scripts like this, this should probably be optional.

dwightguth (changes requested):
This looks really good but I did have one suggestion to make

ehildenb:
Options we should make sure that all scripts support moving forward:

- `--profile`: Coarse-grained (process-level) profiling.
- `--debug`: Stack traces, heap dumps, etc...
- `--verbose`: Extra information about what's going on.

dwightguth (approved):

PR #559, opened 2022-09-12
Build type 2 python library

### LINKED ISSUE #556: Split built Python bindings into distinct libraries
For ease of packaging with Pyk, we don't want these libraries to overlap. They should instead split the required functionality into the different components (Type 1, Type 2 per #555).

Steps to implement:
* [x] Build the Type 1 components directly into a shared Python module `kllvm.ast`, etc. We do this already, and Pyk should be able to consume that library from its K dependency.
* [x] #565
* [ ] #566
* [ ] Generate a proper shared library from the static Type 2 one, to expose the runtime binding features.
  * [ ] #558
  * [ ] #559

Baltoli:
~~Subsumes #558, and should only be merged when that PR is merged to master.~~

This PR adds a new method to `llvm-kompile` that builds the type 2 Python bindings for a definition. It will be followed up by a frontend PR that adds a corresponding flag to `kompile` itself, as well as a PR (#568) that implements the runtime (Type 2) binding features.

Fixes #556


bindings/python/runtime.cpp
```diff
@@ -6,6 +6,6 @@ void bind_runtime(py::module_ &m) {
   auto runtime = m.def_submodule("runtime", "K LLVM backend runtime");
 }
 
-PYBIND11_MODULE(kllvm_static, m) {
+PYBIND11_MODULE(kllvm_runtime, m) {
```
tothtamas28, line 9: Is there a way to include the kompiled definiton's name in the module name?

Baltoli: No, unfortunately not without making pretty radical changes to the entire build process for the bindings. This module name needs to be set statically here so that the generated code knows what init function to call when the shared library is loaded by Python.

I'd view this as being similar to `llvm-kompile` generating an executable called `interpreter`; the files are identified by the kompiled directory they live in, rather than their own names.

tothtamas28: I see, makes sense.

tothtamas28 (approved):

PR #562, opened 2022-09-15
Improve performance of macro expansion

Baltoli:
This PR aims to improve the performance of macro expansion by caching a set of symbols that have macros associated with them.


Baltoli:
I've tested this locally against the K regression test suite and everything passes.

theo25 (approved):
This looks good to me. Only left a minor comment.

PR #570, opened 2022-09-26
Fix argument count check for `object.o` case in `llvm-kompile`

Baltoli:
This code path is exercised directly (as far as I know) only in the C semantics, and so previous testing didn't identify a bug whereby the two-argument case isn't accepted.

The code change here is small, and you can see that the documentation in the script already indicates that two positional arguments should be valid.


dwightguth (approved):
That's not the only place where we use that particular code path I think, but this fix looks fine.

PR #571, opened 2022-09-26
Use kompilex in downstream cmake

Baltoli:
This is part of the gradual migration to the new `llvm-kompile` script; downstream users of these CMake modules should get the new script + argument format now.


dwightguth (approved):

theo25 (approved):

Baltoli:
Merged as the automerger is currently off.

PR #584, opened 2022-10-14
Drop kompilex completely

Baltoli:
Needs the following PRs to be merged first:
* [x] https://github.com/runtimeverification/c-semantics/pull/742
* [x] https://github.com/runtimeverification/avm-semantics/pull/141
* [x] https://github.com/runtimeverification/llvm-backend/pull/583
* [x] https://github.com/runtimeverification/k/pull/2992


dwightguth:
If this is the final change, shouldn't we also delete the script itself?

Baltoli:
Ah yes good catch - not sure how I missed that!

dwightguth (approved):

PR #586, opened 2022-10-14
Use opaque pointers when they're available

Baltoli:
This PR opts the `llvm-kompile-*` scripts into using the optional opaque pointer features of LLVM from version 14 onwards. Doing this means that the opaque pointer migration can be incrementally tested more easily (without this change, the existing code won't compile with LLVM 15 until **every** single use of `getPointerElementType` is removed, but won't run on LLVM 14 if any opaque pointers are used).


Baltoli:
Having this enabled seems to cause LLVM 14 LTO to segfault; working on a change that disables LTO on those versions.

dwightguth (approved):

Issue #539, opened 2022-07-04
[Bug] [krun] - Rule application doesn't respect priority attribute when applying on a cell with copies


msaxena2:
K Version
---------

```
$ kompile --version

K version:    5.3.0
Build date:   Mon Jul 04 14:48:51 CDT 2022

```

Description
-----------

Rule application doesn't seem to respect the `priority` attribute when applying on a configuration with some form of multiplicity.

Input Files
-----------

Definition file (named `priorities.md`):
```
module PRIORITIES
  imports DOMAINS

  syntax Pgm ::= "addCell"
               | "foo"
               | "bar"
               | "buzz"
               | Pgm Pgm [right]

  configuration <cells>
                  <cell multiplicity="*" type="Map">
                    <id> 0 </id>
                    <k> $PGM:Pgm </k>
                  </cell>
                </cells>
                <order> .K </order>

  rule <k> addCell => .  ... </k>
       <id> Id </id>
       (.Bag => <cell>
                  <id> Id +Int 1 </id>
                  <k> foo </k> ...
                </cell> )

  rule P1:Pgm P2:Pgm => P1 ~> P2

  rule <k> foo => buzz ... </k>
       <order> . => ListItem(foo) ... </order>

  rule <k> bar => buzz ... </k>
       <order> . => ListItem(bar) ... </order> [priority(250)]

endmodule
```
Program file (named `test.priorities`):
```
addCell bar
```

Reproduction Steps
------------------

 `kompile priorities.md` followed by  `krun test.priorities`.

Expected Behavior
-----------------
When run, the `order` cell contains `ListItem ( foo ) ~> ListItem ( bar ) ~> .`, indicating rule that rewrites `bar` to `buzz`, with a lower priority (indicated by attribute `priority(250)`) fired before the rule that rewrite `foo` to `buzz` with higher (default) priority. 

Expect rule with lower priority to fire last, i.e., rewrite `bar => buzz` to occur after rewrite `foo => buzz`.

Baltoli:
@dwightguth Is [this](https://github.com/runtimeverification/llvm-backend/blob/60478d66beba786090c444bb4300706cbc83a8a3/matching/src/main/scala/org/kframework/backend/llvm/matching/pattern/Pattern.scala#L322-L323) the check you're referring to?

Baltoli:
Fixed by https://github.com/runtimeverification/llvm-backend/pull/587

PR #587, opened 2022-10-14
Fix for priorities not being checked for map choices

Baltoli:
This is the backend fix for https://github.com/runtimeverification/llvm-backend/issues/539; I'll follow it up by adding a test case with Manasvi's code in the frontend.


dwightguth:
This code is finicky enough I'd actually like this test added to the backend integration test suite.

Baltoli:
Added @dwightguth - you can see the correct bar, foo ordering in the KORE output pretty easily.

matching/src/main/scala/org/kframework/backend/llvm/matching/pattern/Pattern.scala
```diff
@@ -200,7 +200,7 @@ case class MapP[T](keys: Seq[Pattern[T]], values: Seq[Pattern[T]], frame: Option
   def isSpecialized(ix: Constructor, isExact: Boolean, fringe: Fringe, clause: Clause, maxPriority: Int): Boolean = {
     (ix, frame) match {
       case (Empty(), _) => keys.isEmpty && values.isEmpty
-      case (HasKey(_, _, _), Some(_)) => true
+      case (HasKey(_, _, _), Some(_)) => clause.action.priority <= maxPriority
```
dwightguth, line 203: I just noticed that this case isn't checking that a choice is occurring before checking the priority. I'm surprised we haven't found any failures running the test suites due to this, but we should still address it.

dwightguth (changes requested):

dwightguth (approved):

PR #592, opened 2022-11-14
Add `llvm-kompile --bindings-path`

Baltoli:
This changes the way that the LLVM backend installs the compiled Type 1 python bindings. Rather than having them be a Poetry package, we add an option to `llvm-kompile` that prints the directory in which the bindings have been installed.


bindings/python/package/pyproject.toml
```diff
@@ -1,17 +0,0 @@
-[tool.poetry]
```
tothtamas28, line 1: If `poetry` is marked as a dependency somewhere in the project, that can be removed as well.

tothtamas28 (approved):

PR #595, opened 2022-11-15
Migrate Jenkins to Github actions

### LINKED ISSUE #581: Migrate from Jenkins to Github actions
There is a general move to Github actions away from Jenkins; the LLVM backend should follow suit

Baltoli:
This PR migrates the LLVM backend CI from Jenkins to Github actions.

The new CI suite is based fully on the Nix flake build, and runs a matrix of build type / LLVM version combinations on each operating system. This matrix supersedes the old `ciscript`.

Reviewers:
* @goodlyrottenapple: The substantive changes to the Nix code are to parameterize the LLVM backend overlay on the CMake build type, and the LLVM version used. Please make sure the Nix changes all make sense to you.
* @ehildenb @F-WRunTime:
  * The self-hosted linux runner doesn't seem to support Nix, and the hosted ones get OOM killed if I run the full set of integration tests. Is there a way to get Nix working on `[self-hosted, linux, normal]`? I'm happy to merge this PR without that, but it would be good to fix at some point.
  * I have left the submodule update code in the Jenkinsfile, as I presume the dependency update job should still be fired through Jenkins.
  * The list of checks should be audited to make sure that the correct ones are required.
* @dwightguth @theo25: No specific review requirements, but please let me know if you have any queries or concerns here.

Fixes #581


goodlyrottenapple (approved):
The nix changes look sensible to me 👍

theo25 (approved):
No concerns from me

PR #599, opened 2022-11-21
Better error messages from hooks

### LINKED ISSUE #593: Better error message when failing to evaluate hooked function symbol
I have a created a bug-report for the LLVM backend on an input, reproduce by doing:

```
mkdir scratch
mv test-bug-report.tar.gz scratch/
cd scratch
tar -xvf test-bug-report.tar.gz
bash command
```

And I get this output:

image

Which is not very informative about what the problem is.

So instead, modify the file `command` and add `--debugger` to the `krun` call there, and then do `run` and `bt` in gdb to get:

image

This output is more helpful, because it tells me that `hook_STRING_string2base_long` crashed. I still can't really see what it crashed on, but just printing out that hook name is a big help.

Can the default output of the `interpreter` binary give a little more informative output here?

[test-bug-report.tar.gz](https://github.com/runtimeverification/llvm-backend/files/10006652/test-bug-report.tar.gz)

Baltoli:
This PR adds a generic identifier to every exception thrown by hooked function symbols in the LLVM backend.

The new output looks like this:
```
terminate called after throwing an instance of 'std::invalid_argument'
  what():  [hook_STRING_string2base_long]: Not a valid integer
```

Note that the exact hook name printed each time may not precisely match the source of the error (because some hooks are implemented in terms of each other - e.g. `int2string` in terms of `base2string`), but this change will at least make it slightly easier to track down what the problem is.

Closes #593

### DIFF
--- a/include/runtime/header.h
+++ b/include/runtime/header.h
@@ -5,8 +5,6 @@
 #include <cstdint>
 #include <limits>
 
-#include <iostream>
-
 #include <gmp.h>
 #include <mpfr.h>
 
--- a/include/runtime/header.h
+++ b/include/runtime/header.h
@@ -5,6 +5,8 @@
 #include <cstdint>
 #include <limits>
 
+#include <iostream>
+
 #include <gmp.h>
 #include <mpfr.h>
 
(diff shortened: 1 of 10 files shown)


dwightguth (approved):
Very nice! We probably also want to try to print the arguments to the function via the macro somehow, but we can leave that for a future pr.

PR #603, opened 2022-11-23
Print hook argument values on crash

Baltoli:
This PR implements hook argument printing for all hook arguments that aren't K terms (i.e. numeric and string arguments). Combined with #599, this means that when a hook crashes in the LLVM backend, you can see precisely which hook it was, as well as the actual argument values that caused the crash. This should save a lot of debugging and investigating time.

~To implement the change without adding additional dependencies to the runtime components of the backend, this PR contains a tiny implementation of a `std::format` lookalike that we should aim to drop whenever C++20 is mainstream.~ We now just pull in `fmt` as it's not an onerous dependency to use.


runtime/strings/numeric.cpp
```diff
@@ -0,0 +1,62 @@
...
+  } else if (prec == 24 && exp == 8) {
+    suffix[0] = 'f';
+    suffix[1] = 0;
+  } else {
+    snprintf(suffix, sizeof(suffix), "p%" PRIu64 "x%" PRIu64, prec, exp);
+  }
+  return floatToString(f, suffix);
+}
+
+std::string intToString(mpz_t i) {
```
dwightguth, line 52: You reused the existing code for converting floats to strings, but not the existing code for converting ints to strings. I would rather we reuse the code that already exists as it has been thoroughly tested.

include/runtime/format.h
```diff
@@ -0,0 +1,66 @@
+#ifndef RUNTIME_FORMAT_H
```
dwightguth, line 1: I believe boost has a version of this library that the c++20 library is based on. We probably want to use it rather than rolling our own implementation.

dwightguth (changes requested):
I left a couple comments. Also, the string library already depends on the int and float libraries; I don't think we probably need to create a whole separate library just for the conversion functions, which ought to significantly simplify this diff.

include/runtime/format.h
Baltoli: As far as I can tell, the [boost library](https://www.boost.org/doc/libs/1_77_0/libs/format/doc/format.html) has a different API to `std::format`. I can drop in [`fmt`](https://github.com/fmtlib/fmt) easily, which is compatible (and the current de facto formatting library for C++).

Baltoli:
> Also, the string library already depends on the int and float libraries; I don't think we probably need to create a whole separate library just for the conversion functions, which ought to significantly simplify this diff.

There's currently no dependency in either direction between these libraries; I can make the arithmetic library depend on the string one in the interests of simplifying the diff.

runtime/strings/numeric.cpp
Baltoli: No problem; will address this in the appropriate way depending on your opinion on the dependency graph question below.

dwightguth (approved):

PR #607, opened 2022-11-28
C bindings for boolean simplification

Baltoli:
This PR exposes C bindings that allow for KORE ASTs to be constructed by wrapping the C++ data structures, and for concrete function terms of sort `Bool` to be simplified.

The changes are:
* Extend `llvm-kompile` to have a new `c` mode that generates a shared library similarly to the Type 2 Python bindings.
* Build a wrapper library for the KORE AST that is usable from C, including a header file that can be pulled in by downstream consumers. This code is largely pretty mechanical, but includes a few convenience functions
* Add a simple test that loads that library from C using `dlopen` etc.
* (minor) fix the `ubuntu-latest` issue in CI

### DIFF
--- a/bindings/c/include/kllvm-c/kllvm-c.h
+++ b/bindings/c/include/kllvm-c/kllvm-c.h
@@ -36,8 +36,10 @@ kore_pattern *kore_string_pattern_new(char const *);
 block *kore_pattern_construct(kore_pattern const *);
 char *kore_block_dump(block *);
 
-// Expects the argument term to be of the form:
-//   sym{}(BOOL)
+/* 
+ * Expects the argument term to be of the form:
+ *   sym{}(BOOL)
+ */
 bool kore_block_get_bool(block *);
 
 bool kore_simplify_bool(kore_pattern *);
@@ -51,7 +53,7 @@ void kore_sort_free(kore_sort const *);
 bool kore_sort_is_concrete(kore_sort const *);
 
 kore_sort *kore_composite_sort_new(char const *);
-void kore_composite_sort_add_argument(kore_sort *, kore_sort *);
+void kore_composite_sort_add_argument(kore_sort const *, kore_sort const *);
 
 /* KORESymbol */
 
@@ -65,4 +67,3 @@ void kore_symbol_add_formal_argument(kore_symbol *, kore_sort const *);
 }
 #endif
 
-#endif
--- a/bindings/c/include/kllvm-c/kllvm-c.h
+++ b/bindings/c/include/kllvm-c/kllvm-c.h
@@ -1,11 +1,15 @@
 #ifndef KLLVM_C_H
 #define KLLVM_C_H
 
+#ifdef __cplusplus
 extern "C" {
+#endif
+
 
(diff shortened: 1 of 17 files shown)


bindings/c/include/kllvm-c/kllvm-c.h
```diff
@@ -0,0 +1,68 @@
...
+kore_pattern *kore_composite_pattern_from_symbol(kore_symbol *);
+void kore_composite_pattern_add_argument(kore_pattern *, kore_pattern *);
+
+kore_pattern *kore_string_pattern_new(char const *);
+
+block *kore_pattern_construct(kore_pattern const *);
+char *kore_block_dump(block *);
+
+// Expects the argument term to be of the form:
+//   sym{}(BOOL)
```
goodlyrottenapple, line 40: can we change these to /* */ style comments so my preprocessor doesn't complain?

bindings/c/include/kllvm-c/kllvm-c.h
```diff
@@ -0,0 +1,68 @@
...
+/* KOREPattern */
+
+char *kore_pattern_dump(kore_pattern const *);
+
+void kore_pattern_free(kore_pattern const *);
+
+kore_pattern *kore_pattern_new_token(char const *, kore_sort const *);
+kore_pattern *kore_pattern_new_injection(
+    kore_pattern *, kore_sort const *, kore_sort const *);
+kore_pattern *kore_pattern_make_interpreter_input(kore_pattern *);
```
goodlyrottenapple, line 28: what's the purpose of this function? do i use it in conjunction with kore_simplify_bool?

bindings/c/lib.cpp
```diff
@@ -0,0 +1,221 @@
...
+kore_sort *kore_composite_sort_new(char const *name) {
+  auto sort = new kore_sort;
+  sort->ptr_ = kllvm::KORECompositeSort::Create(std::string(name));
+  return sort;
+}
+
+void kore_composite_sort_add_argument(kore_sort *sort, kore_sort *arg) {
+  if (auto cast_ptr
+      = dynamic_cast<kllvm::KORECompositeSort *>(sort->ptr_.get())) {
+    cast_ptr->addArgument(arg->ptr_);
```
goodlyrottenapple, line 194: why do we not release the ptr here like in kore_composite_pattern_add_argument?

bindings/c/include/kllvm-c/kllvm-c.h
Baltoli: It's not needed for what you're doing currently, but you might want it in the future if this binding needs to do actual rewrite steps rather than just evaluating functions of sort Bool.

The function essentially mimics what llvm-krun does in terms of composing a call to `initGeneratedTopCell` from a config variable map.

Baltoli:
@theo25 Sam is happy with the API here - could you cast your eye over the code to make sure nothing is obviously wrong with my C wrappers? Thanks!

theo25 (approved):
Things looks fine to me

PR #610, opened 2022-11-30
Don't run maven update job if we don't need to

Baltoli:
This ports in the equivalent change from the K repo; we don't need to be spending lots of time updating the maven dependencies unless we need to.

### DIFF
--- a/.github/workflows/update.yml
+++ b/.github/workflows/update.yml
@@ -13,15 +13,21 @@ jobs:
     name: 'Nix: Maven'
     runs-on: ubuntu-latest
     steps:
-      - name: 'Check out code'
+      - name: 'Check out code, set up Git'
         uses: actions/checkout@v2.3.4
         with:
           submodules: recursive
+          fetch-depth: 0
+      - run: |
+          git config user.name devops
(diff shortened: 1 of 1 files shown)


.github/workflows/update.yml
```diff
@@ -13,15 +13,21 @@ jobs:
     name: 'Nix: Maven'
     runs-on: ubuntu-latest
     steps:
-      - name: 'Check out code'
+      - name: 'Check out code, set up Git'
         uses: actions/checkout@v2.3.4
```
F-WRunTime, line 17: Lets bump this v3 for consistency

F-WRunTime:
You mention in the header you want to run this less frequently the trigger however runs on every pull request is that still the frequency you want to see after this PR?

Baltoli:
> every pull request is that still the frequency you want to see after this PR?

Yes, it should run on every PR but only do the costly Maven+Nix work if something will have changed (i.e. a pom.xml or a nix file). This is the same logic as K; the projects have a similar structure wrt. this code

Baltoli:
Ping @F-WRunTime

F-WRunTime (approved):

PR #613, opened 2022-12-12
Version updates and nix version pin, OS runner pin

F-WRunTime:


F-WRunTime:
Do these look worthwhile to include and update?

F-WRunTime:
Check will not pass until we update it in the settings.

Baltoli (approved):

PR #614, opened 2022-12-13
Hotfix for shellcheck breaking

Baltoli:
The recent ubuntu update to CI runners has updated shellcheck, which includes this new check for unreachable code. Our usage is indirect via signal trapping, and so we can safely ignore it.


theo25 (approved):
looks fine to me

PR #620, opened 2022-12-21
Rangemap

mariaKt:
This PR implements an immutable ordered map based on a red-black tree, and an immutable rangemap using the ordered map. I have chosen to go with our own implementation instead of looking for another one, since this would be easier for garbage collection. Unit tests are included, attempting to cover many cases, especially for the many different cases of insertion/deletion for the ranges.

There will be a follow up PR to introduce hooks for the rangemap, and handle GC (and pattern matching, if necessary).

Brief note: As I discussed in a meeting, I have chosen not to implement iterators for the treemap. This was because implementing them using the standard approach would change the time and space complexity of the operations on the immutable RB tree. Moreover, they were not needed for the range map implementation. So I defer it for when/if we ever need it, since we already have a full map interface.


include/runtime/collections/RBTree.h
```diff
@@ -0,0 +1,512 @@
...
+#include <memory>
+#include <stdexcept>
+#include <string>
+
+#define CONSTRUCT_MSG_AND_THROW(msg)                                           \
+  do {                                                                         \
+    std::string emsg = std::string("[") + std::string(__func__)                \
+                       + std::string("] ") + std::string(msg);                 \
+    throw std::invalid_argument(emsg);                                         \
+  } while (false)
```
Baltoli, line 18: We have pulled in `fmt` as a dependency in an [in-flight PR](https://github.com/runtimeverification/llvm-backend/pull/603); we should rewrite this code to use that new infrastructure once one or the other is merged. It's fine for now as the dependency isn't available in this PR.

include/runtime/collections/RBTree.h
```diff
@@ -0,0 +1,512 @@
...
+    default: CONSTRUCT_MSG_AND_THROW("Unexpected color");
+    }
+  }
+
+  struct Node {
+    Node(Color c)
+        : _c(c)
+        , _s(0) { }
+    Color _c;
+    size_t _s;
```
Baltoli, line 43: Underscore-prefixed member variables aren't UB, but I tend to avoid them in favour of suffixed names as it's a little bit easier to avoid ending up with UB if you rename something.

include/runtime/collections/RBTree.h
```diff
@@ -0,0 +1,512 @@
...
+    virtual ~Node(){};
+  };
+
+  struct Leaf : public Node {
+    Leaf(Color c)
+        : Node(c) {
+      assert(c == B || c == BB);
+    }
+    virtual bool isLeaf() const override { return true; }
+    virtual ~Leaf(){};
```
Baltoli, line 55: Best practice tends to suggest making empty special member functions `= default;` rather than explicitly having empty bodies. Repeats elsewhere.

include/runtime/collections/RBTree.h
```diff
@@ -0,0 +1,512 @@
...
+  };
+
+  struct InternalNode : public Node {
+    InternalNode(
+        Color c, std::shared_ptr<const Node> const &lft, T key, V val,
+        std::shared_ptr<const Node> const &rgt)
```
Baltoli, line 61: Given that the constructor immediately constructs the member variables `_lft` and `_rgt` with these const-ref arguments, they should just be taken by value in the first place. This helps the compiler figure out when to make copies a bit better than copying from a const-ref. This repeats elsewhere.

include/runtime/collections/RBTree.h
```diff
@@ -0,0 +1,512 @@
...
+    else
+      return rootVal();
+  }
+
+  RBTree inserted(T x, V v) const { return ins(x, v).blacken(); }
+
+  RBTree deleted(T x) const { return redden().del(x); }
+
+  // 1. No red node has a red child.
+  void assert1() const {
```
Baltoli, line 171: This could have a better name if we're exposing it as part of a public interface.

include/runtime/collections/RBTree.h
```diff
@@ -0,0 +1,512 @@
...
+    assert(!isEmpty());
+    return RBTree(c, left(), root(), rootVal(), right());
+  }
+
+private:
+  std::shared_ptr<const Node> _root;
+};
+
+template <class T, class V, class F>
+void forEach(RBTree<T, V> const &t, F f) {
```
Baltoli, line 482: Again nitpicking on best practice here - ideally this would take `f` as a universal reference `F&&`, then use `std::invoke` to call it, and `std::forward` to pass it on to the recursive calls.

include/runtime/collections/rangemap.h
```diff
@@ -0,0 +1,380 @@
...
+    T r2end = r2.second;
+    if (r2end < r1start)
+      return false;
+    else if (r1end < r2start)
+      return false;
+    return true;
+  }
+
+  std::optional<std::pair<Range, V>>
+  getKeyValue(RBTree<Range, V> t, T k) const {
```
Baltoli, line 60: Let's standardise on `snake_case` here rather than mixing naming conventions.

include/runtime/collections/rangemap.h
```diff
@@ -0,0 +1,380 @@
...
+    _map = m._map;
+  }
+
+  RBTree<Range, V> getTreeMap() const { return _map; }
+
+  // Return the number of key ranges in the map.
+  size_t size() const { return _map.size(); }
+
+  // Return true if a range in the map contains the key.
+  bool containsKey(T k) const {
```
Baltoli, line 217: See above re. naming - this interface would be a bit more "stl-like" if it were just `contains`.

include/runtime/collections/rangemap.h
```diff
@@ -0,0 +1,380 @@
...
+    std::optional<std::pair<Range, V>> opt = getKeyValue(k);
+    if (opt.has_value())
+      return true;
+    return false;
```
Baltoli, line 221:
```suggestion
    return getKeyValue(k).has_value();
```

include/runtime/collections/rangemap.h
```diff
@@ -0,0 +1,380 @@
...
+  }
+
+  // If the key is contained in any range in the map, return the value
+  // associated with the key.
+  std::optional<V> getValue(T k) const {
```
Baltoli, line 226: See the upstream method comment - do we definitely need to copy the key to get the associated value?

include/runtime/collections/rangemap.h
```diff
@@ -0,0 +1,380 @@
...
+  bool overlaps(Range r) const { return overlaps(_map, r); }
+
+  void print() {
+    std::cout << "--------------------------" << std::endl;
+    forEach(_map, [](Range x, V v) {
+      std::cout << "[ " << x.first << ".." << x.second << " ) -> " << v
+                << std::endl;
+    });
+    std::cout << "--------------------------" << std::endl;
+  }
```
Baltoli, line 352: If we're exposing `print()` in the public namespace for debugging, let's add an argument so that it doesn't always have to go to `std::cout`.

include/runtime/collections/rangemap.h
```diff
@@ -0,0 +1,380 @@
...
+};
+
+template <class T, class V, class I>
+RangeMap<T, V> inserted(RangeMap<T, V> m, I it, I end) {
+  if (it == end)
+    return m;
+  typename RangeMap<T, V>::Range key = it->first;
```
Baltoli, line 359: Good case for `auto` here I think!

include/runtime/collections/rangemap.h
```diff
@@ -0,0 +1,380 @@
...
+
+#include <cassert>
+#include <iostream>
+#include <optional>
+#include <stdexcept>
+
+template <class T, class V>
+class RangeMap {
+public:
+  using Range = std::pair<T, T>;
```
Baltoli, line 14: There's enough functionality hooked into `Range` below that I would consider extracting it out into its own class; this makes your test + user code for the rangemap a bit nicer to read as well.

Baltoli (commented):
I've left a number of somewhat pedantic comments largely to do with code style and C++ object semantics; the algorithmic side of this looks great to me!

I know that not all of this code is yours and has been adapted from Milewski's code, but given we're making some algorithmic changes to it it's worth making sure it's all consistent style-wise.

unittests/runtime-collections/rangemaps.cpp
```diff
@@ -0,0 +1,596 @@
...
+
+BOOST_AUTO_TEST_CASE(rangemap_test_empty) {
+  auto map = RangeMap<int, int>();
+  auto result = map.size();
+  BOOST_CHECK_EQUAL(result, 0);
+  auto result2 = map.containsKey(0);
+  BOOST_CHECK_EQUAL(result2, false);
+}
+
+BOOST_AUTO_TEST_CASE(rangemap_test_concat_success) {
```
dwightguth, line 15: I don't understand this test. If my understanding is correct, map5 ought to look like:

```
[0, 2) -> 2
[2, 7) -> 1
[8, 9) -> 4
```

So its size ought to be 3, but you assert that it's 1, as if all the ranges got combined. But you can't combine two ranges that map to different values.

Am I missing something?

unittests/runtime-collections/rangemaps.cpp
```diff
@@ -0,0 +1,596 @@
...
+  result = map5.size();
+  BOOST_CHECK_EQUAL(result, 1);
+}
+
+BOOST_AUTO_TEST_CASE(rangemap_test_concat_failure) {
+  auto m = (RangeMap<int, int>()).inserted(std::make_pair(2, 4), 1);
+  auto map1 = m.inserted(std::make_pair(6, 7), 2);
+  auto m1 = (RangeMap<int, int>()).inserted(std::make_pair(0, 3), 2);
+  auto map2 = m1.inserted(std::make_pair(8, 9), 4);
+  BOOST_CHECK_THROW(concat(map1, map2), std::invalid_argument);
```
dwightguth, line 35: now I'm particularly confused. You appear to be concatenating two maps with overlapping ranges and it's generating an error, but the same thing actually happens in the previous test for the range `[4, 6)` and no error occurs. Again, am I missing something?

include/runtime/collections/RBTree.h
```diff
@@ -0,0 +1,512 @@
...
+    return _root->_c;
+  }
+
+  Color leafColor() const {
+    assert(isEmpty());
+    return _root->_c;
+  }
+
+public:
+  RBTree()
```
dwightguth, line 92: all of these public operations need documentation describing what they do and what arguments they take.

dwightguth (dismissed):
I'll review in more detail once I understand things a bit better. Let's start with more documentation (good start though) and some clarifications to my questions regarding some of the tests,

unittests/runtime-collections/rangemaps.cpp
mariaKt: map5 is the concatenation of map1 ( [2,4)->1, [6,7)->1 ) and map4 ( [4,6)->1 ), and therefore is the map [2,7)->1 because all ranges map to the same value. I am reusing map1 here to also test for immutability.

unittests/runtime-collections/rangemaps.cpp
mariaKt: Right, concatenating two maps with overlapping ranges generates an error. I followed the unordered map implementation in K for this, where if the maps to concatenate share a key an error is generated. There is no contradiction with the previous example, as there were no overlapping ranges in that case.

include/runtime/collections/RBTree.h
Baltoli: The PR mentioned is merged, so this code can be moved over to FMT.

include/runtime/collections/rangemap.h
```diff
@@ -0,0 +1,380 @@
...
+    T start = r.first;
+    if (k < start)
+      return getKeyValue(t.left(), k);
+    assert(k >= r.second);
+    return getKeyValue(t.right(), k);
+  }
+
+  bool overlaps(RBTree<Range, V> const &t, Range r) const {
+    if (t.isEmpty())
+      return false;
```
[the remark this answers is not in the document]
mariaKt: Just to clarify my understanding here, do you mean that I should use brackets around the body of the `if` statement (regardless of the number of statements)?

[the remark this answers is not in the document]
mariaKt: [Link](https://runtimeverification.slack.com/archives/C5GE8MBC6/p1674271134666499) to follow discussion in slack.

include/runtime/collections/rangemap.h
mariaKt: Range is not extracted into its own class.

include/runtime/collections/rangemap.h
mariaKt: Addressed.

include/runtime/collections/rangemap.h
mariaKt: Right!

include/runtime/collections/RBTree.h
mariaKt: Addressed.

include/runtime/collections/RBTree.h
mariaKt: I changed to names that are followed by `_` instead.

include/runtime/collections/RBTree.h
```diff
@@ -0,0 +1,512 @@
...
+
+//---               Ordered map on top of a red-black tree.                ---//
+
+// 1. No red node has a red child.
+// 2. Every path from root to empty node contains the same
+// number of black nodes.
+
+template <class T, class V>
+class RBTree {
+  enum Color { R, B, BB };
```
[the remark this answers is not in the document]
mariaKt: Addressed.

include/runtime/collections/RBTree.h
mariaKt: I changed this in the constructors where a copy would immediately follow. There is one case left where this is not the case, rather members of the arguments are given to the constructor that then performs the actual copies.

include/runtime/collections/RBTree.h
mariaKt: Addressed.

include/runtime/collections/rangemap.h
mariaKt: Using `snake_case` throughout.

include/runtime/collections/rangemap.h
```diff
@@ -0,0 +1,380 @@
...
+    else if (r1end < r2start)
+      return false;
+    return true;
+  }
+
+  std::optional<std::pair<Range, V>>
+  getKeyValue(RBTree<Range, V> t, T k) const {
+    if (t.isEmpty())
+      return std::optional<std::pair<Range, V>>();
```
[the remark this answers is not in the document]
mariaKt: Addressed, in other instances as well.

include/runtime/collections/rangemap.h
```diff
@@ -0,0 +1,380 @@
...
+    Range r = t.root();
+    if (range_contains(r, k))
+      return std::make_optional<std::pair<Range, V>>(
+          std::make_pair(r, t.rootVal()));
```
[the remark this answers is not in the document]
mariaKt: Addressed.

include/runtime/collections/rangemap.h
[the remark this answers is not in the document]
mariaKt: In `contains`, `get_value`, `get_key_value`, everything is now passed as `const ref`. Starting from this, I have additionally identified a few more similar places where arguments could be passed as references and the `const` qualifier could be added, e.g. `overlaps`.

include/runtime/collections/rangemap.h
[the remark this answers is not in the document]
mariaKt: Addressed.

include/runtime/collections/rangemap.h
```diff
@@ -0,0 +1,380 @@
...
+      // does not share a boundary. It is not relevant. Continue looking
+      // for relevant ranges to the left of the candidate range.
+      getInsertionRelevantRanges(t.left(), r, v);
+    } else if (start <= rstart && rend <= end) {
+      // Our target range contains the candidate range, and may share its
+      // boundaries (i.e. equal or larger). The candidate range is
+      // relevant, and there may be relevant ranges in both left and
+      // right directions.
+      getInsertionRelevantRanges(t.left(), r, v);
+      v.push_back(std::make_pair(t.root(), t.rootVal()));
```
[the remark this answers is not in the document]
mariaKt: Addressed.

include/runtime/collections/rangemap.h
mariaKt: Changed to `contains`.

include/runtime/collections/rangemap.h
mariaKt: No. The key is now passed as reference, as discussed above.

include/runtime/collections/RBTree.h
mariaKt: Added more documentation, specifically to public methods.

include/runtime/collections/RBTree.h
mariaKt: Using FMT in implementing the error printing macro.

include/runtime/collections/rangemap.h
mariaKt: With `Range` being a separate class (see below), I am able to write `Range<T>`instead. Of course I can change it to `auto` if you prefer that.

include/runtime/collections/RBTree.h
```diff
@@ -6,17 +6,18 @@
...
-#define CONSTRUCT_MSG_AND_THROW(msg)                                           \
+#define CONSTRUCT_MSG_AND_THROW(...)                                           \
   do {                                                                         \
-    std::string emsg = std::string("[") + std::string(__func__)                \
-                       + std::string("] ") + std::string(msg);                 \
-    throw std::invalid_argument(emsg);                                         \
+    auto err_msg                                                               \
+        = ::fmt::format("[{}]: {}", __func__, ::fmt::format(__VA_ARGS__));     \
+    throw std::invalid_argument(err_msg);                                      \
   } while (false)
```
Baltoli, line 21: This code is duplicated here: https://github.com/runtimeverification/llvm-backend/blob/fd2f7cc89cee542ba873c7f73f100a277e948d17/include/runtime/header.h#L282-L287

include/runtime/collections/RBTree.h
```diff
@@ -0,0 +1,606 @@
...
+  RBTree(Color c, RBTree const &lft, T key, V val, RBTree const &rgt)
+      : root_(std::make_shared<const InternalNode>(
+          c, lft.root_, key, val, rgt.root_)) {
+    assert(lft.is_empty() || lft.root_key() < key);
+    assert(rgt.is_empty() || key < rgt.root_key());
+  }
+
+  // Copy constructor.
+  explicit RBTree(std::shared_ptr<const Node> node)
+      : root_(node) { }
```
Baltoli, line 111: Nitpicking but this isn't a true C++ copy constructor; it constructs a new tree from a _node_ rather than from a different tree.

include/runtime/collections/RBTree.h
```diff
@@ -0,0 +1,606 @@
...
+
+  // Return this Node's color when it is not empty.
+  Color root_color() const {
+    assert(!is_empty());
```
Baltoli, line 115: In which situations will the data structure assert, and in which will it throw an exception?

include/runtime/collections/RBTree.h
```diff
@@ -0,0 +1,606 @@
...
+  RBTree(I b, I e) {
+    RBTree t(Color::B);
+    for_each(b, e, [&t](std::pair<T, V> const &p) {
+      t = t.inserted(p.first, p.second);
+    });
+    root_ = t.root_;
+  }
+
+  // Return true if this tree is empty.
+  bool is_empty() const { return root_->is_leaf(); }
```
Baltoli, line 143: `empty` is a little more idiomatic to the STL

include/runtime/collections/RBTree.h
```diff
@@ -0,0 +1,606 @@
...
+    assert(!is_empty());
+    const InternalNode *r = static_cast<const InternalNode *>(root_.get());
+    return RBTree(r->rgt_);
+  }
+
+  // Return the size of this tree, i.e., the number of non-leaf nodes.
+  size_t size() const { return root_->s_; }
+
+  // Return true if key x is found in this tree. Otherwise, return false.
+  bool member(T const &x) const {
```
Baltoli, line 177: Similarly, `contains` is the STL-alike name for this function

include/runtime/collections/RBTree.h
```diff
@@ -0,0 +1,606 @@
...
+    } else if (y < x) {
+      return right().member(x);
+    } else {
+      return true;
+    }
+  }
+
+  // Return the corresponding value if key x is found in this tree. Otherwise,
+  // throw an exception.
+  V lookup(T const &x) const {
```
Baltoli, line 193: Again, the usual name for this is `at`, and ideally it would return a const-ref to the item in the tree so that we don't copy anything the caller doesn't want copied.

include/runtime/collections/RBTree.h
```diff
@@ -0,0 +1,606 @@
...
+template <class T, class V, class I>
+RBTree<T, V> inserted(RBTree<T, V> const &t, I it, I end) {
+  if (it == end) {
+    return t;
+  }
+  T key = it->first;
+  V val = it->second;
+  auto t1 = t.inserted(key, val);
+  return inserted(t1, ++it, end);
+}
```
Baltoli, line 604: I'm not keen on these getting leaked into the global namespace every time someone includes this header; the structure code could be kept in an `rb_tree` namespace or similar.

include/runtime/collections/rangemap.h
```diff
@@ -0,0 +1,426 @@
...
+
+  bool operator!=(Range const &other) const {
+    return *this < other || other < *this;
+  }
+
+  // Returns true if this range contains k.
+  bool contains(T const &k) const { return k >= start_ && k < end_; }
+
+  // Returns true if this range is empty.
+  bool is_empty() const { return start_ >= end_; }
```
Baltoli, line 56: `empty` is preferable as previous comments

include/runtime/collections/rangemap.h
```diff
@@ -0,0 +1,426 @@
...
+  // Create a new range [start, end).
+  Range(T start, T end)
+      : start_(start)
+      , end_(end) { }
+
+  // Getter for the start of this range.
+  T get_start() const { return start_; }
+
+  // Getter for the end of this range.
+  T get_end() const { return end_; }
```
Baltoli, line 29: I prefer `start` and `end` here

include/runtime/collections/rangemap.h
```diff
@@ -0,0 +1,426 @@
...
+template <class T, class V, class I>
+RangeMap<T, V> inserted(RangeMap<T, V> const &m, I it, I end) {
+  if (it == end) {
+    return m;
+  }
+  Range<T> key = it->first;
+  V val = it->second;
+  auto m1 = m.inserted(key, val);
+  return inserted(m1, ++it, end);
+}
```
Baltoli, line 424: See previous comment re. namespaces

Baltoli (commented):
Some very minor comments mostly on naming and hygiene

Baltoli (approved):

include/runtime/collections/RBTree.h
mariaKt: The data structure uses assertions in cases where I want the check to happen in the Debug build, but not necessarily in the Release build. The exceptions are thrown where there is a failure that I want the user to be notified about, e.g., that the failure is due to an empty range given to insert or delete (user caused), or after a check on the data structure has been requested (one of the invariant checks, where any failure should be reported).

include/runtime/collections/RBTree.h
mariaKt: I realize that. The issue is that I intentionally avoided including `runtime/header.h` and use the `KLLVM_HOOK_INVALID_ARGUMENT` directly. The reason is that I intend to include the header file `rangemap.h` (which includes `RBTree.h`) in `include/runtime/header.h` when the hooks PR is merged.

Would it be possible to either leave this as is, or extract the `KLLVM_HOOK_INVALID_ARGUMENT` and `fmt` dependency in a separate (e.g., `fmt_error_printing.h`) header file in `include/runtime`, and including that in both `include/runtime/header.h` and my header file as part of a separate PR?

include/runtime/collections/RBTree.h
mariaKt: Addressed.

include/runtime/collections/RBTree.h
mariaKt: Renamed to `empty`.

include/runtime/collections/RBTree.h
mariaKt: Renamed to `contains`.

include/runtime/collections/rangemap.h
mariaKt: Renamed.

include/runtime/collections/rangemap.h
mariaKt: Renamed.

dwightguth:
It looks like you haven't pushed the changes that you made to address some of Bruce's comments yet, btw.

include/runtime/collections/RBTree.h
mariaKt: The structure is now enclosed in namespace `rb_tree`.

include/runtime/collections/rangemap.h
mariaKt: As above, the structure is enclosed in namespace `rangemap`.

include/runtime/collections/RBTree.h
mariaKt: Renamed. I am now working on the `const` part, which might end up being a larger change. I will follow up with an explanation as soon as I complete this.

mariaKt:
Sorry about the confusion. The changes are now pushed.

include/runtime/collections/RBTree.h
Baltoli: My preference would be to extract the formatting code to a separate header included by both the datastructure and `header.h`

include/runtime/collections/RBTree.h
Baltoli: That's fine, just wanted to check there was a consistent choice!

Baltoli:
I'm happy to merge this once the two outstanding comments have been resolved (const-ref return from `at`, and extracting the `fmt` macro to a separate header).

include/runtime/collections/RBTree.h
mariaKt: Addressed.

include/runtime/collections/RBTree.h
mariaKt: `at` now returns const-ref. The keys and values of the map are returned as const-ref as well, to allow us to avoid some more copies of objects of types `T`, `V`. In the last commit I address the same issue, and even though local variables are used, copies of those are eventually made in the tree constructors.

PR #624, opened 2023-01-10
Handle simplifications in `SortKItem`

### LINKED ISSUE #622: Redundant injections created by `kore_simplify` abort the backend
@goodlyrottenapple notes that:
> the kore_simplify function assumes the term is not of sort k, and when it is, it does inj{SortK, SortK}(...)

Somewhere in the backend, this causes an `abort()` to get thrown. The immediate solution is to avoid creating these injections in `kore_simplify`.

Baltoli:
We don't want to inject terms that are already of sort `KItem` into `KItem` again; doing so crashes the interpreter. This PR checks if the sort passed to `kore_simplify` is `SortKItem`, and if it is, doesn't construct an injection.

The PR also includes a test that exercises this code.

Fixes #622

### DIFF
--- a/bindings/c/include/kllvm-c/kllvm-c.h
+++ b/bindings/c/include/kllvm-c/kllvm-c.h
@@ -87,6 +87,8 @@ void kore_sort_free(kore_sort const *);
 
 bool kore_sort_is_concrete(kore_sort const *);
 
+bool kore_sort_is_kitem(kore_sort const *);
+
 kore_sort *kore_composite_sort_new(char const *);
 void kore_composite_sort_add_argument(kore_sort const *, kore_sort const *);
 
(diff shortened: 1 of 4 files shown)


ehildenb:
What I don't quite understand is why we would inject things into sort KItem by default _anyway_? Shouldn't we just leave the term the user gave us alone, and simplify it directly? Is this because of some internals on how the LLVM backend simplification works?

Baltoli:
Just an artifact of how this entrypoint to the simplifier works; it's implemented this way for maximum simplicity in the interests of getting a working prototype. I _think_ you could relax the requirement to inject into KItem but I would need to have a look through the internals to figure that out.

jberthold (approved):
I was able to use this  branch in tests with `evm-semantics` + `hs-backend-booster`, appears to work well.

PR #631, opened 2023-01-16
Expose length parameter for tokens

### LINKED ISSUE #629: LLVM library API truncates byte arrays at first null byte
While experimenting with the Haskell bindings to the LLVM simplification API , it turned out that byte arrays (like the `<program>` in the EVM configuration) get truncated at the first null byte.
This is because byte arrays are represented as domain values, whose content is represented as a string (the string must be allowed to contain null bytes when it represents a byte array).

It is not clear where exactly the truncation happens, but the [code in `kore_string_pattern_new`](https://github.com/runtimeverification/llvm-backend/blob/c4fe2ce4da12e80a812b2529c7c8a01e3f135ae7/bindings/c/lib.cpp#L181) is probably one of those places (the Haskell wrapper code is another).

Fixing this issue will probably mean that the API functions dealing with string values must take length information instead of using null-terminated strings.

Baltoli:
This PR exposes new C API functions for creating KORE string patterns and tokens (`\dv{Sort}{"..."}`); the new functions accept a _length_ as well as the string data to be used for the token value. This means that tokens that contain null bytes can be constructed; previously, the construction of a `std::string` would truncate them.

The test for this change is similar to the other C bindings tests.

Closes #629

### DIFF
--- a/bindings/c/include/kllvm-c/kllvm-c.h
+++ b/bindings/c/include/kllvm-c/kllvm-c.h
@@ -51,8 +51,12 @@ char *kore_pattern_dump(kore_pattern const *);
 void kore_pattern_free(kore_pattern const *);
 
 kore_pattern *kore_pattern_new_token(char const *, kore_sort const *);
+kore_pattern *
+kore_pattern_new_token_with_len(char const *, size_t, kore_sort const *);
+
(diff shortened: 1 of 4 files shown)


goodlyrottenapple (approved):
lgtm

PR #634, opened 2023-01-24
New Python bindings build structure

### LINKED ISSUE #611: Change Python bindings build process
Currently, the approach we take is to use the Pybind CMake integration to build the AST bindings for the backend at the time the LLVM backend is built. This has the advantage of being easy to integrate with the existing LLVM backend build system, but falls down when we try to ship the bindings.

[Pybind uses the unstable API for binding to Python][pybind], which means that the generated libraries are **not** minor-version compatible. This manifests as the [bindings not being found by the Pyk library][pyk].

@tothtamas28 and I have discussed this issue, and the best thing to do is likely going to be a change to the build system that allows for the bindings to be generated at run time, using the correct Python interpreter (similarly to what we do for the Type 2 runtime bindings).

[pyk]: https://github.com/runtimeverification/pyk/issues/119
[pybind]: https://github.com/pybind/pybind11/issues/1755

Baltoli:
This PR fixes #611 by building both parts of the LLVM backend python bindings from source when they are _used_, rather than when the LLVM backend itself is built. This is a workaround for Pybind using minor-version specific features of the Python API, meaning that if the backend build used Python 3.10, and at runtime only 3.11 was available, the bindings would break.

The signature and functionality of the existing `llvm-kompile defn.kore dt python` code path (for the *runtime* bindings) has not changed; all the existing tests for that feature pass, but the binding will be built from source rather than from a static library. To compile the AST bindings, a new signature `llvm-kompile pythonast` has been added. This does not depend on a definition being available, and will generate a `_kllvm.$version.so` shared library compatible with the one previously shipped and installed by the LLVM backend.

For both of these signatures, `llvm-kompile` now has a `--python CMD` option that specifies the precise Python interpreter to be used to generate the bindings; this interpreter must be able to find a Pybind11 installation to do so.

To implement these changes, we now install the LLVM backend headers and binding source files as part of the backend install package. There is no way to avoid doing this, as far as I can tell (i.e. precompiling a binary for the bindings), as Pybind is a header-only library.

I will follow this PR up with several smaller ones that clean up the old compilation method for the AST bindings, which I have left intact to keep this diff small and avoid breaking changes to Pyk.

Notes for reviewers:
- @dwightguth please make sure that the changes to the shell scripts here seem OK to you.
- @goodlyrottenapple there are some changes to work the new binding structure through into the Nix build - I think I've done the right thing but let me know if any of it doesn't make sense.
- @tothtamas28 have a look at the new workflow for generating the Type 1 / AST bindings and see if it makes sense to you from a Pyk perspective.


Baltoli:
@ehildenb this should solve your issue as discussed once Pyk picks up the new installation method.

tothtamas28 (commented):
I created a PR to test these changes: https://github.com/runtimeverification/pyk/pull/169

The `pythonast` binding seems to be working as expected.

Integration tests for the runtime binding fail with the following error when running `llvm-kompile python ...`:

```
In file included from /home/ttoth/git/k/k-distribution/target/release/k/lib/kllvm/python_src/runtime.cpp:7:
/home/ttoth/git/k/k-distribution/target/release/k/bin/../include/kllvm/runtime/header.h:8:10: fatal error: 'fmt/format.h' file not found
#include <fmt/format.h>
         ^~~~~~~~~~~~~~
1 error generated.
```

Baltoli:
The solution here is to install `fmt`; `libfmt-dev` on Ubuntu. This is already a K dependency, but is now required to build the runtime bindings.

tothtamas28 (approved):

dwightguth:
This looks fine to me as long as we've adequately tested that all the existing ways we invoke llvm-kompile haven't been broken by this change.

PR #639, opened 2023-01-27
Fix printing of ML patterns

Baltoli:
The logic in `kprint` for handling ML symbols appears to be out of date:
* The backend assumes that the modal symbols are `\weak...`, rather than `weak...` as is actually used by the Haskell backend.
* No special-casing of `weakAlwaysFinally` was present, meaning that `kprint` would crash when handling that symbol.

This PR updates the logic to reflect the KORE that is actually generated by the Haskell backend.

Fixes https://github.com/runtimeverification/k/issues/3105. Instead of crashing, the claim in that issue will print correctly:
```
    <T>
      <k>
        while (! n <= 0) {
          sum = sum + n + n;
          n = n + -1;
        } ~> _DotVar2:K
      </k>
      <state>
        n |-> N:Int
        sum |-> S:Int
      </state>
    </T>
  #And
    {
      true
    #Equals
      N:Int >=Int 0
    }
#Implies
  #wAF ( <T>
    <k>
      _DotVar2:K
    </k>
    <state>
      n |-> 0
      sum |-> S:Int +Int ( N:Int +Int 1 ) *Int N:Int
    </state>
  </T> )
```


radumereuta (changes requested):
lgtm but, can we add a test?

Baltoli:
> lgtm but, can we add a test?

Ah, good point - I forgot there was a test suite in the backend for `kprint`. Will add that

radumereuta (approved):

PR #642, opened 2023-01-27
Update ownership model in C bindings

Baltoli:
This PR updates the C types used to bridge to the C++ AST structure such that symbols and terms are stored as reference-counted pointers (rather than owning pointers as previously). This allows symbols and terms to be cached if needed, and for terms to be reused (e.g. you can add a term to two different parents, or simplify the same term twice).

It also cleans up the internal memory leakage in the bindings code; I've tested some examples by hand using ASan for `kore_simplify`, `kore_simplify_bool` etc.


goodlyrottenapple (approved):
Ive tested this in the booster backend and everything works as expected. I can cache any pattern / symbol / sort.

PR #644, opened 2023-02-01
Make serialization and printing green-thread-safe

### LINKED ISSUE #643: Make serializer API thread-safe
The binary serializer (and textual printer) use `static thread_local` storage to maintain state within a call. This causes issues when multiple calls are made to the serializer at the same time from *non-OS threads* (e.g. within the Haskell runtime).

We should thread the static state through each call so that the new haskell backend can do multiple simplifications at a time.

Making this change will require a change to the visitor code-generation; I suspect that the best way to do this without a great deal of added complexity is to add an extra pointer argument that the C++ code can treat as a state structure.

Baltoli:
Original identification by @jberthold: https://github.com/runtimeverification/hs-backend-booster/issues/94#issuecomment-1409912409

In the LLVM backend, the printing and serialization code uses `static thread_local` storage to maintain state throughout a single printing / serialization call. At each call, the state is explicitly cleared before entering the internal routines.

This works fine so long as the code is not used in the context of a _bindings library_ where a runtime system such as Haskell has its own *non-OS* threading model. In these cases, the code is no longer thread-safe, and will produce garbage results if called in parallel.

The solution to this issue is to explicitly thread the state previously kept in static data through each call to the printing and serialization routines. Because the block visitors trampoline back and forth between C++ and code-generated LLVM, this requires some changes to the logic that emits a block traversal.

The changes in this PR are roughly as follows:
* Add an extra `void *` parameter to the C++ / LLVM recursion scheme code that traverses runtime terms when printing and serializing.
* Wrap up the existing static, thread-local state into a single structure type.
* Pass an instance of that structure as an extra argument through the printing / serialization code.
* (small, unrelated) move the `MatchLog` code to its own file away from the printer code.

The diff is large but mostly mechanical, as there are a lot of places that need to be updated to account for the extra parameter.

I'm confident that the existing test suite is sufficient to exercise this change fully; any breakage of configuration printing / serialization would be caught by every test in the suite.

Fixes #643


dwightguth (approved):
Looks good. Have you tested that it solves @jberthold 's original issue?

Baltoli:
@jberthold confirmed that the issue remains on his side.

I'm going to merge this PR anyway, for the following reasons:
* The code has been reviewed and CI is passing.
* I suspect that the changes here are _necessary_ but not _sufficient_ to make the backend fully thread-safe.
* The diff for this PR is largely mechanical (i.e. threading the extra parameter through the visitor recursion scheme). Further related changes will be harder to review if mixed in with this PR.

PR #647, opened 2023-02-03
Audit header includes

### LINKED ISSUE #627: Add include-what-you-use to clang checks
[This PR](https://github.com/runtimeverification/llvm-backend/pull/626) fixes an issue that would have been identified if we'd used IWYU; we should integrate it into our existing CI checks for C++ code.

Baltoli:
This PR audits the includes in each C++ source / header file to ensure that the correct ones are used (i.e. that no extra headers are pulled in, and that we don't rely transitively / implicitly on headers we haven't included).

Doing so addresses a failure mode I've seen a couple of times recently, where a change of C++ standard library versions (usually on Arch or macOS as they have rolling releases) breaks an assumption of transitive inclusion.

The changes to the backend's code made in this PR are large but entirely mechanical, and localised only to include blocks as suggested by IWYU. One exception to this is `scripts/check-includes`, which runs the tool with the right flags for you.

Fixes #627; I have opted _not_ to add this script to CI as it's a low priority check and the tool can be a bit temperamental to get working reliably. The fixes it suggests are likely to be slow-moving as well.

On Ubuntu, with a locally-built backend, you can check the results with:
```sh
apt install iwyu
scripts/check-includes
```


dwightguth (approved):
The fixes seem quite reasonable. Please make sure to test on every version of the C++ and C standard library that we happen to support if you haven't already, prior to merging. I'd also recommend double checking that this doesn't break anything for versions of LLVM we support that are different than the one on CI.

theo25 (approved):

Baltoli:
The Nix CI compiles on every LLVM version we support, and I believe I've tested the C++ libraries adequately.

PR #656, opened 2023-02-07
Rangemap hooks

mariaKt:
This PR implements the hooks for the rangemap type created [here](https://github.com/runtimeverification/llvm-backend/pull/620), and adds support for the rangemap in the llvm backend. Garbage collection for rangemap is not supported yet, and will be added in a followup PR.

I include a KORE test, that includes a definition with a few small range maps and the inputs use each hook a few or more times to test for corner cases. Other than the range map hooks, this also aims to test the ordering function (<). The ordering function (<) used was newly implemented, but was extracted and merged as a different PR. Therefore, it is not being exercised yet by another part of the backend.

There will be two followup PRs:
- A PR to add support for the rangemap in the frontend (already prepared).
- A PR to add GC for rangemap (next item to work on).

As part of working on the hooks, I made a few changes on the underlying data structures (range map and red-black tree). The changes in the data structures were mainly:
- I realized that due to the need for GC, there was a need for the data structures to be able to be altered in place, when needed (expose a non-const interface). A few changes due to that.
- Some additional functionality implemented as methods, needed for the hooks.
- I ended up needing an iterator for the hooks as well. Changes related to that.


bin/llvm-kompile-clang
```diff
@@ -195,7 +195,7 @@ else
 fi
 
 if $link; then
-  run @CMAKE_CXX_COMPILER@ -Wno-override-module -Wno-return-type-c-linkage "$modopt" "${files[@]}" \
+  run @CMAKE_CXX_COMPILER@ -std=c++17 -Wno-override-module -Wno-return-type-c-linkage "$modopt" "${files[@]}" \
```
Baltoli, line 198: https://github.com/runtimeverification/llvm-backend/pull/696 subsumes this change

include/runtime/collections/RBTree.h
```diff
@@ -140,14 +138,37 @@ class RBTree {
...
+    return r->data_;
+  }
+
+  /* WARNING: The following method returns a non-const reference in order to  *
+   * enable altering this data structure (and potentially ones implemented on *
+   * top of it, namely RangeMap) in place. The data structure is intended to  *
+   * be immutable: the exposed functionality should not alter the data        *
+   * structure's contents in place. Only request a non-const reference if you *
+   * in fact need to edit the data structure in place for a specific reason,  *
+   * e.g. garbage collection.                                                 */
```
Baltoli, line 165: Perhaps worth just expanding slightly on this point - I presume the restriction is that you can't use this to modify the contents in a way that affects the tree's internal invariants? e.g. the tree ordering isn't looking at the GC bits

runtime/collections/kelemle.cpp
```diff
@@ -0,0 +1,230 @@
...
+    return false;
+  }
+}
+
+// We arbitrarily impose the following ordering betwene different types of
+// blocks:
+// - Constants < Blocks.
+// - Symbols < Strings.
+// - Symbols with different tags are ordered based on their tags.
+// - Symbols with same tags are ordered be laxicoghraphically comparing their
```
Baltoli, line 149:
```suggestion
// - Symbols with same tags are ordered be lexicographically comparing their
```

Baltoli (commented):
Some very minor comments, but as it is this all looks good to me!

bin/llvm-kompile-clang
mariaKt: Yes, I will revert after it is merged.

include/runtime/collections/RBTree.h
mariaKt: Addressed.

bin/llvm-kompile-clang
mariaKt: It turned out that this flag is needed even after merging master (including PR [#696](https://github.com/runtimeverification/llvm-backend/pull/696). Removing this caused multiple (62) integration tests to fail, so I kept it.

mariaKt: Addressed.

include/runtime/header.h
```diff
@@ -130,6 +132,8 @@ void hash_exit(void);
 
 class KElem {
 public:
+  KElem() { this->elem = NULL; }
```
dwightguth, line 135: what is this constructor used for?

runtime/collections/rangemaps.cpp
```diff
@@ -0,0 +1,282 @@
...
+static struct blockheader range_header() {
+  static struct blockheader hdr = {(uint64_t)-1};
+  if (hdr.hdr == -1) {
+    hdr = getBlockHeaderForSymbol(
+        (uint64_t)getTagForSymbolName("LblRangemap'Coln'Range{}"));
+  }
+  return hdr;
+}
+rangemap hook_RANGEMAP_elementRng(SortKItem rng, SortKItem value) {
+  range *ptr = (range *)rng;
```
dwightguth, line 25: This doesn't seem quite right. You marked rng as being of sort KItem, but then you cast it to `range *`, when I'm pretty sure you also explicitly created a `Range` type. Either `rng` needs to be of sort `Range`, or else you're going to need to deconstruct an injection here. It depends on what the frontend code is expecting in terms of a type.

runtime/collections/rangemaps.cpp
```diff
@@ -0,0 +1,282 @@
...
+
+set hook_RANGEMAP_keys(SortRangeMap m) {
+  auto tmp = hook_SET_unit();
+  for (auto iter = rng_map::ConstRangeMapIterator<KElem, KElem>(*m);
+       iter.has_next(); ++iter) {
+    range *ptr = (range *)koreAlloc(sizeof(range));
+    ptr->h = range_header();
+    ptr->start = iter->first.start();
+    ptr->end = iter->first.end();
+    auto elem = hook_SET_element((SortRange)ptr);
```
dwightguth, line 113: this is also not quite right. a `Set` contains elements of sort `KItem`, but here you are putting elements of sort Range into it. You need to create an injection from Range into KItem to wrap around it.

runtime/collections/rangemaps.cpp
```diff
@@ -0,0 +1,282 @@
...
+
+list hook_RANGEMAP_keys_list(SortRangeMap m) {
+  auto tmp = list().transient();
+  for (auto iter = rng_map::ConstRangeMapIterator<KElem, KElem>(*m);
+       iter.has_next(); ++iter) {
+    range *ptr = (range *)koreAlloc(sizeof(range));
+    ptr->h = range_header();
+    ptr->start = iter->first.start();
+    ptr->end = iter->first.end();
+    tmp.push_back((SortRange)ptr);
```
dwightguth, line 127: same deal here

runtime/collections/rangemaps.cpp
```diff
@@ -0,0 +1,282 @@
...
+SortKItem hook_RANGEMAP_choiceRng(SortRangeMap m) {
+  if (m->empty()) {
+    KLLVM_HOOK_INVALID_ARGUMENT("Cannot choose from an empty range map");
+  }
+  auto pair = m->treemap().root_data();
+  range *ptr = (range *)koreAlloc(sizeof(range));
+  ptr->h = range_header();
+  ptr->start = pair.first.start();
+  ptr->end = pair.first.end();
+  return (SortRange)ptr;
```
dwightguth, line 162: here's another inconsistency. You declared the return type of this function as KItem, but you're returning a term of sort Range. Again, either an injection needs to be wrapped around the term, or else the return type needs to be changed to SortRange.

test/defn/k-files/rangemap-simple.k
```diff
@@ -0,0 +1,33 @@
...
+  imports COLLECTIONS
+  imports RANGEMAP
+  imports INT
+  imports JSON
+
+  syntax RangeMap ::= rangemap (Int) [function]
+// ---------------------------------------
+  rule rangemap(0) => .RangeMap
+
+  rule rangemap(1) => range(0, 20) r|-> 40
```
dwightguth, line 14: We should consider just using standard range syntax like `[0, 20)` instead of `range(0, 20)`

dwightguth (commented):
This is big, so I'm reviewing it in multiple sittings. I haven't looked at the red-black-tree code or the tests in detail yet. We might want to get someone other than myself with more familiarity with red-black trees to look at the former, honestly. I trust myself to notice if there are oddities in the code or the design of the API, but I don't trust myself to spot if there are problems with the implementation of the trees and their algorithms for various operations.

With the caveat that I haven't reviewed those sections yet, I do already have several comments. I marked some of them inline, but here are the more general comments:

* I noticed a lot of confusion between use of SortKItem and SortRange. While they are both typedef'd to `block *`, they represent different K sorts and terms that are valid instances of one are not valid instances of the other. It's important that we use each one correctly in correspondance with the K sorts that they correspond to when it comes to the signatures of the hooks, because this is what will allow GDB to pretty print those terms correctly. It's also important that we make sure to distinguish between the two because if we pass a term of sort Range somewhere a term of sort KItem is expected, or vice versa, this will corrupt the term representation and lead to segfaults and other types of bugs and crashes. I tried to mark everywhere I could spot that something was amiss, but I don't know that I spotted every single case. Please read through your code changes with an eye to whether or not these types are used correctly with regards to one another.
* I had a question: your implementation of the range maps is intended to preserve the invariant that they are in normal form, right? i.e., that all ranges that are able to be combined are always combined whenever a new collection is created? I noticed a couple places where the algorithms you wrote would not be correct if that invariant wasn't preserved, so I wanted to make sure that the code was deliberately written to be this way.

dwightguth (commented):
This is big, so I'm reviewing it in multiple sittings. I haven't looked at the red-black-tree code or the tests in detail yet. We might want to get someone other than myself with more familiarity with red-black trees to look at the former, honestly. I trust myself to notice if there are oddities in the code or the design of the API, but I don't trust myself to spot if there are problems with the implementation of the trees and their algorithms for various operations.

With the caveat that I haven't reviewed those sections yet, I do already have several comments. I marked some of them inline, but here are the more general comments:

* I noticed a lot of confusion between use of SortKItem and SortRange. While they are both typedef'd to `block *`, they represent different K sorts and terms that are valid instances of one are not valid instances of the other. It's important that we use each one correctly in correspondance with the K sorts that they correspond to when it comes to the signatures of the hooks, because this is what will allow GDB to pretty print those terms correctly. It's also important that we make sure to distinguish between the two because if we pass a term of sort Range somewhere a term of sort KItem is expected, or vice versa, this will corrupt the term representation and lead to segfaults and other types of bugs and crashes. I tried to mark everywhere I could spot that something was amiss, but I don't know that I spotted every single case. Please read through your code changes with an eye to whether or not these types are used correctly with regards to one another.
* I had a question: your implementation of the range maps is intended to preserve the invariant that they are in normal form, right? i.e., that all ranges that are able to be combined are always combined whenever a new collection is created? I noticed a couple places where the algorithms you wrote would not be correct if that invariant wasn't preserved, so I wanted to make sure that the code was deliberately written to be this way.

dwightguth:
Overall this looks excellent though! I'm very happy this is coming together and with the monumental effort you've put into this.

theo25:
Regarding the red-black tree data structure implementation, I have reviewed both the code and cited paper on which the implementation was based. The code seems correct to my best understanding.

mariaKt:
I looked through the uses of `SortKItem` and `SortRange` and pushed a commit that makes the necessary changes in the code and unit tests. The KORE test it will be soon.

Regarding your question, yes, the implementation of rangemaps preserves this invariant that all ranges that can be combined will be merged into one after an operation that creates a new rangemap is performed.

include/runtime/header.h
mariaKt: This constructor was added because I included the constructor `Range()` ([here](https://github.com/runtimeverification/llvm-backend/blob/5330cee8a1833256aa3e64d9b26a7a5900159194/include/runtime/collections/rangemap.h#L32)), that creates an empty `Range`. That was needed in the implementation of a function that needs to construct `Range`s from `KElem`s, [here](https://github.com/runtimeverification/llvm-backend/blob/5330cee8a1833256aa3e64d9b26a7a5900159194/include/runtime/collections/RBTree.h#L385).

runtime/collections/rangemaps.cpp
mariaKt: Addressed, using sort Range.

runtime/collections/rangemaps.cpp
mariaKt: Addressed, adding the missing injection to the set items.

runtime/collections/rangemaps.cpp
mariaKt: As above.

runtime/collections/rangemaps.cpp
mariaKt: Returning sort `Range`.

include/runtime/collections/rangemap.h
```diff
@@ -245,6 +265,41 @@ class RangeMap {
     }
   }
 
+  // Gather the intersection ranges of this rangemap and rangemap m, i.e. all
+  // ranges in this rangemap that overlap (fully or partially) with ranges in m
+  // and are mapped to the same value in both rangemaps, in res.
+  void
+  get_intersection_ranges(RangeMap const &m, std::vector<Range<T>> &res) const {
```
Baltoli, line 272: There might be usages outwith the files in this diff, but I only see the method being called as:
```cpp
vector<...> v;
get_intersection_ranges(m, v);
```

Could it be refactored to just return the result? If it's called somewhere such that that wouldn't work then no problem.

Baltoli (approved):
Approving as this all seems good to me on my first inspection, with the obvious caveat that it's a very big change and I could well have missed something! Nevertheless it's seriously good work to get a change like this implemented, so let's get it merged and we can deal with any issues arising if we need to. Congratulations @mariaKt!

include/runtime/collections/rangemap.h
mariaKt: Yes, addressed.

mariaKt:
As discussed in today's RVMatch standup, I will go ahead and merge this PR. @dwightguth , if you have any further feedback, please feel free to leave it here and I will go ahead and address it in a new PR.

PR #662, opened 2023-02-15
Cleanup old directory

Robertorosmaninho:
This simple PR deletes the `hello` directory and the `hello.h` from the `include/kllvm` include directory.
This folder and file were included at the begging of the project and today isn't used by or as a dependency for any other file, so it's safe to delete them.


Baltoli:
Closing as this PR is from a fork so CI won't run.

PR #665, opened 2023-02-20
Make python output relative

Baltoli:
Related to https://github.com/runtimeverification/k/issues/3165, https://github.com/runtimeverification/k/pull/3171

The Python bindings build process chooses an output filename by running the Python configuration module, rather than having one passed through the kompilation pipeline. This means that we need to special-case the handling of relative output paths in a couple of places when generating Python modules:
* Setting the output directory for Python modules rather than automatically dumping them in the CWD. This directory defaults to the parent of the `definition.kore` argument to `llvm-kompile`.
* Allowing this directory to be overridden when running the Nix test suite, which separates input definitions and the generated modules.

`llvm-kompile ... python[ast]` has not had its _default_ behaviour changed by this PR.

The existing test suite ensures that this is not a breaking change; the Python and non-Python code paths are both executed.

### DIFF
--- a/bin/llvm-kompile
+++ b/bin/llvm-kompile
@@ -186,7 +186,7 @@ if [[ "$OSTYPE" != "darwin"* ]]; then
 fi
 
 if [[ "${python_output_dir}" == "" ]]; then
-  python_output_dir="$(dirname ${definition})"
+  python_output_dir="$(dirname "${definition}")"
 fi
 kompile_clang_flags+=("--python-output-dir" "${python_output_dir}")
 
--- a/bin/llvm-kompile
+++ b/bin/llvm-kompile
@@ -8,13 +8,16 @@ Usage: $0 <definition.kore> <dt_dir> <object type> [options] [clang flags]
    or: $0 <interpreter.o>            <object type> [options] [clang flags]
 
 Options:
-  -h, --help      Print this message and exit.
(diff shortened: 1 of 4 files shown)


bin/llvm-kompile
```diff
@@ -93,6 +94,10 @@ while [[ $# -gt 0 ]]; do
       python_cmd="$2"
       shift; shift
       ;;
+    --python-output-dir)
```
tothtamas28, line 97: Currently, `llvm-kompile --help` does not list this option.

tothtamas28: `pyk` calls `llvm-kompile` as

```
llvm-kompile pythonast --python /usr/bin/python3 -- -o /tmp/tmpr2d3kbct/_kllvm.cpython-38-x86_64-linux-gnu.so
```

If I understand correctly, this change will simplify the above call to

```
llvm-kompile pythonast --python /usr/bin/python3 --python-output-dir /tmp/tmpr2d3kbct
```

tothtamas28 (approved):
Tested in `pyk`, everything seems to work fine.

bin/llvm-kompile
Baltoli: Yes, that's correct. The value of `--python` will get called to generate the filename using the same module that I believe you use in Pyk.

I'll add the help message and merge.

PR #669, opened 2023-02-23
Fix subtle bug in map/set choice algorithm

dwightguth:
When we wrote the code for the decision tree generation, we hadn't yet fully cleaned up the pattern matching algorithm, so we didn't enforce the invariant that map/set patterns with a map/set variable had to have at least one element. As a result, variable patterns in the code can have multiple representations: as a VariableP, or as a MapP or SetP with no elements and a variable. Unfortunately, we were not treating these two cases as functionally equivalent in all cases; in particular, if we do a map/set choice operation and the collection at runtime happens to be empty, we were discarding rows even if the pattern in that row ought to match the empty collection.

This code change ought to address that issue, but I have not fully tested it to determine if there are any further changes that will need to be made as a result of this change. I'm away from my computer at the moment, so I'm just kicking off some testing at this point. I will add an integration test to this PR before marking it as ready for review.


dwightguth:
I'm closing this as I am going to implement a different fix for this issue shortly.

PR #670, opened 2023-02-24
Fix compiler warnings

Baltoli:
This fixes two warnings-as-errors experienced by a Nethermind developer on:
> ubuntu 22.04.1 with clang12.0.1-19ubuntu3

I'm not sure why this doesn't happen in CI, but there is no harm in fixing the issues:
* Explicitly make a loop variable unsigned
* Add a default-case abort to a switch; we do this elsewhere to generate unreachable code, and error-handling elsewhere in the parser will catch bad tokens

### DIFF
--- a/lib/parser/KOREParser.cpp
+++ b/lib/parser/KOREParser.cpp
@@ -58,6 +58,7 @@ static std::string str(token tok) {
   case token::ID: return "<identifier>";
   case token::STRING: return "<string>";
   case token::TOKEN_EOF: return "<EOF>";
+  default: abort();
   }
 }
 
@@ -362,7 +363,7 @@ ptr<KOREPattern> KOREParser::applicationPattern(std::string name) {
     consume(token::RIGHTPAREN);
     if (name == "\\left-assoc") {
       ptr<KOREPattern> accum = std::move(pats[0]);
-      for (int i = 1; i < pats.size(); i++) {
+      for (auto i = 1u; i < pats.size(); i++) {
         auto newAccum = KORECompositePattern::Create(pat->getConstructor());
         newAccum->addArgument(std::move(accum));
         newAccum->addArgument(std::move(pats[i]));


anvacaru (approved):

Robertorosmaninho:
Hey @Baltoli, do you mind providing us a minimal example to reproduce this behavior?

Baltoli:
@Robertorosmaninho - @anvacaru reported the following issue:
```
/media/data/home/julek/work/nethermind/Synnax/Lamport/code/evm-semantics/deps/k/llvm-backend/src/main/native/llvm-backend/lib/parser/KOREParser.cpp:365:25: error: comparison of integer expressions of different signedness: 'int' and 'std::vector<std::unique_ptr<kllvm::KOREPattern, std::default_delete<kllvm::KOREPattern> > >::size_type' {aka 'long unsigned int'} [-Werror=sign-compare]
  365 |       for (int i = 1; i < pats.size(); i++) {
      |                       ~~^~~~~~~~~~~~~
/media/data/home/julek/work/nethermind/Synnax/Lamport/code/evm-semantics/deps/k/llvm-backend/src/main/native/llvm-backend/lib/parser/KOREParser.cpp: In function 'std::string kllvm::parser::str(kllvm::parser::token)':
/media/data/home/julek/work/nethermind/Synnax/Lamport/code/evm-semantics/deps/k/llvm-backend/src/main/native/llvm-backend/lib/parser/KOREParser.cpp:62:1: error: control reaches end of non-void function [-Werror=return-type]
   62 | }
      | ^
At global scope:
cc1plus: note: unrecognized command-line option '-Wno-return-type-c-linkage' may have been intended to silence earlier diagnostics
```

The changes to fix this are small and obvious enough that I am happy to fix them without reproducing precisely the problem; I'm not convinced tracking down the exact combination of OS + Clang + standard library that causes the warnings is worth the effort.

Robertorosmaninho:
Fair enough! Thanks! The modification LGTM!

PR #671, opened 2023-02-24
Fix bug involving non-canonicalized MapP, ListP, or SetP

dwightguth:
When we wrote the code for the decision tree generation, we hadn't yet fully cleaned up the pattern matching algorithm, so we didn't enforce the invariant that map/set patterns with a map/set variable had to have at least one element. As a result, variable patterns in the code could have multiple representations: as a VariableP, or as a MapP or SetP with no elements and a variable. Unfortunately, we were not treating these two cases as functionally equivalent in all cases; in particular, if we do a map/set choice operation and the collection at runtime happens to be empty, we were discarding rows even if the pattern in that row ought to match the empty collection.

This code change addresses this issue fully by making it so that the companion object of MapP, SetP, and ListP's `apply` method returns a `Pattern` rather than the corresponding class, and automatically canonicalizes patterns which are simply variables into their underlying VariableP or WildcardP. We then make the constructor private so all instantiations outside of the Pattern classes go through the apply method.

Hopefully this will not introduce any bugs since it ought to be a straightforward fix for something that was behaving inconsistently prior, but we'll see when we run the tests.


Baltoli (approved):

Baltoli:
Makes sense to me, and the tests seem to be fine.

PR #673, opened 2023-02-27
Fix uses of soon-to-be-deprecated code

Baltoli:


Baltoli:
Closing as this is a bit tricky to work on currently and I found a workaround for what I was trying to do.

Issue #490, opened 2022-02-24
[llvm] Tests for `--search`


Baltoli:
Currently, the backend test suite doesn't have any tests for `--search` - this means that failures might go unnoticed until the submodule update PR to K is tested.

We should implement some tests for `--search` in the backend to avoid this.

PR #674, opened 2023-02-27
Fix bug involving krun --search, map/set choice, and priority

### LINKED ISSUE #672: Subtle bug involving search, choice, and priorities
Consider the following K definition:

```
module TEST
  imports INT

  configuration <tasks> <task multiplicity="*" type="Map"> <id> 0 </id> <k> $PGM:K </k> </task> </tasks> <x> 0 </x>

  rule <k> 0 => 1 </k> [priority(25)]

  rule <x> 0 => 1 </x>

endmodule
```

If I were to run:

```
kompile test.k --enable-search
krun -cPGM=0 ---depth 1
```

I would expect to get only a single search result, representing the case when the rule with priority 25 was applied. However, the llvm backend incorrectly returns a second search result: the case where the second rule applies.

The reason this occurs represents another subtle issue with the interaction between the implementation of Map/Set choice and rule priorities, but in this case, the issue only occurs when using search to obtain all possible successor states; in the event that only a single successor is requested, the correct successor will always be applied.

The issue that occurs is that the decision tree is attempting to determine if the first rule can apply. In order to do this, it must enter a Map choice operation. Once it does this, it determines correctly that the first rule can apply. It then proceeds to the decision tree generated to determine if any other rules can apply. This decision tree, correctly, indicates that no other rules can apply, because the decision tree in this case is simply the `fail` leaf of the decision tree. However, due to the way that Map/Set choice is implemented, this does not immediately halt the matching operation; after all, we need to determine for certain that there isn't a different Map element we might have chosen that would also lead to a rule being applied. Thus, the Map/Set choice algorithm backtracks, returning to the decision tree node immediately after the Map/Set choice operation, and attempts to apply rules on the next element in the collection. This is fine thus far; after all, if a second `task` cell existed which also contained a `0` in the `k` cell, the first rule might also apply on that cell additionally.

However, since there is no other task cells present in the map, it then terminates the choice and attempts to determine whether or not  other rules might apply with lower priority. This is normally the behavior we want, if we had not already applied a rule with higher priority; after all, it's possible that no rules involving the choice applied at all, and therefore we would have to try and see whether lower priority rules apply instead. However, since a rule with higher priority has already applied, this is now incorrect behavior; we ought instead to discard all rows which have lower priority from consideration.

The problem, then, is that the node we ought to return to when the decision tree reaches `fail` after applying the first rule is actually a node which is not an ancestor of the failure node. It's a node that ought to look like this:

```
get next element of map from iterator
if next element exists:
  this part of the decision tree is correct
else:
  fail
```

Note that the else case is not necessarily a "fail". It could be that there are other rules with priority 25 that don't involve the choice that need to be tried still. However, we should not try any rule with priority greater than 25 in this case.

I think I know how to address this. We need a second type of MakeIterator node which is specific to search decision trees. It then has two children instead of one; one is the child that is entered initially and jumped to if no rule has applied. The other is the child that is jumped to if a rule has already been applied when you reach a `fail` node. The difference between these two nodes is that the first node, on the "no more elements" branch, contains all remaining rules with priority greater than or equal to the highest priority rule in the matrix. The second node is identical on the "more elements" branch, but on the "no more elements branch", contains only rules with priority equal to the highest priority rule in the matrix.

The idea during code generation is that we keep track of both basic block labels when we enter the choice, but if we apply a rule, then for each choice currently on the stack, we replace the basic block label we jump to with the second basic block label associated with that choice.

I ought to be able to get started implementing this tomorrow, but I'd like @Baltoli 's feedback on whether he thinks this sounds like the correct solution to implement, since it's rather complicated.

dwightguth:
Fixes #672 

This turned out to be somewhat simpler than I expected. When you do a choice, you automatically try all rules with the current priority level inside the choice on each element. Thus, the rules that are left after the choice exits are guaranteed to only be rules with higher priorities. Thus, if a rule applies, and then a choice exits after that rule applies, we simply immediately jump to the fail node of the decision tree. This means that we can implement this change just with a change to the decision tree code generation process and we don't have to modify how the decision trees are generated at all.


dwightguth:
The test for this feature is dependent on #490, which @Robertorosmaninho is working on, so this is not ready to be merged yet, but you ought still to be able to review the code fix.

lib/codegen/Decision.cpp
```diff
@@ -983,6 +996,15 @@ void makeStepFunction(
...
 
+  llvm::AllocaInst *HasSearchResults = nullptr;
+  if (search) {
+    HasSearchResults = new llvm::AllocaInst(
+        llvm::Type::getInt1Ty(module->getContext()), 0, "hasSearchResults",
+        block);
+    new llvm::StoreInst(
+        llvm::ConstantInt::getFalse(module->getContext()), HasSearchResults,
+        block);
+  }
```
Baltoli, line 1007: In the interests of keeping the backend documented moving forward, perhaps this code could include a comment explaining why we do this.

Baltoli (approved):

Baltoli:
Other than adding some comments, I'm happy with this change once @Robertorosmaninho has landed a comprehensive enough set of tests for `--search` that we can be fully confident this won't break.

The change itself seems safe to me given the ordering of rules we use in practice:
> @dwightguth: [...] we see that actually, the rules that are placed outside the choice consist only of rules with lower priority; all rules with the highest priority will still exist inside the choice and be tried on each element.

Baltoli (approved):

PR #678, opened 2023-03-01
Update `kprint` command line options implementation

### LINKED ISSUE #677: [kprint] Error invoking help flag
The `kprint --help` execution terminates with an error:

```
$ kprint --help
usage: kprint <definition.kore> <pattern.kore> [true|false|auto] [true|false]
terminate called after throwing an instance of 'std::logic_error'
  what():  basic_string::_M_construct null not valid
[1]    314988 IOT instruction (core dumped)  kprint --help
```

The expected behavior is:
```
$ kprint --help
usage: kprint <definition.kore> <pattern.kore> [true|false|auto] [true|false]
```

Robertorosmaninho:
Fixes #677 

This PR modifies the `kprint` implementation to use `llvm::cl` to parse the command line options correctly.


Baltoli:
Seems to be failing CI consistently; will review properly when that's sorted out.

tools/kprint/main.cpp
```diff
@@ -16,7 +16,7 @@ cl::opt<std::string> DefinitionFilename(
 cl::opt<std::string>
     PatternFilename(cl::Positional, cl::desc("<pattern.kore>"), cl::Required);
 
-cl::opt<bool> FilterSubst(cl::Positional, cl::desc("[true|false]"));
+cl::opt<std::string> FilterSubst(cl::Positional, cl::desc("[true|false]"));
```
Baltoli, line 19: This should be a `cl::opt<bool>`; doing so will save you from checking the string against the literal `"true"` later on.

tools/kprint/main.cpp
```diff
@@ -1,33 +1,36 @@
...
+
+cl::opt<std::string> DefinitionFilename(
+    cl::Positional, cl::desc("<definition.kore>"), cl::Required);
+
+cl::opt<std::string>
+    PatternFilename(cl::Positional, cl::desc("<pattern.kore>"), cl::Required);
+
+cl::opt<std::string> FilterSubst(cl::Positional, cl::desc("[true|false]"));
+
+cl::opt<std::string> ArgColor(cl::Positional, cl::desc("[true|false|auto]"));
```
Baltoli, line 21: Again rather than a string, let's make this a dedicated enum with three values.

tools/kprint/main.cpp
```diff
@@ -1,33 +1,36 @@
...
-  printKORE(std::cout, argv[1], argv[2], hasColor, filterSubst);
+  cl::PrintOptionValues();
+  StringMap<cl::Option *> &Map = cl::getRegisteredOptions();
+  Map["color"]->setHiddenFlag(cl::Hidden);
+
+  cl::ParseCommandLineOptions(argc, argv);
+
+  bool hasColor = ArgColor == "true" || (ArgColor == "auto" && isatty(1))
+                  || (ArgColor == "" && isatty(1));
+  bool filterSubst = FilterSubst == "true" || "" == FilterSubst;
```
Baltoli, line 32: Here you're checking against an implicit default value for both options; a slightly more ergonomic way to do it is to pass a `cl::init` parameter to the initializer for each option.

tools/kprint/main.cpp
```diff
@@ -1,33 +1,36 @@
...
-  }
-  if (argc >= 5) {
-    std::string arg = argv[4];
-    filterSubst = arg == "true";
-  } else {
-    filterSubst = true;
-  }
-
-  printKORE(std::cout, argv[1], argv[2], hasColor, filterSubst);
+  cl::PrintOptionValues();
```
Baltoli, line 24: What does this print / do?

tools/kprint/main.cpp
```diff
@@ -1,33 +1,36 @@
...
+  StringMap<cl::Option *> &Map = cl::getRegisteredOptions();
+  Map["color"]->setHiddenFlag(cl::Hidden);
```
Baltoli, line 26: _Very_ nitpicky, but I have a preference for almost-always-auto and `snake_case` for new code. Working out a set of C++ guidelines and style choices is on my list of things to do as part of the backend documentation updates.

tools/kprint/main.cpp
```diff
@@ -1,33 +1,36 @@
 #include "kllvm/ast/AST.h"
 #include "kllvm/printer/printer.h"
+#include "llvm/Support/CommandLine.h"
+#include "llvm/Support/WithColor.h"
```
Baltoli, line 4: What does this header pull in? I recommend running [IWYU](https://include-what-you-use.org/) over your code to make sure your includes are correct.

Baltoli (changes requested):
This seems broadly very good - some minor changes in the code, and two top-level comments:
* I think this change has quietly reversed the order of the colour and filtersubst parameters.
* It would be a useful change to document the additional parameters; your change keeps the documentation the same (modulo ordering), but we could display some extra information about the tool API.

tools/kprint/main.cpp
Robertorosmaninho: typo, I'll fix it!

tools/kprint/main.cpp
Robertorosmaninho: In this file, we have 2 conventions:
- upper camel case for `cl::opt` variables; and
- camelCase for the other variables

Should I modify all variables for the `snake_case`?

tools/kprint/main.cpp
Robertorosmaninho: I found out that it asks to include more than the necessary libraries. Should I include them?

```
../tools/kprint/main.cpp should add these lines:
#include <__functional/function.h>       // for __base
#include "llvm/ADT/StringMap.h"          // for StringMap
#include "llvm/ADT/StringRef.h"          // for StringRef
#include "llvm/Support/ErrorHandling.h"  // for llvm

../tools/kprint/main.cpp should remove these lines:
- #include <string>  // lines 7-7
- #include "llvm/Support/WithColor.h"  // lines 4-4

The full include-list for ../tools/kprint/main.cpp:
#include <__functional/function.h>       // for __base
#include <unistd.h>                      // for isatty
#include <iostream>                      // for string
#include "llvm/ADT/StringMap.h"          // for StringMap
#include "llvm/ADT/StringRef.h"          // for StringRef
#include "llvm/Support/CommandLine.h"    // for opt, desc, init, Positional
#include "llvm/Support/ErrorHandling.h"  // for llvm
```

tools/kprint/main.cpp
Robertorosmaninho: That's a good idea. I completely missed the cl::inti on llvm's documentation. Thanks for pointing that out!

tools/kprint/main.cpp
Baltoli: Good point - option variables I do prefer to keep as UpperCamelCase, but locals as snake_case

tools/kprint/main.cpp
Baltoli: It can be a bit finicky to get it to run properly, so take the suggestions with a pinch of salt - e.g. `<__functional/function.h>` probably isn't one we need. It's a sanity check rather than a prescriptive guide

tools/kprint/main.cpp
Robertorosmaninho: I don't think we can use `cl::Positional` with enums because `clEnumVal` and `clEnumValN` creates a list of "GeneralOptions" like:
```
USAGE: kprint [options] <definition.kore> <pattern.kore> Colorize output [true|false]

OPTIONS:

General options:

  Colorize output
      --true     - always colorize
      --false    - never colorize
      --auto     - colorize if stdout is a tty
```

This improves the documentation for this additional parameter. The downside is that the API is modified to use `--true` instead of `true`.

tools/kprint/main.cpp
Robertorosmaninho: Fair enough, I'll change it back!

tools/kprint/main.cpp
Baltoli: Sure, looking at the documentation that does indeed seem to be the case. I see two options:
* Specialize `llvm::cl::parser` for your enum
* Validate by hand that the option is in `{"true", "false", "auto"}`.

Either of these is probably fine; weak preference for the first to get LLVM's parse error handling, but if the code is problematic then don't worry.

Baltoli: https://llvm.org/docs/CommandLine.html#writing-a-custom-parser ah, you don't get parse errors for free. Just validate the string in that case

Baltoli (approved):

PR #680, opened 2023-03-03
Fix llvm build on Arch

Baltoli:
We were relying on a [somewhat outdated CMake command](https://bugzilla.opensuse.org/show_bug.cgi?id=1096168#c2) to identify the LLVM libraries we need to link against.

This PR replaces uses of `map_llvm_components_to_libnames` with the more up-to-date `llvm_config` macro.


theo25 (approved):
Looks good to me

PR #685, opened 2023-03-06
Creating new test to `--search`

### LINKED ISSUE #490: [llvm] Tests for `--search`

Robertorosmaninho:
Fixes #490

The `search.k` test introduced in this PR is the same one used on the [haskel-backend](https://github.com/runtimeverification/haskell-backend/blob/master/test/search/search.k) and has the following definition:

```k
module SEARCH

  imports BOOL

  syntax S
    ::= "initial"
      | "next1"  | "next2"
      | "final1" | "final2"
      | "unreachable"

  rule initial => next1

  rule initial => next2

  rule next1 => final1

  rule next2 => final2

endmodule
```


Baltoli (approved):
Looks good to me!

dwightguth (approved):

PR #688, opened 2023-03-07
Improve hiding of LLVM options

Baltoli:
This PR fixes two issues with one stone:
* For some reason, accessing the registered options to hide `--color` in `kprint` causes a segfault when running the backend on arm64/macOS Nix, when it's built as part of K. I noticed because it was breaking the submodule update, but I can reproduce this locally.
* We get a lot of LLVM options from `--help` for the LLVM backend tools; we can give our options an explicit category and hide the rest.

I have verified locally that these changes will allow the K submodule update to go through.

### DIFF
--- a/tools/kore-arity/main.cpp
+++ b/tools/kore-arity/main.cpp
@@ -10,9 +10,13 @@
 
 using namespace llvm;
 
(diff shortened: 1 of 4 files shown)


Robertorosmaninho (approved):
LGTM!

Issue #667, opened 2023-02-22
Support for `[functional]` to `[total]` migration


Baltoli:
Multi-stage deprecation

ehildenb:
Might be the case that the llvm backend largely ignores functional/total anyway.

ehildenb:
One complication: the LLVM backend CI integration tests use raw kore files committed directly to repo.

- Could change the kore files directly (maybe will invalidate digests in those files though?)
- Could regenerate the kore files if needed (hard).

PR #689, opened 2023-03-07
Add support to `total` like `functional`

Robertorosmaninho:
Fixes: #667 

Create support for the `[total]` attribute to replace `[functional]` in the future.

### DIFF
--- a/lib/ast/AST.cpp
+++ b/lib/ast/AST.cpp
@@ -1234,9 +1234,8 @@ bool KOREAxiomDeclaration::isRequired() {
   return !attributes.count(ASSOC) && !attributes.count(COMM)
          && !attributes.count(IDEM) && !attributes.count(UNIT)
          && !attributes.count(FUNCTIONAL) && !attributes.count(CONSTRUCTOR)
-         && !attributes.count(TOTAL) && !attributes.count(CONSTRUCTOR)
-         && !attributes.count(SUBSORT) && !attributes.count(CEIL)
-         && !attributes.count(NON_EXECUTABLE)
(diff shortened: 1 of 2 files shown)


dwightguth (approved):

radumereuta:
How did you test if it works?
You can locally change the front-end: https://github.com/runtimeverification/k/blob/master/kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java#L1486
and it will print `total` instead of `functional`. If all tests from `regression-new` passes, then it should be fine.

lib/ast/AST.cpp
```diff
@@ -1233,6 +1234,7 @@ bool KOREAxiomDeclaration::isRequired() {
   return !attributes.count(ASSOC) && !attributes.count(COMM)
          && !attributes.count(IDEM) && !attributes.count(UNIT)
          && !attributes.count(FUNCTIONAL) && !attributes.count(CONSTRUCTOR)
+         && !attributes.count(TOTAL) && !attributes.count(CONSTRUCTOR)
```
Baltoli, line 1237: This duplicates the Constructor check.

Baltoli (approved):
One minor comment; otherwise looks good to me if the frontend tests pass.

Robertorosmaninho:
> How did you test if it works?
> You can locally change the front-end: https://github.com/runtimeverification/k/blob/master/kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java#L1486
> and it will print `total` instead of `functional`. If all tests from `regression-new` passes, then it should be fine.

Good idea! Thanks, Radu! Changing this line, I got an error on test `issue-1760` without my modification, and the error was solved within it!

Issue #693, opened 2023-03-08
Output KORE from `kprint`


Baltoli:
`kprint` applies some normalization passes to the AST it operates on before printing. @amelieled requests that we allow KORE output after these passes, as well as pretty-printing, as a feature for the Dedukti backend.

PR #694, opened 2023-03-08
don't assign special priorities to cooling rules

dwightguth:
The first commit is a cherry pick of #674 which I had to revert because it was causing test failures. The second commit is a fix for the test failures. The third commit adds the failing test to the backend test suite.

For context, this is the issue: The llvm backend reserves priorities between 100 and 150 inclusive for internal use to help optimize straight-line execution. As a result, we were assigning certain rules priorities in this range which were not explicitly given any unusual priority by the user. Because of the bug that was fixed in #674, we were still getting the correct set of solutions, however. When that bug was fixed, we stopped getting some of the solutions because they involved a sequence of rewrites that required that the successor states of a state include a state where a rule that was assigned such a priority applied in addition to lower priority rules. In other words, we were now correctly applying the priorities assigned to rules by the pattern matching compiler, but those priorities were not, strictly speaking, sound with respect to the semantic meaning of priority due to the optimizations the llvm backend was trying to make by assigning priorities in this range.

The solution is straightforward; assign rules the priorities that they, semantically speaking, have, when constructing the matrix to be used for search.


dwightguth:
I had to remove the test from this PR because it was failing on Nix. The reason, as far as I can tell, is that Nix was compiled just enough differently that it causes a different ordering of elements in a commutative collection. The test in question was taken directly from the K frontend, so it ought to be able to be added back in later very easily. However, that would depend on #693 .

dwightguth:
I'd like to try to get this merged now because it is blocking upstream issues, and I don't want to delay those issues being fixed simply due to the inability to write the proper tests. We will still ultimately be testing this functionality prior to releasing any version of K, due to its presence in the frontend test suite, so I think it's safe to wait until later to add the test to the backend test suite.

Baltoli (approved):

PR #696, opened 2023-03-09
Enforce C++17 in llvm-kompile-clang

### LINKED ISSUE #695: C++17 enforcement for `llvm-kompile-clang`
@mariaKt's ongoing work to implement a hooked rangemap type for the LLVM backend requires C++17 support. We require C++17 when building the backend already, but the `search` entrypoint exposed by `llvm-kompile` includes the runtime headers in a C++ file that is built at `kompile` time. We also require 17 when building the Python bindings.

We should require C++17 consistently at the top level of `llvm-kompile-clang`.

Baltoli:
Fixes #695 by making C++17 a top-level requirement in `llvm-kompile-clang`; we already depended on it implicitly in a number of places (building the backend, Python bindings), so this change shouldn't be a departure in practice.


dwightguth (approved):
This looks good. Is ci testing on llvm 12? If so, this should be fine to merge.

Baltoli:
Yes, the CI matrix runs against LLVM 12.

PR #698, opened 2023-03-10
Initial K LLDB support

Baltoli:
This PR implements initial LLDB support for the following parts of the K debugging interface:
* Pretty-printing of runtime K terms
* Match debugging and introspection (`k match $rule $term`).

The `k start` and `k step` commands are **not** implemented; `lldb` crashes when running the required commands non-interactively. I am in the process of minimising an example[^1] to submit as a question to the maintainers to make sure I'm using the API correctly. Because these commands are thin wrappers over traditional debugger functionality, I have left them unimplemented with a message suggesting how best to emulate their behaviour.

We do not currently have any tests for the debugger interfaces; I have followed through the K GDB tutorial by hand using the LLDB version, and have verified that everything works as expected.

I will follow this PR up with a frontend PR that updates the documentation and tutorial appropriately.

[^1]: I'm pretty confident this is me misusing the API somehow, rather than a re-entrancy bug as @dwightguth mentioned GDB experienced, as it manifests even without calling functions in the inferior process.


include/runtime/header.h
```diff
@@ -207,6 +210,14 @@ void printSortedConfigurationToFile(
     FILE *file, block *subject, char const *sort);
 void printConfigurationInternal(
     writer *file, block *subject, const char *sort, bool, void *);
+
+// This function injects its argument into KItem before printing, using the sort
+// argument as the source sort. Doing so allows the term to be pretty-printed
+// using the existing recursion scheme code (and without manually inspecting the
+// sort to see what printer we need to call if the term isn't an entire
+// configuration).
+string *debug_print_term(block *subject, char const *sort);
```
dwightguth, line 219: Is there a reason besides convenience we did it this way? I would tend to think we should try to handle it in a more principled manner if we can, because we debug k programs often enough in contexts where we want to see the exact kore that the term represents, and this technically alters it, which will be visible to the user if they disable the pretty printer for kore.

Baltoli: Yes, I agree that a more principled way would be preferable. The problem is roughly as follows:

The [existing configuration printer entrypoint](https://github.com/runtimeverification/llvm-backend/blob/5bcdedd51a7874afcb19af5340b4af85d0c37830/runtime/util/ConfigurationPrinter.cpp#L114-L116) assumes that its argument is a block (and we can therefore inspect the pointer bits). The generated-code recursion scheme handles calling the `print$SORT` family of functions safely once we know the sort of a term _below the top level_.

We could dispatch over the known sort of the term roughly like this to get similar behaviour:
```c++
if (sort_str == "SortInt{}") printInt(...)
else if (sort_str == "SortBool{}") printBool(...)
...
```

The issue I came to is that `printList` (c.f. sets, maps) require that the collection symbols (unit, element, concat) are known if you have a term of `SortList{}`. We could hard-code `Lbl'Unds'List'Unds` etc., but I wasn't sure if this is potentially unsound given that the recursion code has to[ look up the symbols in the definition](https://github.com/runtimeverification/llvm-backend/blob/79c01a60fa1bf3219a189e23fe3003e7ed7313c4/lib/codegen/EmitConfigParser.cpp#L887-L890). I note that this is actually what the GDB code does, but in the interests of keeping the code small I opted for the "inject into KItem" possible unsoundness rather than the "guess the collection symbols" one.

Let me know your thoughts here - happy to change the approach to whatever is the most sensible.

dwightguth: Hmmm. I guess this is a reasonable tradeoff. Let's go ahead and keep this as is for now.

PR #703, opened 2023-03-15
Add support to `--binary-output` on search interpreter

### LINKED ISSUE #690: search --binary-output
The search binary main method doesn't support the --binary-output flag. We probably should support this, but it requires modifications made to add support for serializing a disjunction via binary, something which we haven't written code to do yet.

Robertorosmaninho:
Fixes #690 

This PR enables the interpreter created with search support to print the file using the KORE binary.

### DIFF
--- a/runtime/util/ConfigurationSerializer.cpp
+++ b/runtime/util/ConfigurationSerializer.cpp
@@ -335,7 +335,7 @@ void serializeConfigurations(
   auto state = serialization_state();
 
   writer w = {file, nullptr};
-  ssize_t size = results.size();
+  auto size = results.size();
   if (size == 0) {
     emitConstantSort(state.instance, "SortGeneratedTopCell");
     emitSymbol(state.instance, "\\bottom{}", size, 1);
@@ -343,12 +343,12 @@ void serializeConfigurations(
     auto result = *results.begin();
     serializeConfigurationInternal(&w, result, nullptr, false, &state);
   } else {
-    for (const auto &subject : results) {
+    for (auto const&subject : results) {
       serializeConfigurationInternal(&w, subject, nullptr, false, &state);
     }
 
     emitConstantSort(state.instance, "SortGeneratedTopCell");
-    emitSymbol(state.instance, "\\or{}", results.size(), 1);
+    emitSymbol(state.instance, "\\or{}", size, 1);
   }
 
   auto buf_size = state.instance.data().size();
--- a/runtime/util/ConfigurationSerializer.cpp
+++ b/runtime/util/ConfigurationSerializer.cpp
@@ -339,6 +339,9 @@ void serializeConfigurations(
   if (size == 0) {
     emitConstantSort(state.instance, "SortGeneratedTopCell");
     emitSymbol(state.instance, "\\bottom{}", size, 1);
+  } if (size == 1) {
+    auto result = *results.begin();
+    serializeConfigurationInternal(&w, result, nullptr, false, &state);
   } else {
     for (const auto &subject : results) {
       serializeConfigurationInternal(&w, subject, nullptr, false, &state);
--- a/runtime/util/ConfigurationSerializer.cpp
+++ b/runtime/util/ConfigurationSerializer.cpp
@@ -6,6 +6,7 @@
 
 #include <cassert>
 #include <cstdio>
+#include <iostream>
 #include <limits>
 #include <set>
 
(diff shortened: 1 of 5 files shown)


runtime/util/ConfigurationSerializer.cpp
```diff
@@ -327,6 +329,33 @@ void serializeConfigurationInternal(
   }
 }
 
+void serializeConfigurations(
+    const char *filename, std::unordered_set<block *, HashBlock, KEq> results) {
+  FILE *file = fopen(filename, "w");
+  auto state = serialization_state();
+
+  writer w = {file, nullptr};
+  ssize_t size = results.size();
```
Baltoli, line 338: Minor style nitpick: `auto`

runtime/util/ConfigurationSerializer.cpp
```diff
@@ -327,6 +329,33 @@ void serializeConfigurationInternal(
...
+  if (size == 0) {
+    emitConstantSort(state.instance, "SortGeneratedTopCell");
+    emitSymbol(state.instance, "\\bottom{}", size, 1);
+  } else {
+    for (const auto &subject : results) {
```
Baltoli, line 343: Minor style nitpick: `auto const&`

runtime/util/ConfigurationSerializer.cpp
```diff
@@ -327,6 +329,33 @@ void serializeConfigurationInternal(
...
+      serializeConfigurationInternal(&w, subject, nullptr, false, &state);
+    }
+
+    emitConstantSort(state.instance, "SortGeneratedTopCell");
+    emitSymbol(state.instance, "\\or{}", results.size(), 1);
```
Baltoli, line 348:
```suggestion
    emitSymbol(state.instance, "\\or{}", size, 1);
```

Baltoli (commented):
Some very minor style comments; otherwise looks great

runtime/util/ConfigurationSerializer.cpp
Robertorosmaninho: writer can't be inferred by `auto`

```
 error: deduced conflicting types ('_IO_FILE *' vs 'nullptr_t') for initializer list element type
 auto w = {file, nullptr};

Comment (Baltoli) [reply runtime/util/ConfigurationSerializer.cpp]: `auto w = writer{file, nullptr}` in these situations.

Comment (Robertorosmaninho) [reply runtime/util/ConfigurationSerializer.cpp]: Done! Thanks!

File: runtime/util/ConfigurationSerializer.cpp [hunk 4]
```diff
@@ -327,6 +329,36 @@ void serializeConfigurationInternal(
...
+  auto state = serialization_state();
+
+  auto w = writer{file, nullptr};
+  auto size = results.size();
+  if (size == 0) {
+    emitConstantSort(state.instance, "SortGeneratedTopCell");
+    emitSymbol(state.instance, "\\bottom{}", size, 1);
+  } if (size == 1) {
+    auto result = *results.begin();
+    serializeConfigurationInternal(&w, result, nullptr, false, &state);
```
Baltoli, line 344:
```suggestion
    serializeConfigurationInternal(&w, results[0], nullptr, false, &state);
```

runtime/util/ConfigurationSerializer.cpp
```diff
@@ -327,6 +329,36 @@ void serializeConfigurationInternal(
...
+  } else {
+    for (auto const&subject : results) {
```
Baltoli, line 346: This will fail CI; there seems to be a github outage affecting the action, but you should make sure to run clang-format locally on your code.

Baltoli (changes requested):

runtime/util/ConfigurationSerializer.cpp
Robertorosmaninho: I don't think that it's possible to get the element by index on a `std::unordered_set`.

runtime/util/ConfigurationSerializer.cpp
Robertorosmaninho: You're right! My bad!

runtime/util/ConfigurationSerializer.cpp
Baltoli: Of course - my bad!

test/lit.cfg.py
```diff
@@ -100,6 +100,16 @@ def one_line(s):
...
+    
+    ('%check-search-convert-test', one_line('''
+        for out in %test-dir-out/*.out.diff; do
+            in=%test-dir-in/`basename $out .out.diff`.in
+            %t.interpreter $in 3 /dev/stdout | diff - $out
+            %kore-convert $in -o %t.bin && %t.interpreter %t.bin 3 /dev/stdout | diff - $out
+            %t.interpreter $in 3 %t.bin --binary-output
+            %kore-convert %t.bin -o %t.kore && %kore-convert $out --to=text | diff - %t.kore
+        done
+    ''')),
```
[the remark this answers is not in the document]
Baltoli: Yes, but this particular substitution hard-codes precisely 3 rewrite steps into the test. I would prefer to keep the substitutions in the test suite _general_ (and have unusual test cases, that for example don't rewrite fully, have longer test headers), rather than only being applicable to the single test case that you use it in.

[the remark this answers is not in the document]
Robertorosmaninho: Ok!

Baltoli (approved):

Baltoli:
Looks good - nice work @Robertorosmaninho!

PR #704, opened 2023-03-15
Output processed KORE from `kprint`

Baltoli:
Fixes https://github.com/runtimeverification/llvm-backend/issues/693; this issue is a feature request to get access to the post-processed KORE that `kprint` currently uses to produce pretty surface syntax.


Robertorosmaninho (approved):
LGTM!

PR #706, opened 2023-03-17
Duplicate dependabot PR #687

Baltoli:
#687 is the original PR; this is a duplicate to get CI to run properly. @dwightguth this seems to pass CI happily, but if there's a reason to stick with the older YAML parser version that's fine by me and we can close both PRs.


dwightguth (approved):
I'm approving so as not to block you, but please test this before merging.

Baltoli:
Size of the entire contents of `dt` from the C semantics for both versions of Snakeyaml:

| 1.32 | 2.0 |
| ------- | ---- |
| 62826702 |62830835|

I'm happy to merge this given that it doesn't seem to be a major increase.

PR #708, opened 2023-03-17
Introducing `k-rule-find` tool

### LINKED ISSUE #702: Implement `k-rule-find`
Per https://github.com/runtimeverification/k/issues/3180, we want to expose "why doesn't this rule apply" as a standalone feature, rather than something that only GDB / LLDB have access to.

We should implement a tool that consumes the LLVM backend's match log to expose this at the command line.

It might also be possible to expose a single function that GDB / LLDB can call into and get match failures printed, rather than having the Python code re-implement the logic.

Robertorosmaninho:
Fixes #702


tools/k-rule-find/main.cpp
```diff
@@ -0,0 +1,84 @@
...
+  if (!handle) {
+    std::cerr << "Error: " << dlerror() << "\n";
+    return EXIT_FAILURE;
+  }
+
+  // Parse the given KTerm and get the block* to use as input for the match
+  // function.
+  kllvm::parser::KOREParser parser(KTermFilename);
+  auto InitialConfiguration = parser.pattern();
+  //auto b = (block *)constructInitialConfiguration(InitialConfiguration.get());
```
Baltoli, line 71: This won't work as written. You'll need to dlopen the shared library to get a pointer to `constructInitialConfiguration`.

tools/k-rule-find/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
...
+  PUBLIC util BinaryKore Codegen Parser AST gmp mpfr numeric_strings
+)
+
+install(
+  TARGETS k-rule-find
+  RUNTIME DESTINATION bin
+)
+
+separate_arguments(LLVM_DEFINITIONS_LIST NATIVE_COMMAND ${LLVM_DEFINITIONS})
+add_definitions(${LLVM_DEFINITIONS_LIST})
```
Baltoli, line 15: What are these for? Are we missing preprocessor definitions that we should be using elsewhere?

tools/k-rule-find/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
+kllvm_add_tool(k-rule-find
+  main.cpp
+)
+
+target_link_libraries(k-rule-find 
+  PUBLIC util BinaryKore Codegen Parser AST gmp mpfr numeric_strings
```
Baltoli, line 6: Why do you need to link this tool against `gmp` etc.?

tools/k-rule-find/main.cpp
```diff
@@ -0,0 +1,84 @@
...
+  block *b = nullptr;
+
+  // Get the llvm match function pointer and cast it to right type.
+  void *llvm_function_ptr = dlsym(handle, "llvm_match_function");
+  auto match_function
+      = reinterpret_cast<void (*)(block *, std::string, int64_t, int64_t)>(
```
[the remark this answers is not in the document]
Baltoli: Because you'll be generating the implementation of the function as LLVM.

tools/k-rule-find/main.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+  // Parse the given KTerm and get the block* to use as input for the match
+  // function.
+  kllvm::parser::KOREParser parser(KTermFilename);
+  auto InitialConfiguration = parser.pattern();
+  //auto b = (block *)constructInitialConfiguration(InitialConfiguration.get());
+  block *b = nullptr;
+
+  // Get the llvm match function pointer and cast it to right type.
+  void *llvm_function_ptr = dlsym(handle, "llvm_match_function");
```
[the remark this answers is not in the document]
Robertorosmaninho: Done!

tools/k-rule-find/main.cpp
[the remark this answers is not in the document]
Robertorosmaninho: Yeah, but I will not return it in this variable, right?

Shouldn't this function have this signature?
`llvm_match_function(KORE_pattern_block, kdefinition_filename, line, column);`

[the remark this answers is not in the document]
Baltoli: Yes, that's the correct signature abstractly. The problem comes when you need to emit the LLVM code that inspects the filename data, though. How are you imagining that this function will be implemented taking the `std::string` argument?

tools/k-rule-find/main.cpp
```diff
@@ -0,0 +1,121 @@
...
+using namespace kllvm;
+
+struct Location {
+  char *filename;
+  int64_t line;
+  int64_t column;
+};
+
+std::string SOURCE_ATT = "org'Stop'kframework'Stop'attributes'Stop'Source";
+std::string LOCATION_ATT = "org'Stop'kframework'Stop'attributes'Stop'Location";
```
Baltoli, line 21: Let's pull these out into their own module somewhere now that we're using them in two places.

tools/k-rule-find/main.cpp
```diff
@@ -0,0 +1,121 @@
...
+#include <iostream>
+#include <stdlib.h>
+
+#include "runtime/header.h"
+
+using namespace llvm;
+using namespace kllvm;
+
+struct Location {
+  char *filename;
```
Baltoli, line 15: Don't store the raw `char *` in the struct - see comment below. This is a use-after-free waiting to happen!

tools/k-rule-find/main.cpp
```diff
@@ -0,0 +1,121 @@
...
+Location parseLocation(std::string loc) {
+  char *filename;
+  int64_t line, column = -1;
+  size_t pos = loc.find(":");
+  if (pos == std::string::npos) {
+    std::cerr
+        << "Rule's location must me in the format: defintion.k:line[:column]\n";
+    exit(EXIT_FAILURE);
+  }
+  filename = (char *)loc.substr(0, pos).c_str();
```
Baltoli, line 42: You're leaking a temporary here - `loc.substr` returns a new `std::string`, and its `c_str` isn't guaranteed to live after the temporary is constructed. Just store the `std::string` and only turn it into a `char *` at the very last point you need to do so.

tools/k-rule-find/main.cpp
```diff
@@ -0,0 +1,121 @@
...
+cl::opt<std::string> KompiledDir(
+    cl::Positional, cl::desc("<kompiled-dir>"), cl::Required,
+    cl::cat(KRuleCat));
+
+cl::opt<std::string> RuleLocation(
+    cl::Positional, cl::desc("<filename.k:line[:column]>"), cl::Required,
+    cl::cat(KRuleCat));
+
+Location parseLocation(std::string loc) {
+  char *filename;
```
Baltoli, line 34: Why does this need to be declared first?

tools/k-rule-find/main.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+Location getLocation(KOREAxiomDeclaration *axiom) {
+  auto *locationAtt = axiom->getAttributes().at(LOCATION_ATT).get();
+  assert(locationAtt->getArguments().size() == 1);
+
+  auto strPattern
+      = dynamic_cast<KOREStringPattern *>(locationAtt->getArguments()[0].get());
+
+  std::string location = strPattern->getContents();
+  auto filename = (char *)location.c_str();
```
Baltoli, line 75: Use after free again here!

tools/k-rule-find/main.cpp
```diff
@@ -0,0 +1,121 @@
...
+  // Parse the definition.kore to get the AST.
+  kllvm::parser::KOREParser parser(definition);
+  auto kore_ast = parser.definition();
+
+  // Iterate through axioms.
+  for (auto axiom : kore_ast.get()->getAxioms()) {
+    if (axiom->getAttributes().count(SOURCE_ATT)) {
+      std::string source = getSource(axiom);
+      if (source.find(source_filename) != std::string::npos) {
+        Location source_loc = getLocation(axiom);
```
Baltoli, line 104: Generally throughout this change AAA is preferable

tools/k-rule-find/main.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+  // Parse the definition.kore to get the AST.
+  kllvm::parser::KOREParser parser(definition);
+  auto kore_ast = parser.definition();
+
+  // Iterate through axioms.
+  for (auto axiom : kore_ast.get()->getAxioms()) {
+    if (axiom->getAttributes().count(SOURCE_ATT)) {
+      std::string source = getSource(axiom);
+      if (source.find(source_filename) != std::string::npos) {
```
Baltoli, line 103: I think you probably want slightly different semantics here; a suffix match rather than general contents, though I don't think it's likely to break things in practice.

tools/k-rule-find/main.cpp
```diff
@@ -0,0 +1,121 @@
...
+        Location source_loc = getLocation(axiom);
+        if (loc.line == source_loc.line
+            && (loc.column == -1 || loc.column == source_loc.column)) {
```
Baltoli, line 106: This isn't quite right - you're looking to see if the rule's location _contains_ the location passed at the command line. As implemented, it's implementing an exact match. The motivation as Radu put it is to be able to drop the cursor in the middle of some rule and have the tool identify its label.

tools/k-rule-find/main.cpp
```diff
@@ -0,0 +1,121 @@
...
+std::string getSource(KOREAxiomDeclaration *axiom) {
+  auto *sourceAtt = axiom->getAttributes().at(SOURCE_ATT).get();
+  assert(sourceAtt->getArguments().size() == 1);
+
+  auto strPattern
+      = dynamic_cast<KOREStringPattern *>(sourceAtt->getArguments()[0].get());
+  return strPattern->getContents();
+}
+
+Location getLocation(KOREAxiomDeclaration *axiom) {
```
Baltoli, line 67: You will need two location types - see comment below, but remember that the locations in the definition are _ranges_. You need to extract their start and end lines / columns. The format is:
```
Location(start_line, start_col, end_line, end_col)
```

tools/k-rule-find/main.cpp
```diff
@@ -0,0 +1,121 @@
...
+  return {filename, lineNumber, columnNumber};
+}
+
+int main(int argc, char **argv) {
+  cl::HideUnrelatedOptions({&KRuleCat});
+  cl::ParseCommandLineOptions(argc, argv);
+
+  auto loc = parseLocation(RuleLocation);
+  auto definition = KompiledDir + "/definition.kore";
+  auto source_filename = realpath(loc.filename, NULL);
```
Baltoli, line 92: You don't necessarily want to find the absolute path here; a suffix match should be enough.

tools/k-rule-find/main.cpp
```diff
@@ -0,0 +1,121 @@
+#include "kllvm/ast/AST.h"
+#include "kllvm/parser/KOREParser.h"
+#include "llvm/Support/CommandLine.h"
+
+#include <dlfcn.h>
+#include <iostream>
+#include <stdlib.h>
```
Baltoli, line 7:
```suggestion
#include <cstdlib>
```

tools/k-rule-find/main.cpp
```diff
@@ -0,0 +1,121 @@
+#include "kllvm/ast/AST.h"
+#include "kllvm/parser/KOREParser.h"
+#include "llvm/Support/CommandLine.h"
```
Baltoli, line 3: sorting and splitting includes:

```suggestion
#include <kllvm/ast/AST.h>
#include <kllvm/parser/KOREParser.h>

#include <llvm/Support/CommandLine.h>
```

Baltoli (changes requested):

tools/k-rule-find/main.cpp
Robertorosmaninho: No, do you want it to be declared after `getSource` and `getLocation`?

tools/k-rule-find/main.cpp
Robertorosmaninho: I don't think I get what you mean. Could you provide me with a further explanation of the problem here?

tools/k-rule-find/main.cpp
Robertorosmaninho: Would you prefer a second Location type instead of extending the current one with `start_line`, `end_line`, etc.?
In this case, the input location will have `start_line == end_line`

tools/k-rule-find/main.cpp
Baltoli: I mean the variable `filename` - in general leaving things uninitialised isn't best practice if you can declare + initialise at the same time.

tools/k-rule-find/main.cpp
Baltoli: Yeah, that makes sense as an approach. The check you're implementing can then be a general one for whether one location range contains another

tools/k-rule-find/main.cpp
Robertorosmaninho: Aah, okok!

tools/k-rule-find/main.cpp
Baltoli: You don't necessarily know the absolute path of the K source file on disk that the definition was compiled from. For example, the `kbuild` tool moves things into a temporary build directory rather than compiling from your cwd.

This means that you want to see if the command line argument is a _suffix_ of the source attribute, rather than if the realpath of the command line argument is an exact match.

tools/k-rule-find/main.cpp
```diff
@@ -0,0 +1,133 @@
+#include <kllvm/ast/AST.h>
+#include <kllvm/codegen/Debug.h>
+#include <kllvm/parser/KOREParser.h>
+#include <runtime/header.h>
```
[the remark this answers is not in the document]
Robertorosmaninho: Don't worry at all! Actually, it's not being used in this version of `k-rule-find`. Thanks for pointing that out!

include/kllvm/codegen/Debug.h
```diff
@@ -19,6 +19,11 @@ namespace kllvm {
 
 extern int CODEGEN_DEBUG;
 
+static std::string SOURCE_ATT
```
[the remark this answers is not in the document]
Robertorosmaninho: Okok

tools/k-rule-find/main.cpp
```diff
@@ -0,0 +1,132 @@
...
+  int64_t end_column = std::stoi(location.substr(third_comma + 1, length));
+
+  return {location, start_line, end_line, start_column, end_column};
+}
+
+Location parseLocation(std::string loc) {
+  size_t pos = loc.find(":");
+  if (pos == std::string::npos) {
+    std::cerr
+        << "Rule's location must me in the format: defintion.k:line[:column]\n";
```
Baltoli, line 72:
```suggestion
        << "Rule's location must be in the format: definition.k:line[:column]\n";
```

tools/k-rule-find/main.cpp
```diff
@@ -0,0 +1,132 @@
+#include <kllvm/ast/AST.h>
+#include <kllvm/codegen/Debug.h>
+#include <kllvm/parser/KOREParser.h>
+
+#include <llvm/Support/CommandLine.h>
+
+#include <cstdlib>
+#include <dlfcn.h>
```
Baltoli, line 8: No longer needed

```suggestion
```

Baltoli (approved):
Two minor suggestions but otherwise this looks really good. Great work!

PR #711, opened 2023-03-21
Infer shared library extension in `c` mode

Baltoli:
Currently, we require callers of `kompile` to know their platform's shared library extension (and invoke the K tools accordingly). It would be convenient to add the correct extension for them automatically when compiling in C mode.

The positive case of this change is not yet used by any code (https://github.com/runtimeverification/k/pull/3259 and related K as a library work will eventually integration test it), and I'm confident that the existing test suite will catch any inadvertent breakages caused by the PR.


Robertorosmaninho (approved):
That's a great idea! LGTM!

PR #714, opened 2023-03-22
Implement KORE parser for C bindings

Baltoli:
Previously, the only way to construct a KORE pattern from the C bindings was to build it by hand. This PR exposes and tests the KORE parser to the bindings.

### DIFF
--- a/test/c/Inputs/parse.c
+++ b/test/c/Inputs/parse.c
@@ -1,0 +1,38 @@
+#include <assert.h>
+#include <dlfcn.h>
+#include <kllvm-c/kllvm-c.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+typedef kore_pattern *parse_t(char const *);
+typedef char *dump_t(kore_pattern const *);
+
+int main(int argc, char **argv) {
+  if (argc <= 1) {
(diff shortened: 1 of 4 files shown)


test/c/Inputs/parse.c
```diff
@@ -0,0 +1,38 @@
...
+  if (!parse || !dump) {
+    return 3;
+  }
+
+  /*
+    module TEST
+      syntax Foo ::= foo() | bar()
+      rule foo() => bar()
+    endmodule
+  */
```
Robertorosmaninho, line 34: Will we keep the test here as a comment?

Baltoli: It's not normative, but I like to keep the K source around as a convention so it's clearer what behaviour the test should expect. Here it's not super important because we don't do anything other than parsing + printing, but take a look at the other tests in the directory for an idea.

They don't get updated frequently enough for them to be at risk of serious breakage via drifting.

Robertorosmaninho: Got it! That's a great idea, tbh. Maybe we could try it in other contexts, too; I've noticed that it can be hard to debug a test without knowledge about its source code as we only test its definition.

Robertorosmaninho (approved):

PR #719, opened 2023-03-27
Embed KPrint into shared libraries

Baltoli:
The motivation for this PR is similar to the "bison parser libraries" one merged to K recently: https://github.com/runtimeverification/k/pull/3259

When using the LLVM bindings to C currently, a set of files need to be dragged around on disk along with the shared library to get access to all the features exposed by the backend. This PR resolves that issue for _pretty printing_ (as provided by `kprint`).

The implementation strategy taken by this PR is:
1. ~~Build a new tool `kore-embed` that generates a C++ header / implementation combination containing the bytes making up a file, as an array / size combination.~~
2. Add a new `--embed` flag to `llvm-kompile` that uses `xxd -i` to bake files into a compiled backend interpreter library on request.
3. Expose pretty-printing as a new C API function that looks for correctly-named embedded symbols, then parses them to run the `kprint` internals as usual.

An integration test of these changes is also added to the C bindings test suite.

For full K integration, I will follow this up with a frontend PR adding a flag that passes `--embed` correctly to `llvm-kompile`.


dwightguth (changes requested):
I would rather, instead of the kore-embed tool, we simply make use of `ld -r -b binary BLOB -o BLOB.o` which takes a file BLOB and emits an object file containing its binary data verbatim, with the following symbols: 
```
0000000000000004 D _binary_BLOB_end
0000000000000004 A _binary_BLOB_size
0000000000000000 D _binary_BLOB_start
```

Baltoli:
I wasn't aware of that tool; thanks for the pointer! Will update and get rid of `kore-embed`

Baltoli:
Looking further into this, it seems like that option isn't portable to the macOS MachO linker unfortunately.

One alternative is to use `.incbin` on a pre-processed assembly file like this:
```asm
.global SYM_start
foo_start:
  .incbin FILE
.global SYM_end
foo_end:
  .byte 0
.global SYM_size
foo_size:
  .int SYM_end - SYM_start
```

What do you make of that solution @dwightguth?

Baltoli:
I have updated this PR to use `xxd` rather than hand-rolling a binary embedding tool; this is the conclusion that Laurie Tratt comes to as the most portable way to do what we're doing: https://tratt.net/laurie/blog/2022/whats_the_most_portable_way_to_include_binary_blobs_in_an_executable.html

PR #721, opened 2023-03-30
Introducing `r-rule-apply` tool

Robertorosmaninho:
This PR introduces the `k-rule-apply` tool. This tool verifies whether or not a given KORE Pattern from a K Definition matches a given rule.

### To call it from your terminal, simply execute the following command:
``` Bash
$ k-rule-apply <kompiled-dir>  <rule_label_filename>  <kore_patten_filename>  <path_to_shared_lib>
```

### Currently, this PR has the following issues that block it from being merged:
- It can only partially deal when the [match.kind](https://github.com/runtimeverification/llvm-backend/blob/d6386a2bcdafa92b1f100b93f52c94bf5e69bf2c/runtime/util/match_log.cpp#L12) is function.
  - Issue when printing subjects of type `%map*` `%set`


runtime/util/ConfigurationPrinter.cpp
```diff
@@ -284,3 +288,89 @@ void printSortedConfigurationToFile(
   writer w = {file, nullptr};
   printConfigurationInternal(&w, subject, sort, false, &state);
 }
+
+void printMatchResult(
+    std::ostream &os, MatchLog *matchLog, size_t logSize,
+    std::string &definitionPath) {
```
Baltoli, line 294: Does this need to be a non-const reference?

runtime/util/ConfigurationPrinter.cpp
```diff
@@ -284,3 +288,89 @@ void printSortedConfigurationToFile(
...
+  char subjectFilename[15] = "subject_XXXXXX";
+  int sf = mkstemp(subjectFilename);
+  FILE *subject = fdopen(sf, "w");
```
Baltoli, line 297: Not the right place to fix this as we do the same thing in several other places in the backend, but see: https://github.com/runtimeverification/llvm-backend/issues/738

runtime/util/ConfigurationPrinter.cpp
```diff
@@ -284,3 +288,89 @@ void printSortedConfigurationToFile(
...
+
+  char patternFilename[15] = "pattern_XXXXXX";
+  int pf = mkstemp(patternFilename);
+  FILE *pattern = fdopen(pf, "w");
+
+  if (logSize > 0) {
+    for (int i = 0; i < logSize; i++) {
+      if (matchLog[i].kind == MatchLog::SUCCESS)
+        os << "Match succeeds\n";
```
Baltoli, line 306: Very minor, but let's add braces to `if` statements. Unfortunately clang format won't catch this for us

runtime/util/ConfigurationPrinter.cpp
```diff
@@ -284,3 +288,89 @@ void printSortedConfigurationToFile(
...
+  char subjectFilename[15] = "subject_XXXXXX";
+  int sf = mkstemp(subjectFilename);
+  FILE *subject = fdopen(sf, "w");
+
+  char patternFilename[15] = "pattern_XXXXXX";
+  int pf = mkstemp(patternFilename);
+  FILE *pattern = fdopen(pf, "w");
+
+  if (logSize > 0) {
+    for (int i = 0; i < logSize; i++) {
```
Baltoli, line 304: I think the outer if here is redundant; the loop won't iterate if there's no entries, and there is no else associated with it.

tools/k-rule-apply/main.cpp
```diff
@@ -0,0 +1,173 @@
...
+cl::opt<std::string> KompiledDir(
+    cl::Positional, cl::desc("<kompiled-dir>"), cl::Required,
+    cl::cat(KRuleCat));
+
+cl::opt<std::string> RuleLabel(
+    cl::Positional, cl::desc("<rule_label_filename>"), cl::Required,
+    cl::cat(KRuleCat));
+
+cl::opt<std::string> KOREPatternFilename(
+    cl::Positional, cl::desc("<kore_patten_filename>"), cl::Required,
```
Baltoli, line 28:
```suggestion
    cl::Positional, cl::desc("<kore_pattern_filename>"), cl::Required,
```

lib/codegen/Decision.cpp
```diff
@@ -1045,13 +1058,44 @@ void makeStepFunction(
...
+    KOREAxiomDeclaration *axiom, std::string name) {
+  auto blockType = getValueType({SortCategory::Symbol, 0}, module);
+  llvm::FunctionType *funcType = llvm::FunctionType::get(
+      llvm::Type::getVoidTy(module->getContext()), {blockType}, false);
+  std::string wrapperName = "match_" + std::to_string(axiom->getOrdinal());
+  llvm::Function *matchFunc
+      = getOrInsertFunction(module, wrapperName, funcType);
+  std::string debugName = name;
+  if (axiom->getAttributes().count("label")) {
+    debugName = axiom->getStringAttribute("label") + "_fastcc_" + ".match";
```
Baltoli, line 1072: Do other usages respect this `_fastcc_` suffix? I don't see any other references to it in the code

Baltoli (changes requested):
Looks very good largely - I have some minor comments, but the test suite fails to run on my machine and there are some merge conflicts that will need to be addressed; I will re-review when those issues are fixed.

runtime/util/ConfigurationPrinter.cpp
```diff
@@ -284,3 +288,89 @@ void printSortedConfigurationToFile(
...
+
+  close(pf);
+  remove(patternFilename);
+}
+
+void printValueOfType(
+    std::ostream &os, std::string definitionPath, void *value,
+    std::string type) {
+  if (type.compare("%mpz*") == 0) {
+    os << reinterpret_cast<mpz_ptr>(value);
```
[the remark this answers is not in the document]
Robertorosmaninho: Yes, it has an `operator<<` overload, so we get the integer representation printed with this operation!

runtime/util/ConfigurationPrinter.cpp
Robertorosmaninho: not at all! I'll change it to a const reference!

runtime/util/ConfigurationPrinter.cpp
Robertorosmaninho: Any ideas on the best way to handle this? I may change here and take the PR to work on applying the same solution!

tools/k-rule-apply/main.cpp
```diff
@@ -0,0 +1,173 @@
...
+
+cl::OptionCategory KRuleCat("k-rule-apply options");
+
+cl::opt<std::string> KompiledDir(
+    cl::Positional, cl::desc("<kompiled-dir>"), cl::Required,
+    cl::cat(KRuleCat));
+
+cl::opt<std::string> RuleLabel(
+    cl::Positional, cl::desc("<rule_label_filename>"), cl::Required,
+    cl::cat(KRuleCat));
```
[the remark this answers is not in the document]
Robertorosmaninho: Awesome!

tools/k-rule-apply/main.cpp
```diff
@@ -0,0 +1,173 @@
...
+  // Check if the shared library exits in the given location.
+  if (!handle) {
+    std::cerr << "Error: " << dlerror() << "\n";
+    return EXIT_FAILURE;
+  }
+
+  // Get utils functions from the shared lib, cast them to their right type, and
+  // call each with its appropriate argument if any.
+  void *construct_ptr = dlsym(
+      handle, "_Z29constructInitialConfigurationPKN5kllvm11KOREPatternE");
```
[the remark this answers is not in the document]
Robertorosmaninho: I addressed this issue, and to make the `main.cpp` clean, I created a header file `auxiliar.h` with those functions. The name of this fine may need refactoring. I'm open to suggestions on this.

tools/k-rule-apply/main.cpp
```diff
@@ -0,0 +1,173 @@
...
+    dlclose(handle);
+    return EXIT_FAILURE;
+  }
+
+  void *resetMatchReason_ptr = dlsym(handle, "_Z16resetMatchReasonv");
+  if (resetMatchReason_ptr == NULL) {
+    std::cerr << "Error: " << dlerror() << "\n";
+    dlclose(handle);
+    return EXIT_FAILURE;
+  }
```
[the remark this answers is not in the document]
Robertorosmaninho: I could only extract a little of this logic as the caller of the functions should be responsible for returning error codes to the user, but I limited it to the functions that return a value that must be used in the future.

Robertorosmaninho:
As far as I can see, this tool is as good as our current implementation of GDB @Baltoli.

runtime/util/ConfigurationPrinter.cpp
Baltoli: Sure, I'll write a bit more context in #738 and assign you.

Baltoli (approved):
Really nice work @Robertorosmaninho - thanks for this!

Issue #730, opened 2023-04-11
Support LLVM 16


Baltoli:
LLVM 16 was released a few weeks ago and is already the default version for Homebrew. We should test the LLVM backend to make sure that it can be built with version 16; if not we need to update documentation pointing to Homebrew formulae to reflect this.

Running with version 16 in CI is blocked until the Nix maintainers ship it properly; work appears to already be ongoing here. We should nevertheless still check macOS informally.

* Homebrew formula page: https://formulae.brew.sh/formula/llvm#default
* Nixpkgs PR: https://github.com/NixOS/nixpkgs/pull/223282

Things that need to be changed:
* [x] Uses of `getInstList` on a basic block
* [x] `getABITypeAlign[ment]`
* [x] Invocation of `opt`

RaitoBezarius:
FWIW, LLVM16 will (hopefully) be part of NixOS 23.05 and land in unstable soon.

Baltoli:
Thanks @RaitoBezarius - that's helpful to know!

Baltoli:
@theo25 to investigate the use of opaque pointers in this version

theo25:
I looked into the release notes of LLVM 16 and it does not seem that the use of opaque pointers is mandated by this release. However, the release notes found in the main branch on github do list [the opaque pointer change](https://github.com/llvm/llvm-project/blob/main/llvm/docs/ReleaseNotes.rst#changes-to-the-llvm-ir), so we can assume that the change is coming with the LLVM 17 release.

Baltoli:
The Opaque Pointers migration document reveals that their introduction is not _mandatory_ in LLVM 16, but they are enabled by default and the old system is no longer supported. This suggests that the LLVM 16 transition is the right point at which to make sure they work properly for the LLVM backend.

Some debugging reveals that there is a subtle issue we have previously missed when migrating over to opaque pointers. In transparent mode, we make equality comparisons between pointer types: https://github.com/runtimeverification/llvm-backend/blob/046b5968778c4cb1fa8bc5b69918248e41504bb4/lib/codegen/EmitConfigParser.cpp#L852-L854

[...]

PR #731, opened 2023-04-11
Support LLVM 16 informally

Baltoli:
This PR adds preliminary support for LLVM 16

Related to #730


Baltoli:
Things that need to be changed:
* Uses of `getInstList` on a basic block
* `getABITypeAlign[ment]`

PR #737, opened 2023-04-14
Add handling of \u and \U unicode escape sequences

Scott-Guest:
Previously, KOREScanner failed to handle \u and \U escape sequences, contributing to the issue in runtimeverification/k#3086. 

This PR adds proper support for them, choosing to UTF-8 encode everything for the char[] representation.


lib/parser/KOREScanner.l
```diff
@@ -95,12 +100,12 @@ yyin = in;
...
 <COMM,STR,INITIAL><<EOF>> {
   if(YYSTATE != INITIAL)
-    error(*loc, "Either a comment or string haven't been closed\n");
+    error(*loc, "Either a comment or string hasn't been closed\n");
 
   return token::TOKEN_EOF;
 }
 
-. { error(*loc, std::string("Unknown token \"") + yytext + std::string("\"\n")); }
+<COMM,STR,INITIAL>. { error(*loc, std::string("Unknown token \"") + yytext + std::string("\"\n")); }
```
dwightguth, line 108: If your goal is to include every start condition, you can do that with `<*>`

dwightguth (approved):

PR #743, opened 2023-04-18
Fix mysterious unicode segfault

Baltoli:
The issue is that on some systems `lit` defaults to the `POSIX` locale, which causes a segfault when decoding decision tree YAML. This PR fixes the issue by just forcing a UTF-8 locale for the new test case.

We can't really recover from the issue, but the PR also adds some nicer error reporting to prevent the confusing segfaults we were seeing previously.

### DIFF
--- a/test/defn/test-unicode.kore
+++ b/test/defn/test-unicode.kore
@@ -1,4 +1,4 @@
-// RUN: %interpreter
+// RUN: LC_ALL=en_US.UTF-8 %interpreter
 // RUN: %check-diff
 [topCellInitializer{}(LblinitGeneratedTopCell{}()), org'Stop'kframework'Stop'attributes'Stop'Source{}("Source(/home/sguest/work/k-test/emoji/emoji.k)")]
 
--- a/lib/codegen/CMakeLists.txt
+++ b/lib/codegen/CMakeLists.txt
@@ -11,5 +11,5 @@ add_library(Codegen
 )
 
 target_link_libraries(Codegen
-  PUBLIC  AST
+  PUBLIC  AST fmt::fmt
   PRIVATE base64)
--- a/lib/codegen/DecisionParser.cpp
+++ b/lib/codegen/DecisionParser.cpp
@@ -5,6 +5,8 @@
 
 #include <base64/base64.h>
 
+#include <fmt/format.h>
+
 #include <yaml.h>
 
 #include <llvm/IR/DerivedTypes.h>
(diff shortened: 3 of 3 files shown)


test/defn/test-unicode.kore
```diff
@@ -1,4 +1,4 @@
-// RUN: %interpreter
+// RUN: LC_ALL=en_US.UTF-8 %interpreter
```
Scott-Guest, line 1: Why don't we just ensure that the YAML is always read/written in UTF-8 regardless of the system locale?

```
diff --git a/matching/src/main/scala/org/kframework/backend/llvm/matching/dt/DecisionTree.scala b/matching/src/main/scala/org/kframework/backend/llvm/matching/dt/DecisionTree.scala
index 817ea0c..e0d9f1c 100644
--- a/matching/src/main/scala/org/kframework/backend/llvm/matching/dt/DecisionTree.scala
+++ b/matching/src/main/scala/org/kframework/backend/llvm/matching/dt/DecisionTree.scala
@@ -2,22 +2,29 @@ package org.kframework.backend.llvm.matching.dt
 
 import org.kframework.backend.llvm.matching.Occurrence
 import org.kframework.backend.llvm.matching.pattern._
-import java.io.File
-import java.io.FileWriter
+
+import java.io.{File, FileOutputStream, OutputStreamWriter}
 import java.util
 import java.util.concurrent.ConcurrentHashMap
-
 import org.yaml.snakeyaml.Yaml
 
+import java.nio.charset.Charset
+
 sealed trait DecisionTree {
   def serializeToYaml(file: File): Unit = {
-    val writer = new FileWriter(file)
+    val writer = new OutputStreamWriter(
+      new FileOutputStream(file),
+      Charset.forName("UTF-8").newEncoder()
+    );
     new Yaml().dump(representation, writer)
     writer.close()
   }
 
   def serializeToYaml(file: File, residuals: Seq[(Pattern[String], Occurrence)]): Unit = {
-    val writer = new FileWriter(file)
+    val writer = new OutputStreamWriter(
+      new FileOutputStream(file),
+      Charset.forName("UTF-8").newEncoder()
+    );
     val residualRepr = new util.ArrayList[AnyRef]()
     for (entry <- residuals) {
       val pair = new util.ArrayList[AnyRef]()
```

```
diff --git a/lib/codegen/DecisionParser.cpp b/lib/codegen/DecisionParser.cpp
index 9e9d191..0d80fde 100644
--- a/lib/codegen/DecisionParser.cpp
+++ b/lib/codegen/DecisionParser.cpp
@@ -362,6 +362,7 @@ DecisionNode *parseYamlDecisionTreeFromString(
   yaml_parser_t parser;
   yaml_document_t doc;
   yaml_parser_initialize(&parser);
+  yaml_parser_set_encoding(&parser, YAML_UTF8_ENCODING);
   yaml_parser_set_input_string(
       &parser, (unsigned char *)yaml.c_str(), yaml.size());
   yaml_parser_load(&parser, &doc);
```

Scott-Guest (changes requested):
Nice find, this was a confusing one!

I think we can fix it more generally though, see my comment.

Baltoli:
Ah, really nice work - that's definitely a better fix! I'll close this PR @Scott-Guest - please go ahead and open one with your fix and I'll review.

PR #745, opened 2023-04-19
Modernize tests

dwightguth:
This PR does the following:

1. Disables support for old rule encodings in the llvm backend. Going forward, the llvm backend will not be backwards compatible with out of date rule encodings.
2. For each test that fails due to the out of date rule encodings present in its kore definition, we reconstruct a version of the original .k file that, when compiled, generates the same test behavior and passes the test. We then check in the compiled versions of these k definitions, but we also check in the original k files so that they can be used to regenerate the kore definition in the future.
3. We update the output files where necessary in order to fix ways in which the behavior of the k frontend has changed since the test was written. For example, old tests may not include the result sort of a production in a klabel, and they may also allow k sequences inside the List/Map/Set sorts. There are also a few other miscellaneous differences of this type.

Note: this is not a complete undertaking. Some of the tests already passed, presumably because they already had the new rule encoding. These tests will need to have their .k files reconstructed as part of future work. We should also eventually write a script that regenerates all the kore files.


Scott-Guest:
For my own sanity, I rewrote the changed functions declaratively:

`KOREAxiomDeclaration::getRightHandSide()` before:
```
RHS(\implies{\bottom{},\and{_,\and{_,\rewrites{_,X}}}}) = X
RHS(\implies{\bottom{},\and{_,\rewrites{_,\and{_,X}}}}) = X
RHS(\implies{\bottom{},\and{_,\rewrites{_,\and{_,X}}}}) = X
RHS(\implies{\bottom{},_}) = assert
RHS(\implies{_,\and{\equals{_,X},_}}) = X
RHS(\implies{_,\rewrites{_,\and{_,X}}}) = X
RHS(\implies{_,\equals{_,\and{X,_}}}) = X
RHS(\and{_,\and{_,\rewrites{_,X}}}) = X
RHS(\equals{_,X}) = X
RHS(\rewrites{_,\and{X,Y}})
  | X's constructor is not built-in = X
  | otherwise = Y
RHS(_) = assert 
```
`KOREAxiomDeclaration::getRightHandSide()` after:
```
RHS(\implies{_,\equals{_,\and{X,_}}}) = X
RHS(\equals{_,X}) = X
RHS(\rewrites{_,\and{X,Y}})
  | X's constructor is not built-in = X
  | otherwise = Y
RHS(_) = assert
```
`KOREAxiomDeclaration::getRequires()` before:
```
requires(\implies{\and{\not{_},\equals{X,_}},_}) = X
requires(\implies{\and{\not{_},\top{}},_}) = nullptr
requires(\implies{\and{\not{_},\bottom{}},\and{\equals{X,_},_}}) = X
requires(\implies{\and{\not{_},\bottom{}},\and{\top{},_}}) = nullptr
requires(\implies{\and{\not{_},\bottom{}},\rewrites{\and{\equals{X,_},_},_}}) = X
requires(\implies{\and{\not{_},\bottom{}},\rewrites{\and{\top{},_},_}}) = nullptr
requires(\implies{\and{\not{_},\and{\top{},_}},_}) = nullptr
requires(\implies{\and{\not{_},\and{\equals{X,_},_}},_}) = X
requires(\implies{\equals{X,_},_}) = X
requires(\implies{\top{},_}) = nullptr
requires(\implies{\bottom{},\and{\equals{X,_},_}}) = X
requires(\implies{\bottom{},\and{\top{},_}}) = nullptr
requires(\implies{\bottom{},\rewrites{\and{\equals{X,_},_},_}}) = X
requires(\implies{\bottom{},\rewrites{\and{\top{},_},_}}) = nullptr
requires(\implies{\and{\top{},_},_}) = nullptr
requires(\implies{\and{\equals{X,_},_},_}) = X 
requires(\and{\equals{X,_},_}) = X
requires(\and{\top{},_}) = nullptr
requires(\equals{_,_}) = nullptr
requires(\rewrites{\and{\equals{X,_},_},_}) = X
requires(\rewrites{\and{\top{},\equals{X,_}},_}) = X
requires(\rewrites{\and{\not{_},\and{\equals{X,_},_}},_}) = X
requires(\rewrites{\and{_,_},_}) = nullptr
requires(_) = assert
```
`KOREAxiomDeclaration::getRequires()` after:
```
requires(\implies{\and{\not{_},\and{\top{},_}},_}) = nullptr
requires(\implies{\and{\not{_},\and{\equals{X,_},_}},_}) = X
requires(\implies{\and{\top{},_},_}) = nullptr
requires(\implies{\and{\equals{X,_},_},_}) = X
requires(\equals{_,_}) = nullptr
requires(\rewrites{\and{\equals{X,_},_},_}) = X
requires(\rewrites{\and{\top{},\equals{X,_}},_}) = X
requires(\rewrites{\and{\not{_},\and{\equals{X,_},_}},_}) = X
requires(\rewrites{\and{_,_},_}) = nullptr
requires(_) = assert
```

Does this seem to match the semantics you were going for?

Baltoli (approved):
The "after" semantics here all look reasonable to me per @Scott-Guest's comment; I'm not familiar with the details of the old rule encoding, but I'm happy with that change given that the K test suite passes without the old encoding.

Likewise, the K definitions that have been reintroduced all look sensible compared to the old KORE that we were using. I will open a new issue for the regenerating-KORE feature.

Baltoli:
I commented [here](https://github.com/runtimeverification/k/issues/3320) regarding @Scott-Guest's translation of the LLVM backend's handling of rewrite rules. We should ensure that any specification effort for KORE validates these translations.

PR #750, opened 2023-04-27
Introducing API to create temporary files

### LINKED ISSUE #738: Clean up temporary file handling
In a number of places in the backend, we're following a similar set of actions to create temporary files with `mkstemp`:
1. Allocate `char[]` buffer on the stack with a template string.
2. Call `mkstemp` to create a unique name from that template
3. `fdopen` the descriptor returned by `mkstemp`
4. `fclose` and `remove` the file

We should abstract this behind a reusable `RAII` class; there are enough different resources being managed here that it gets fiddly and repetitive trying to clean them all up properly. Doing so will also help us to move away from `mkstemp` in the future if (say) `std::filesystem` offers a more modern API.

Robertorosmaninho:
Fixes #738

### DIFF
--- a/include/kllvm/util/FileRAII.h
+++ b/include/kllvm/util/FileRAII.h
@@ -1,0 +1,51 @@
+#include <cstdlib>
+#include <fstream>
+#include <iostream>
+#include <unistd.h>
+
+class FileRAII {
+private:
+  int temp_fd;
+  char *temp_filename = nullptr;
+  FILE *temp_c_file = nullptr;
+  std::ofstream *temp_cpp_file = nullptr;
+
+public:
+  FileRAII(char *template_name) {
+    temp_filename = template_name;
+    temp_fd = mkstemp(temp_filename);
+
+    if (temp_fd == -1) {
+      std::perror("Could not create temporary file: ");
+      std::exit(1);
+    }
+  }
+
+  ~FileRAII() {
+    if (temp_cpp_file != nullptr) {
+      temp_cpp_file->close();
+      delete temp_cpp_file;
+    }
+
+    if (temp_c_file) {
+      fclose(temp_c_file);
+    }
+    
+    close(temp_fd);
+    remove(temp_filename);
+  }
+
+  FILE *getFILE() {
+    if (!temp_c_file) {
+      temp_c_file = fdopen(temp_fd, "r");
+    }
+    return temp_c_file;
+  }
+
+  std::ofstream *getOFStream() {
+    if (!temp_cpp_file) {
+      temp_cpp_file = new std::ofstream(temp_filename);
+    }
+    return temp_cpp_file;
+  }
+};
--- a/include/kllvm/util/FileRAII.h
+++ b/include/kllvm/util/FileRAII.h
@@ -30,14 +30,16 @@ public:
     if (temp_c_file) {
       fclose(temp_c_file);
     }
-    
+
     close(temp_fd);
     remove(temp_filename);
   }
 
-  FILE *getFILE() {
+  int getTempFd() { return temp_fd; }
+
+  FILE *getFILE(char const *mode = "r") {
     if (!temp_c_file) {
-      temp_c_file = fdopen(temp_fd, "r");
+      temp_c_file = fdopen(temp_fd, mode);
     }
     return temp_c_file;
   }
--- a/lib/parser/KOREParser.cpp
+++ b/lib/parser/KOREParser.cpp
@@ -1,5 +1,6 @@
-#include "kllvm/parser/KOREParser.h"
 #include "kllvm/ast/AST.h"
+#include "kllvm/util/FileRAII.h"
+#include "kllvm/parser/KOREParser.h"
 #include "kllvm/parser/KOREScanner.h"
 
 #include <cstdio>
@@ -13,20 +14,10 @@ namespace parser {
 
 std::unique_ptr<KOREParser> KOREParser::from_string(std::string text) {
   char temp_file_name[] = "tmp.parse.XXXXXX";
-
-  int temp_fd = mkstemp(temp_file_name);
-  if (temp_fd == -1) {
-    std::perror("Could not create temporary parsing file: ");
-    std::exit(1);
-  }
-  close(temp_fd);
-
-  auto os = std::ofstream(temp_file_name);
-  os << text;
-  os.close();
+  auto os =  FileRAII(temp_file_name).getOFStream();
+  *os << text;
 
   auto parser = std::make_unique<KOREParser>(temp_file_name);
-  std::remove(temp_file_name);
   return parser;
 }
 
--- a/runtime/util/ConfigurationPrinter.cpp
+++ b/runtime/util/ConfigurationPrinter.cpp
@@ -14,6 +14,7 @@
 #include <vector>
 
 #include <kllvm/parser/KOREParser.h>
+#include <kllvm/util/FileRAII.h>
 
 #include "runtime/alloc.h"
 #include "runtime/header.h"
@@ -305,12 +306,9 @@ extern "C" void printMatchResult(
     std::ostream &os, MatchLog *matchLog, size_t logSize,
     const std::string &definitionPath) {
   char subjectFilename[15] = "subject_XXXXXX";
-  int sf = mkstemp(subjectFilename);
-  FILE *subject = fdopen(sf, "w");
-
+  auto subject = FileRAII(subjectFilename).getFILE();
   char patternFilename[15] = "pattern_XXXXXX";
-  int pf = mkstemp(patternFilename);
-  FILE *pattern = fdopen(pf, "w");
+  auto pattern = FileRAII(patternFilename).getFILE();
 
   for (int i = 0; i < logSize; i++) {
     if (matchLog[i].kind == MatchLog::SUCCESS) {
@@ -344,12 +342,6 @@ extern "C" void printMatchResult(
       os << ") => " << *reinterpret_cast<bool *>(matchLog[i].result) << "\n";
     }
   }
-
-  close(sf);
-  remove(subjectFilename);
-
-  close(pf);
-  remove(patternFilename);
 }
 
 void printValueOfType(
@@ -359,16 +351,13 @@ void printValueOfType(
     os << reinterpret_cast<mpz_ptr>(value);
   } else if (type.compare("%block*") == 0) {
     if ((((uintptr_t)value) & 3) == 1) {
-      char subjectFilename[15] = "pattern_XXXXXX";
-      int sf = mkstemp(subjectFilename);
-      FILE *pattern = fdopen(sf, "w");
+      char subjectFilename[15] = "subject_XXXXXX";
+      auto subject = FileRAII(subjectFilename).getFILE();
       string *s = printConfigurationToString(reinterpret_cast<block *>(value));
       auto strSubjectSort = std::string(s->data, len(s));
-      fprintf(pattern, "%s", strSubjectSort.c_str());
-      fflush(pattern);
+      fprintf(subject, "%s", strSubjectSort.c_str());
+      fflush(subject);
       kllvm::printKORE(os, definitionPath, subjectFilename, false, true);
-      close(sf);
-      remove(subjectFilename);
     } else if ((((uintptr_t)value) & 1) == 0) {
       auto s = reinterpret_cast<string *>(value);
       os << std::string(s->data, len(s));
--- a/runtime/io/logTerm.cpp
+++ b/runtime/io/logTerm.cpp
@@ -1,3 +1,4 @@
+#include "kllvm/util/FileRAII.h"
 #include "runtime/header.h"
 
 #include <kllvm/printer/printer.h>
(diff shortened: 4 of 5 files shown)


Robertorosmaninho:
@Baltoli This PR is mostly ready for review, although some tests for c bindings seem flaky after this modification.

Robertorosmaninho:
Also, both [runtime/io/io.cpp](https://github.com/runtimeverification/llvm-backend/blob/2415a6435bf2fe0585b988a3a6034b2d19037128/runtime/io/io.cpp#L634) and [tools/kore-strip/main.cpp](https://github.com/runtimeverification/llvm-backend/blob/2415a6435bf2fe0585b988a3a6034b2d19037128/tools/kore-strip/main.cpp#L107) still have its own `mkstemp` due to their unique uses.

runtime/util/ConfigurationPrinter.cpp
```diff
@@ -327,11 +325,13 @@ extern "C" void printMatchResult(
...
       fflush(subject);
-      kllvm::printKORE(os, definitionPath, subjectFilename, false, true);
+      kllvm::printKORE(
+          os, definitionPath, subject_raii.getFilename(), false, true);
       os << "does not match pattern: \n";
       fprintf(pattern, "%s\n", matchLog[i].pattern);
       fflush(pattern);
-      kllvm::printKORE(os, definitionPath, patternFilename, false, true);
+      kllvm::printKORE(
+          os, definitionPath, pattern_raii.getFilename(), false, true);
```
Baltoli, line 334: Can we clean this code up so we're using a consistent style for working with these files, rather than mixing C and C++-style code?

runtime/util/ConfigurationPrinter.cpp
```diff
@@ -359,16 +353,13 @@ void printValueOfType(
...
       string *s = printConfigurationToString(reinterpret_cast<block *>(value));
       auto strSubjectSort = std::string(s->data, len(s));
-      fprintf(pattern, "%s", strSubjectSort.c_str());
-      fflush(pattern);
-      kllvm::printKORE(os, definitionPath, subjectFilename, false, true);
-      close(sf);
-      remove(subjectFilename);
+      fprintf(subject, "%s", strSubjectSort.c_str());
+      fflush(subject);
+      kllvm::printKORE(os, definitionPath, f.getFilename(), false, true);
```
Baltoli, line 362: Similar comment re. C/C++ mixing here

Baltoli (changes requested):

Baltoli:
> Also, both [runtime/io/io.cpp](https://github.com/runtimeverification/llvm-backend/blob/2415a6435bf2fe0585b988a3a6034b2d19037128/runtime/io/io.cpp#L634) and [tools/kore-strip/main.cpp](https://github.com/runtimeverification/llvm-backend/blob/2415a6435bf2fe0585b988a3a6034b2d19037128/tools/kore-strip/main.cpp#L107) still have its own `mkstemp` due to their unique uses.

I agree in the case of `hook_IO_mkstemp`, as that function just models a user's own call to `mkstemp`. I think we can probably find a way to refactor the usage in `kore_strip` though.

include/kllvm/util/FileRAII.h
```diff
@@ -0,0 +1,55 @@
...
+#include <fstream>
+#include <iostream>
+#include <unistd.h>
+
+class FileRAII {
+private:
+  int temp_fd;
+  char *temp_filename = nullptr;
+  FILE *temp_c_file = nullptr;
+  std::ofstream *temp_cpp_file = nullptr;
```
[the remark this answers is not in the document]
Baltoli: I _think_ on first inspection of the test results, this use-after free is breaking an integration test.

Baltoli:
> This PR is mostly ready for review, although some tests for c bindings seem flaky after this modification.

See my comment above; I think there's a use-after free in your code that's breaking the test.

include/kllvm/util/FileRAII.h
```diff
@@ -0,0 +1,55 @@
...
+      temp_cpp_file->close();
+      delete temp_cpp_file;
+    }
+
+    if (temp_c_file) {
+      fclose(temp_c_file);
+    }
+
+    close(temp_fd);
+    remove(temp_filename);
```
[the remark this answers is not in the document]
Robertorosmaninho: If I don't use something like it:
```
if (temp_cpp_file.has_value()) {
      temp_cpp_file.value()->close();
      delete temp_cpp_file.value();
 }
```
I'll end up having some memory leaks:
```
valgrind ./a.out              
==63322== Memcheck, a memory error detector
==63322== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==63322== Using Valgrind-3.18.1 and LibVEX; rerun with -h for copyright info
==63322== Command: ./a.out
==63322== 
==63322== 
==63322== HEAP SUMMARY:
==63322==     in use at exit: 9,176 bytes in 3 blocks
==63322==   total heap usage: 4 allocs, 1 frees, 81,880 bytes allocated
==63322== 
==63322== LEAK SUMMARY:
==63322==    definitely lost: 512 bytes in 1 blocks
==63322==    indirectly lost: 8,192 bytes in 1 blocks
==63322==      possibly lost: 0 bytes in 0 blocks
==63322==    still reachable: 472 bytes in 1 blocks
==63322==         suppressed: 0 bytes in 0 blocks
==63322== Rerun with --leak-check=full to see details of leaked memory
==63322== 
==63322== For lists of detected and suppressed errors, rerun with: -s
==63322== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0)
```
Given this test file:
```
#include "temporary_file.h"

int main() {
    auto file = temporary_file("pattern_XXXXXX");
    auto cppFile = file. ofstream();

    return 0;
}
```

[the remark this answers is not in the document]
Baltoli: My suggestion was to make that member a `std::optional` so that the memory management is handled for you.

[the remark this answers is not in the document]
Baltoli: Are you making the member a `std::optional<std::ostream *>`? Why not just a `std::optional<std::ostream>`?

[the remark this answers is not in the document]
Robertorosmaninho: Yes! I can use it, but then the `ofstream()` function would be like
```
  std::ofstream *ofstream() {
    if (!temp_cpp_file) {
      temp_cpp_file = std::ofstream(temp_filename);
    }
    return &temp_cpp_file.value();
  }
```
ok?

lib/parser/KOREParser.cpp
```diff
@@ -14,8 +14,8 @@ namespace parser {
 
 std::unique_ptr<KOREParser> KOREParser::from_string(std::string text) {
   auto temp_file = temporary_file("tmp.parse.XXXXXX");
-  auto os = temp_file.ofstream();
-  *os << text;
+  auto &os = temp_file.ofstream();
+  os << text;
```
Baltoli:
```suggestion
  temp_file.ofstream() << text;
```

include/kllvm/util/temporary_file.h
```diff
@@ -0,0 +1,50 @@
...
+    temp_filename = template_name;
+
+    if (temp_fd == -1) {
+      std::runtime_error("Could not create temporary file!");
+    }
+  }
+
+  ~temporary_file() {
+    close(temp_fd);
+    remove(temp_filename.data());
```
Baltoli, line 30: This can be a call to `std::filesystem::remove`, I think: https://en.cppreference.com/w/cpp/filesystem/remove

include/kllvm/util/temporary_file.h
```diff
@@ -0,0 +1,50 @@
...
+public:
+  temporary_file(std::string template_name)
+      : temp_c_file(nullptr, &fclose) {
+    temp_fd = mkstemp(template_name.data());
+    temp_filename = template_name;
+
+    if (temp_fd == -1) {
+      std::runtime_error("Could not create temporary file!");
+    }
+  }
```
Baltoli, line 26: There's a slightly nuanced improvement we can make to this constructor. As written, we're default-initializing `temp_filename`, then assigning it a new value with its `operator=(std::string const&)`. What we can do instead is member-initialize it, then call `mkstemp` on the member variable rather than the constructor argument. The code will look something like:
```c++
  temporary_file(std::string template_name)
      : temp_filename(template_name)
      , temp_c_file(nullptr, &fclose)
      , temp_fd(mkstemp(temp_filename.data())) {
    if (temp_fd == -1) {
      std::runtime_error("Could not create temporary file!");
    }
  }
```

Note that you'll need to reorder the class members to get this to compile.

include/kllvm/util/temporary_file.h
```diff
@@ -0,0 +1,50 @@
+#include <cstdlib>
+#include <fstream>
+#include <iostream>
+#include <memory>
+#include <optional>
+#include <unistd.h>
+
+typedef std::unique_ptr<FILE, decltype(&fclose)> temp_c_file_type;
```
Baltoli, line 8: Another non-obvious, nuanced improvement to this code is to use a default-constructible functor type for the deleter, rather than `decltype(&fclose)`: https://stackoverflow.com/a/63942318

include/kllvm/util/temporary_file.h
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  ~temporary_file() {
+    close(temp_fd);
+    remove(temp_filename.data());
+  }
+
+  int descriptor() { return temp_fd; }
+
+  std::string filename() { return temp_filename; }
```
Baltoli, line 35:
```suggestion
  int descriptor() const { return temp_fd; }

  std::string const& filename() const { return temp_filename; }
```

include/kllvm/util/temporary_file.h
```diff
@@ -38,6 +38,11 @@ class temporary_file {
     if (!temp_c_file.get()) {
       temp_c_file = temp_c_file_type(fdopen(temp_fd, mode.data()), &fclose);
     }
+
+    if (!temp_c_file.get()) {
+      std::runtime_error("Could not open file " + temp_filename);
+    }
+
```
Robertorosmaninho, line 45: I added this second check to sanitize the `fdopen` function and to include the [`check_fopen`](https://github.com/runtimeverification/llvm-backend/blob/aa0a2278a8da15e298a0aa616928ac98e2fd006a/tools/kore-strip/main.cpp#L40) behavior from `kore-strip`, any thoughts on this @Baltoli?

include/kllvm/util/temporary_file.h
Baltoli: No problem - nice digging

include/kllvm/util/temporary_file.h
```diff
@@ -0,0 +1,55 @@
...
+#include <filesystem>
+#include <fstream>
+#include <iostream>
+#include <memory>
+#include <optional>
+#include <unistd.h>
+struct deleter {
+  void operator()(FILE *fp) const { std::fclose(fp); }
+};
+class temporary_file {
```
Baltoli, line 11:
```suggestion
#include <unistd.h>

struct deleter {
  void operator()(FILE *fp) const { std::fclose(fp); }
};

class temporary_file {
```

include/kllvm/util/temporary_file.h
```diff
@@ -0,0 +1,55 @@
...
+
+  std::string const &filename() const { return temp_filename; }
+
+  FILE *file_pointer(std::string const &mode = "r") {
+    if (!temp_c_file.get()) {
+      auto f = fdopen(temp_fd, mode.data());
+      if (f) {
+        temp_c_file = std::unique_ptr<FILE, deleter>(f);
+      } else {
+        std::runtime_error("Could not open file " + temp_filename);
```
Baltoli, line 42: We can add the information from `perror` here; if a call that sets `errno` fails I almost always find myself looking for that information if it's not printed.

include/kllvm/util/temporary_file.h
```diff
@@ -0,0 +1,55 @@
...
+    close(temp_fd);
+    remove(temp_filename.data());
+  }
+
+  int descriptor() const { return temp_fd; }
+
+  std::string const &filename() const { return temp_filename; }
+
+  FILE *file_pointer(std::string const &mode = "r") {
+    if (!temp_c_file.get()) {
```
Baltoli, line 37:
```suggestion
    if (!temp_c_file) {
```

Baltoli (approved):
A handful of really small changes, but feel free to automerge when they are addressed. Great work!

include/kllvm/util/temporary_file.h
Robertorosmaninho:
```
} else {
        auto str = std::stringstream{};
        str << "Could not open file " << temp_filename;
        std::perror(str.str().c_str());
        std::runtime_error(str.str());
}
```
better this way?

Baltoli: https://en.cppreference.com/w/cpp/string/byte/strerror

Sorry, my bad - I meant `std::strerror` rather than `perror`. That way we can put the error data into the exception, rather than forcing it to `stderr`.

Robertorosmaninho: Done! I also notice that I wasn't throwing the errors and already fix it.

Issue #747, opened 2023-04-27
C semantics fail with new-format KORE rewrites


Baltoli:
Context: https://github.com/runtimeverification/k/issues/3035

If the new-format rewrites are enabled for the C semantics, then the translation step fails when trying to rewrite Implementation.Initialize().

We should identify why this happens and extract the issue to a test case.

Baltoli:
Looking into this some more, it seems that much smaller examples from the regression test suite are also failing with what seems like a similar problem. In the interests of debugging faster, I'm going to look at why that is first, then come back to the C semantics.

Baltoli:
The problem is just that there's a missing case in the Scala pattern matching code for rules that do not have an explicit side condition.

PR #751, opened 2023-04-28
Add missing case to pattern matching for new-format rules

Baltoli:
This case corresponds to rules with no explicit side condition where the order of arguments has been swapped to support the new-format [no-antileft rewrites](https://github.com/runtimeverification/k/pull/3314).

The original change to support these rules added the corresponding case for rules _with_ side conditions, and the line added here was missing.

I will add a test to the frontend for this change while https://github.com/runtimeverification/llvm-backend/pull/745 is still in flight; once that is merged I will add the same test to the backend suite.

Fixes https://github.com/runtimeverification/llvm-backend/issues/747; I have confirmed by running the C semantics test suite with the new-format rewrites.

### DIFF
--- a/matching/src/main/scala/org/kframework/backend/llvm/matching/Parser.scala
+++ b/matching/src/main/scala/org/kframework/backend/llvm/matching/Parser.scala
@@ -157,6 +157,7 @@ object Parser {
       case Rewrites(s, And(_, Equals(_, _, pat, _), l), And(_, _, r)) => Some((None, B.Rewrites(s, l, r), Some(pat)))
       case Rewrites(s, And(_, l, Equals(_, _, pat, _)), And(_, r, _)) => Some((None, B.Rewrites(s, l, r), Some(pat)))
       case Rewrites(s, And(_, Top(_), l), And(_, _, r)) => Some((None, B.Rewrites(s, l, r), None))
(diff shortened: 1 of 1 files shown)


matching/src/main/scala/org/kframework/backend/llvm/matching/Parser.scala
```diff
@@ -157,6 +157,7 @@ object Parser {
       case Rewrites(s, And(_, Equals(_, _, pat, _), l), And(_, _, r)) => Some((None, B.Rewrites(s, l, r), Some(pat)))
       case Rewrites(s, And(_, l, Equals(_, _, pat, _)), And(_, r, _)) => Some((None, B.Rewrites(s, l, r), Some(pat)))
       case Rewrites(s, And(_, Top(_), l), And(_, _, r)) => Some((None, B.Rewrites(s, l, r), None))
+      case Rewrites(s, And(_, l, Top(_)), And(_, r, _)) => Some((None, B.Rewrites(s, l, r), None))
```
Baltoli, line 160: This line is just the same as the line above, but with the order of the constrained-term arguments swapped.

Robertorosmaninho (approved):
LGTM!

PR #755, opened 2023-05-10
Revert "Revert "Introducing API to create temporary files""

### LINKED ISSUE #738: Clean up temporary file handling

Robertorosmaninho:
Reverts runtimeverification/llvm-backend#753 that reverted #750:

This PR introduces a reusable RAII class that manages the lifetime of temporary files necessary for internal backend usage.
This new resource avoids using several `mkstemp`, `char[] allocation` and `fclose`, and `delete` on files. It's also a good abstraction if we need to update how we deel with files.

This class creates a temporary file with a unique name on its initialization. It has methods to return the `descriptor()`, `filename()` and the file opened itself as `FILE*` in `file_pointer(mode)` and as `ofstream` in `ofstream()`. Its destructor closes the file descriptor and removes the file created, avoiding memory leaks.

Fixes #738


test/binary/test_61.kore
```diff
@@ -0,0 +1,2 @@
+// RUN: %strip-binary -i %test-input -k -o %t.out
+// RUN: diff %t.out %test-diff-out
```
Baltoli, line 2: Perhaps add a comment here explaining why we're testing like this, and with large files specifically.

Baltoli (approved):

test/binary/test_61.kore
Robertorosmaninho: Done!

PR #763, opened 2023-05-15
Reorganizing `set -x` calls for verbose execution

### LINKED ISSUE #754: Support extra-verbose output from `llvm-krun`
Currently, `llvm-krun -v` only gives you a limited view of what's going on internally. If, for example, you're debugging an issue `kore-strip` or the other binary tools, the only way to get a hold of the commands being run is to patch `set -ex` into them.

Robertorosmaninho:
Fixes #754 

A simple PR  to improve the `--verbose` use on `llvm-krun` by outputting more steps of the execution process!

### DIFF
--- a/bin/llvm-krun
+++ b/bin/llvm-krun
@@ -69,7 +69,7 @@ Mandatory arguments to long options are mandatory for short options too.
   -i, --initializer INIT   Use INIT as the top cell initializer 
   -nm, --no-expand-macros  Don't expand macros in initial configuration
   -v, --verbose            Print major commands executed to standard error
-  -vv                      Print also commands executed by intermediate tools to
+  -vv                      Also print commands executed by intermediate tools to
                            standard error
   -vvv                     Print all commands executed by this script to
                            standard error
@@ -162,13 +162,13 @@ do
     ;;
 
     -vvv)
-    verbose+=3
-    shift;
+    verbose=$((verbose+=3))
+    shift
     ;;
 
     -vv)
-    verbose+=2
-    shift;
+    verbose=$((verbose+=2))
+    shift
     ;;
 
     -v|--verbose)
@@ -339,7 +339,6 @@ construct_binary_input()
     file_name="file_$param"
     binary_file_name="binary_$param"
 
-    set +e
     (
     if [ "$verbose" -ge 2 ]; then
       set -x
--- a/bin/llvm-krun
+++ b/bin/llvm-krun
@@ -162,18 +162,18 @@ do
     ;;
 
     -vvv)
-    verbose=$((verbose+=3))
+    verbose=$((verbose + 3))
     shift
     ;;
 
     -vv)
-    verbose=$((verbose+=2))
+    verbose=$((verbose + 2))
     shift
     ;;
 
     -v|--verbose)
-    verbose+=1
-    shift;
+    verbose=$((verbose + 1))
+    shift
     ;;
 
     -save-temps)
--- a/bin/llvm-krun
+++ b/bin/llvm-krun
@@ -204,7 +204,7 @@ for name in "${pretty[@]}"; do
   pretty_name="pretty_$name"
   set +e
   (
-  if [ $verbose -ge 1 ]; then
+  if [ "$verbose" -ge 1 ]; then
     set -x
   fi
   printf %s "${!pretty_name}" | "$dir/parser_$name" /dev/stdin > "$parser_file"
@@ -238,7 +238,7 @@ for name in "${params[@]}"; do
       temp_inputs+=("$temp_file")
 
       (
-      if [ $verbose -ge 2 ]; then
+      if [ "$verbose" -ge 2 ]; then
         set -x
       fi
       {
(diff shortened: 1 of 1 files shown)


Baltoli (changes requested):
This wasn't quite the intent of the original issue; my point was that the existing verbose mode is usually pretty much what you need to get a reproducible input example from the backend (by copying intermediate files in conjunction with `-save-temps`). However, where it falls down is when you're doing internal developer debugging on something like `kore-strip`. In these cases, we should add a flag `-vv` (or similar) that prints in greater detail the commands that are executed.

Additionally, I recommend reviewing your understanding of bash semantics - as written, this PR appears to move `set -x` calls outside of their original subshell contexts, and adds some in the root shell generally.

Baltoli (changes requested):
This wasn't quite the intent of the original issue; my point was that the existing verbose mode is usually pretty much what you need to get a reproducible input example from the backend (by copying intermediate files in conjunction with `-save-temps`). However, where it falls down is when you're doing internal developer debugging on something like `kore-strip`. In these cases, we should add a flag `-vv` (or similar) that prints in greater detail the commands that are executed.

Additionally, I recommend reviewing your understanding of bash semantics - as written, this PR appears to move `set -x` calls outside of their original subshell contexts, and adds some in the root shell generally.

bin/llvm-krun
```diff
@@ -298,40 +326,76 @@ HERE
...
     echo "Lbl'Stop'Map{}()" | $convert_term >> "$input_file"
+    )
   fi
 
   for param in "${params[@]}"; do
     var_name="params_$param"
     file_name="file_$param"
     binary_file_name="binary_$param"
 
+    set +e
```
dwightguth, line 342: Why was this line added?

bin/llvm-krun
```diff
@@ -156,8 +161,18 @@ do
     shift; shift
     ;;
 
+    -vvv)
+    verbose+=3
```
Baltoli, line 165: This line doesn't do what you think it does; this will concatenate the string `3` onto the _string_ value currently stored in `verbose`. For example:

**`test.sh`**
```bash
#!/usr/bin/env bash
verbose=0
verbose+=1
echo $verbose
```

```console
$ ./test.sh
01
```

The correct solution is to use shell [arithmetic expansion](https://tldp.org/LDP/abs/html/arithexp.html)

bin/llvm-krun
```diff
@@ -156,8 +161,18 @@ do
...
+    shift;
```
Baltoli, line 166: You don't need the semicolon here
```suggestion
    shift
```

Baltoli (changes requested):
Mostly looks good but a couple of small changes needed - thanks @Robertorosmaninho !

bin/llvm-krun
Robertorosmaninho: my bad

bin/llvm-krun
```diff
@@ -67,7 +68,11 @@ Mandatory arguments to long options are mandatory for short options too.
       --depth INT          Execute up to INT steps
   -i, --initializer INIT   Use INIT as the top cell initializer 
   -nm, --no-expand-macros  Don't expand macros in initial configuration
-  -v, --verbose            Print commands executed to standard error
+  -v, --verbose            Print major commands executed to standard error
+  -vv                      Print also commands executed by intermediate tools to
+                           standard error
```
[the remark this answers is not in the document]
Robertorosmaninho: Thanks

bin/llvm-krun
Robertorosmaninho: My bad, so sorry!

bin/llvm-krun
Robertorosmaninho: Thanks

bin/llvm-krun
```diff
@@ -156,8 +161,18 @@ do
...
+
+    -vv)
+    verbose=$((verbose+=2))
+    shift
+    ;;
+
     -v|--verbose)
-    verbose=1
+    verbose+=1
     shift;
```
Baltoli, line 176:
```suggestion
    shift
```

bin/llvm-krun
```diff
@@ -156,8 +161,18 @@ do
...
+    ;;
+
+    -vv)
+    verbose=$((verbose+=2))
+    shift
+    ;;
+
     -v|--verbose)
-    verbose=1
+    verbose+=1
```
Baltoli, line 175: This line still has the error I reviewed previously

bin/llvm-krun
```diff
@@ -156,8 +161,18 @@ do
     shift; shift
     ;;
 
+    -vvv)
+    verbose=$((verbose+=3))
+    shift
+    ;;
+
+    -vv)
+    verbose=$((verbose+=2))
```
Baltoli, line 170:
```suggestion
    verbose=$((verbose + 2))
```

bin/llvm-krun
```diff
@@ -156,8 +161,18 @@ do
     shift; shift
     ;;
 
+    -vvv)
+    verbose=$((verbose+=3))
```
Baltoli, line 165:
```suggestion
    verbose=$((verbose + 3))
```

Baltoli (changes requested):

Baltoli (approved):

Issue #757, opened 2023-05-12
Add Python binding for desugaring KORE terms


tothtamas28:
I.e. to recursively eliminate `left-assoc` and `right-assoc`.

tothtamas28:
A few test inputs from the KEVM conformance test suite: [tests.zip](https://github.com/runtimeverification/llvm-backend/files/11477519/tests.zip)

PR #764, opened 2023-05-15
Desugar `\{left,right}-assoc`

Baltoli:
This has historically been handled by the KORE parser, but it would be useful for Pyk and the C bindings to be able to desugar _programmatically-constructed_ input.

In this PR, a new `desugarAssociative` method is added to the KORE Pattern AST hierarchy; this method expands associative syntax sugar into nested calls using the same logic as the parser (except that the parser will not recurse into subterms when doing so).

Testing of the method is done through the C bindings, which programatically construct a term with a `\left-assoc` symbol, desugar it, then evaluate it. No explicit python test is added as it's awkward to construct these terms without going through the parser; Pyk tests of this feature + the C bindings test will be sufficient coverage.

Fixes: https://github.com/runtimeverification/llvm-backend/issues/757


tothtamas28 (approved):
Tried this on some large inputs, works amazingly well.

PR #765, opened 2023-05-15
Try to include all variables in apply_rule functions, even if they are unused

dwightguth:
Previously we were only passing variables to the apply_rule function if they were actually used by the RHS of that rule. That is the only case where they are needed normally, but during proof object generation we need the variables that appeared only on the LHS also in order to be able to instantiate the axiom with the value of all its variables in the resulting proof object. Thus we need to pass all of this information into the proof hint, which will be inserted into the apply_rule function.

As a result, we need to be able to pass all of these variables to these functions even though they are currently unused for anything. This involves the following changes:

* make sure all variables in the rule are included in the decision tree generation algorithm in the Leaf nodes it generates
* create a getLeftHandSide function in C++ analogous to the getRequires and getRightHandSide functions, which is used to:
* ensure that all variables are part of the parameters of apply_rule functions during code generation

Still working on testing this, so don't review yet.

### DIFF
--- a/include/kllvm/ast/AST.h
+++ b/include/kllvm/ast/AST.h
@@ -749,6 +749,7 @@ public:
   bool isTopAxiom();
   bool isClaim() { return _isClaim; }
   KOREPattern *getRightHandSide() const;
+  std::vector<KOREPattern *> getLeftHandSide() const;
   KOREPattern *getRequires() const;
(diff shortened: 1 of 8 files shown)


dwightguth:
Closing this for now; we will reopen it again later.

PR #769, opened 2023-05-18
Eliminate pointer equality tests

Baltoli:
This PR makes a necessary (but insufficient) change that gets us one step closer to supporting LLVM 16 (and opaque pointers).

The specific change made is to eliminate pointer type equality checks (`A* == B*`) as their [semantics will silently change in LLVM 16](https://github.com/runtimeverification/llvm-backend/issues/730). The fixed version has semantics that remain valid for versions < 16, so this is not an intrusive change.

While this PR doesn't get us all the way to LLVM 16 support, I'd like to get it merged to avoid the change staying open for much longer.


Robertorosmaninho (approved):
LGTM! The `isCollectionSort` function seems much more readable than the simple comparing with the previous comparison!

Issue #768, opened 2023-05-17
Issue with `%check-dir-diff` lit directive


mariaKt:
There seems to be an issue with the `%check-dir-diff` check. Specifically, it seems to be able to report a failure in a test with multiple inputs only if the last input to run fails. I.e., if a KORE test has inputs in1, in2, in3, and failure occurs in in3, the test will be reported as a FAIL. If in2 fails, the run will continue and if in3 succeeds, the test will be reported as a PASS.

https://github.com/runtimeverification/llvm-backend/blob/046b5968778c4cb1fa8bc5b69918248e41504bb4/test/lit.cfg.py#L99

Baltoli:
I confirmed that this is indeed the case; will investigate. Thanks @mariaKt!

PR #772, opened 2023-05-24
Fix broken test mechanism

Baltoli:
The previous way of writing this test stanza would allow for a single input to these definitions to fail, while reporting that the test itself was passing overall.

This PR refactors the test to ensure that failures are properly reported. I have inspected and validated that:
* No tests were silently failing as a result of the old code.
* A test that I would expect to fail does fail with the new directive, when it would have silently passed before.

Fixes https://github.com/runtimeverification/llvm-backend/issues/768


Robertorosmaninho (approved):
LGTM! The if comparison using the value of `"$?"` seems much clear to me than the previous strategy.

PR #773, opened 2023-05-24
Add K references for bindings tests

Baltoli:
Mirrors https://github.com/runtimeverification/llvm-backend/pull/745, but for bindings tests rather than the main execution test suite.

No changes to the tests are made; all the relevant K sources were commented into these files beforehand, and were written recently enough that there was no need to make changes to account for newer K versions.


Robertorosmaninho (approved):
LGTM! Awesome work, @Baltoli! Thanks!

Issue #770, opened 2023-05-19
Potential memory leak when using Python bindings


tothtamas28:
When running concrete interpretation thousands of times from a single process using the Python bindings, the system eventually runs out of memory.

## Reproduction

https://github.com/runtimeverification/evm-semantics/pull/1816

1. Clone `evm-semantics`
3. Check out `llvm-interpret` 
4. Update submodules
5. Build: `make poetry deps build`
6. Run tests: `make -C ./kevm-pyk test-integration`
7. Observe memory consumption

ehildenb:
Could try different test-suites independently. The VMTests are run with `make test-vm`, so maybe pytest could separate that ouut, and the bchain tests are called with `make test-bchain`. These test-suites are in different directories, so that can probably be used to filter which tests.

If we can isolate those two test-suites, that would be a hint that probably the memory leak is happening in `CALL*` opcodes. Those opcodes do a lot of bytes manipulation (the bchain tests which have `CALL*` opcodes), which may be a hint.

Baltoli:
My thought here is that this is a manifestation of the issue described in this blog post: https://ayende.com/blog/195681-B/production-postmortem-the-memory-leak-that-only-happened-on-linux, where `posix_memalign` doesn't play ball with another allocator (here, the Python heap allocation subsystem), producing fragmentation and leaks.

The blog post describes a mechanism via `mmap()` to solve the issue; I have yet to investigate whether this will work in our use case, and I'm wary of changing details of the backend's memory subsystem without substantial testing.

As a possible workaround, I have exposed an explicit call to the backend GC to the Python bindings (https://github.com/runtimeverification/llvm-backend/pull/776). If we call this after each test, it may reduce the pressure to allocate new megabyte blocks in the backend and therefore decrease fragmentation and leaks.

Baltoli:
Summary of discussion 5/6:

* @dwightguth agrees that the issue is likely to be with `posix_memalign` re. fragmentation and overallocation.
* The garbage collector's core algorithmic requirement is that the core block allocator hands over blocks of 1MB aligned memory; this allows it to do fast block detection by rounding down addresses.
* The solution here is likely to be to roll our own version of the functionality provided by `posix_memalign` using `mmap`; we can map a 16MB region and align it to a megabyte boundary by hand.

Baltoli:
[...]

PR #774, opened 2023-05-25
Fix minor memory leaks

Baltoli:
While debugging https://github.com/runtimeverification/llvm-backend/issues/770, I found a few small memory leaks in the C bindings and associated tests. This PR cleans up places where we were not properly freeing temporary objects.

The fixes are unrelated to the Python bindings, but allow me to check that nothing obvious is being leaked by the underlying associative desugaring code.


tothtamas28 (approved):

tothtamas28:
`pyk` integration tests are passing.

Robertorosmaninho (approved):

PR #775, opened 2023-05-25
Improving `get_ui` error message

### LINKED ISSUE #771: Improve reporting of errors with `get_ui`
We should report hook errors more specifically by passing the name of the calling function through to `get_ui` when we use it; this will help people produce specific bug reports. I think the solution is to turn `get_ui` into a macro that expands `__pretty_function__` and calls a similar underlying `get_ui_named` implementation.

Robertorosmaninho:
Fixes #771 

This simple PR modifies the `get_ui` function to `get_ui_named` while keeping its interface the same as the previous one. The new function receives a string representing the name of the caller function that will be used to improve the error message if the integer has an overflow!

### DIFF
--- a/runtime/strings/bytes.cpp
+++ b/runtime/strings/bytes.cpp
@@ -71,7 +71,7 @@ SortInt hook_BYTES_bytes2int(
   return move_int(result);
 }
 
-unsigned long get_ui_named(mpz_t i, std::string const& caller) {
+unsigned long get_ui_named(mpz_t i, std::string const &caller) {
   if (!mpz_fits_ulong_p(i)) {
     std::string const error_msg
         = "Integer overflow from " + caller + ": " + intToString(i);
--- a/runtime/strings/bytes.cpp
+++ b/runtime/strings/bytes.cpp
@@ -10,6 +10,8 @@
 
 extern "C" {
 
+#undef get_ui
+#define get_ui(x) get_ui_named(x, __PRETTY_FUNCTION__)
 #define KCHAR char
 
 mpz_ptr move_int(mpz_t);
@@ -69,9 +71,10 @@ SortInt hook_BYTES_bytes2int(
   return move_int(result);
 }
 
-unsigned long get_ui(mpz_t i) {
+unsigned long get_ui_named(mpz_t i, std::string caller) {
   if (!mpz_fits_ulong_p(i)) {
-    KLLVM_HOOK_INVALID_ARGUMENT("Integer overflow");
+    KLLVM_HOOK_INVALID_ARGUMENT(
+        "Integer overflow from " + caller + ": ", intToString(i));
   }
   return mpz_get_ui(i);
 }
--- a/runtime/strings/bytes.cpp
+++ b/runtime/strings/bytes.cpp
@@ -11,7 +11,7 @@
 extern "C" {
 
 #undef get_ui
-#define get_ui(x) get_ui_named(x, __PRETTY_FUNCTION__)
+#define get_ui(x) get_ui_named(x, __func__)
 #define KCHAR char
 
 mpz_ptr move_int(mpz_t);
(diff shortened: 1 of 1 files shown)


Robertorosmaninho:
@Baltoli, I couldn't really reproduce the bug in `kevm`, but I would like to add a unit test for this here. Any ideas?

runtime/strings/bytes.cpp
```diff
@@ -69,9 +71,10 @@ SortInt hook_BYTES_bytes2int(
   return move_int(result);
 }
 
-unsigned long get_ui(mpz_t i) {
+unsigned long get_ui_named(mpz_t i, std::string caller) {
```
Baltoli, line 74: We don't need to copy this argument:
```suggestion
unsigned long get_ui_named(mpz_t i, std::string const& caller) {
```

You could (if you were so inclined) even make the argument type `char const*` and avoid constructing a `std::string` from the literals being passed in, but the const-ref is probably fine here.

runtime/strings/bytes.cpp
```diff
@@ -69,9 +71,10 @@ SortInt hook_BYTES_bytes2int(
...
   if (!mpz_fits_ulong_p(i)) {
-    KLLVM_HOOK_INVALID_ARGUMENT("Integer overflow");
+    KLLVM_HOOK_INVALID_ARGUMENT(
+        "Integer overflow from " + caller + ": ", intToString(i));
```
Baltoli, line 77: This macro does variadic formatting via a call to `fmt::format`, so we should avoid doing string concatenation to build its argument. See the implementation here: https://github.com/runtimeverification/llvm-backend/blob/571667745aefdcfb4820ba13f3c3f5dcabe12489/include/runtime/fmt_error_handling.h#L6

runtime/strings/bytes.cpp
```diff
@@ -10,6 +10,8 @@
 
 extern "C" {
 
+#undef get_ui
+#define get_ui(x) get_ui_named(x, __PRETTY_FUNCTION__)
```
Baltoli, line 14: My bad here when specifying the issue; we probably want to use `__func__` rather than `__PRETTY_FUNCTION__` as the latter is an extension rather than part of the standard.

Baltoli (changes requested):

Baltoli:
> @Baltoli, I couldn't really reproduce the bug in `kevm`, but I would like to add a unit test for this here. Any ideas?

I would personally be happy leaving this without a unit test; to trigger the behaviour without a K definition you'd need to:
* Use the `mpz` API to construct an arbitrary-precision integer with a value not representable in a 64-bit unsigned integer.
* Call one of the hooks that uses `get_ui` to trigger the behaviour.
* Verify that the exception thrown has the function name you expect to see in its message string.

runtime/strings/bytes.cpp
```diff
@@ -69,9 +71,11 @@ SortInt hook_BYTES_bytes2int(
   return move_int(result);
 }
 
-unsigned long get_ui(mpz_t i) {
+unsigned long get_ui_named(mpz_t i, std::string const &caller) {
   if (!mpz_fits_ulong_p(i)) {
-    KLLVM_HOOK_INVALID_ARGUMENT("Integer overflow");
+    std::string const error_msg
+        = "Integer overflow from " + caller + ": " + intToString(i);
+    KLLVM_HOOK_INVALID_ARGUMENT(error_msg);
```
Baltoli, line 78: This is still performing string concatenation and constructing an unnecessary `std::string` temporary. See [here](https://github.com/runtimeverification/llvm-backend/blob/571667745aefdcfb4820ba13f3c3f5dcabe12489/runtime/arithmetic/int.cpp#L156-L157) for an example of how to use the variadic formatting macro.

Baltoli (changes requested):

runtime/strings/bytes.cpp
Robertorosmaninho: Thanks for pointing that out! Just fixed it!

Baltoli (approved):

PR #776, opened 2023-05-26
Expose GC to python bindings

Baltoli:
This PR adds a function `free_all_kore_memory` to the KLLVM runtime module. Calling this function will run the garbage collector once, freeing backend terms that are no longer live.

Calls are added to the existing Python test that takes execution steps to ensure that we can call the function properly.

### DIFF
--- a/bindings/python/runtime.cpp
+++ b/bindings/python/runtime.cpp
@@ -11,7 +11,7 @@ namespace py = pybind11;
 using namespace kllvm;
 
 extern "C" {
-void freeAllKoreMem(void);
+void resetMemory(void);
 }
 
 /*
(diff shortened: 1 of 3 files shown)


tothtamas28:
I tried using `free_all_kore_memory` in https://github.com/runtimeverification/evm-semantics/pull/1816 like this:

```python3
def interpret(gst_data: Any, schedule: str, mode: str, chainid: int) -> Pattern:
    init_kore = gst_to_kore(gst_data, schedule, mode, chainid)
    init_llvm = kore_to_llvm(init_kore).desugar_associative()
    final_llvm = runtime.interpret(init_llvm)
    final_kore = llvm_to_kore(final_llvm)
    runtime.free_all_kore_memory()
    return final_kore
```

Memory still leaks significantly. I'm I using it wrong?

Baltoli:
Nope, that's my expected usage. I guess we're already garbage collecting frequently enough but are still ending up with fragmentation. Let me have a think about this some more

PR #780, opened 2023-06-02
Deserialize pattern variables correctly

### LINKED ISSUE #779: Serialize + deserialize pattern variables properly to binary KORE
Currently, these can be serialized (unclear if it's _correct_) but not deserialized. This should be fixed for the Pyk integration to work fully.

Baltoli:
There was a missing case in the binary deserializer that meant we couldn't deserialize pattern variables (`X : Sort`) properly. This PR implements, tests and documents the missing case.

No change to the format version is required as the serializer was already handling this case properly, and all existing files that were previously valid remain so.

Fixes #779

### DIFF
--- a/docs/binary-kore.md
+++ b/docs/binary-kore.md
@@ -107,6 +107,14 @@ Composite patterns encode each of their arguments (recursively, following this
 schema), then the constructor symbol (as described above). These are followed by
 the byte `04`, and a variable-length arity for the pattern.
 
+Pattern variables encode their sort argument, followed by the byte `09`, then a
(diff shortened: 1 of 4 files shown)


Robertorosmaninho:
In a k source level this is related to records uses like `foo(x:Int)`?

Robertorosmaninho (approved):

Robertorosmaninho:
Thanks for the clarification @Baltoli!

tothtamas28 (approved):
Works with the `pyk` tests.

PR #785, opened 2023-06-07
Delimited binary

### LINKED ISSUE #781: Expose more permissive API for config deserializer
Motivated by the in-development format being used for the proof generation prototype, which embeds binary KORE objects as part of a larger file format.

We would like to expose a nicer API to the Python bindings, so that a _portion_ of a file-like object can be consumed to return a `_kllvm.KorePattern`, leaving the file at the end of the byte sequence representing that pattern.

I think this will require a bit of digging into the Python protocols for file-like objects to get a Pybind version working. Additionally, a clean implementation probably wants to bump the format version to include an extra bit of _size_ metadata in the files. We can keep version `1.1.0` backwards compatible when doing so.

Example Python code I might expect to see:
```python
with open('in.proof', 'b') as f:
    ... = f.read(...)
    term = pyk.kllvm.read_pattern(f)
    ... = f.read(...)
```

cc @gtrepta

Baltoli:
This PR extends the existing binary format for KORE terms to include an optional length field, which encodes the size in bytes of the serialized pattern. This is motivated by https://github.com/runtimeverification/proof-generation/issues/39, which embeds binary KORE terms into a larger file. It is desirable here to be able to read a term from a file pointer without knowing its length up front.

The changes made in this PR are as follows:
* Bump the binary KORE format to version `1.2.0`. This new version adds an 8-byte field immediately following the version bytes, which if non-zero, encodes the length of the subsequent term.
* Update the serializer and deserializer to accommodate this field, including version checks for backwards-compatibility.
* Update `kore-convert`, `kore-strip` etc. for compatibility.
* Add a Python binding that reads from a file-like object, rather than from a preloaded bytes object.
* Add Python-side tests for the new binding, and update existing tests to exercise the new code paths.

These changes are backwards-compatible; terms serialized using version `1.1.0` or older can still be read, and the interface to the existing command-line tools and concatenation workflow does not change. I have tested this on the C semantics, which makes extensive use of the binary format.

Closes #781


nishantjr:
Thanks @Baltoli. The python interface looks good to me, but I don't have the expertise to review the backend changes, so I'll leave it to the other reviewers to approve.

dwightguth (approved):

PR #787, opened 2023-06-15
Fix for small papercuts in hook interface

Baltoli:
While working on a minimal hook implementation, I identified the following papercuts in usage of the LLVM backend:
* `size_t` is not defined when including `alloc.h`
* Users need to define their own `dotK` block-getter - this is a common enough task we should put it in the backend.

### DIFF
--- a/include/runtime/alloc.h
+++ b/include/runtime/alloc.h
@@ -1,6 +1,8 @@
 #ifndef ALLOC_H
 #define ALLOC_H
 
+#include <cstddef>
+
 extern "C" {
 
 // The maximum single allocation size in bytes.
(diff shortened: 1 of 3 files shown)


Robertorosmaninho (approved):
LGTM!

gtrepta (approved):

dwightguth:
Code looks great. @Baltoli You might want to do a follow-up change where you look for places where that function could replace existing code though.

Baltoli:
Yep, I know the C semantics has an implementation at the very least, and I'll dig through for any other places. Thanks!

dwightguth:
I'm reasonably certain there already exists some in the llvm backend codebase itself. You might consider looking in the frontend for hooked functions in the prelude that return sort K and then mapping those back to their implementation in the backend

dwightguth:
It also might be worth using a local static variable to cache the return value of the function, btw

Issue #795, opened 2023-07-07
fastcc -> tailcc


Baltoli:
Now that we no longer need to support older versions of LLVM, we can convert all our usages of `fastcc` to `tailcc` so that the `tailcallopt` flag in the LLVM plumbing is no longer needed.

PR #796, opened 2023-07-07
Replacing fastcc with tailcc

Robertorosmaninho:
This PR replaces the LLVM `CallingConv` from `Fast` to `Tail`. By doing so, we no more need to call the `tailcallopt` optimization pass and decrease the kompilation time.

Fixes: #795

### DIFF
--- a/bin/llvm-kompile-clang
+++ b/bin/llvm-kompile-clang
@@ -149,7 +149,7 @@ fi
 # straight to invoking the C++ compiler.
 if [ "$main" != "python_ast" ]; then
   if [ "$lto" = "lto" ]; then
-    flags+=("-flto" "-Wl,-mllvm,-tailcallopt")
+    flags+=("-flto" "-Wl,-mllvm")
     files=("$LIBDIR"/llvm/*.ll)
     if ! $link; then
       mv "$modopt" "$output_file"
@@ -162,7 +162,7 @@ if [ "$main" != "python_ast" ]; then
         modasm="$output_file"
       fi
       run @LLC@ \
-        -tailcallopt "$modopt" -mtriple=@BACKEND_TARGET_TRIPLE@ \
+        "$modopt" -mtriple=@BACKEND_TARGET_TRIPLE@ \
         -filetype=obj "$llc_opt_flags" "${llc_flags[@]}" -o "$modasm"
       modopt="$modasm"
     fi
@@ -170,7 +170,7 @@ if [ "$main" != "python_ast" ]; then
       for file in "$LIBDIR"/llvm/*.ll; do
         tmp="$tmpdir/$(basename "$file").o"
         run @LLC@ \
-          -tailcallopt "$file" -mtriple=@BACKEND_TARGET_TRIPLE@ \
+          "$file" -mtriple=@BACKEND_TARGET_TRIPLE@ \
           -filetype=obj "$llc_opt_flags" "${llc_flags[@]}" -o "$tmp"
         files+=("$tmp")
       done
--- a/bin/llvm-kompile-clang
+++ b/bin/llvm-kompile-clang
@@ -149,7 +149,7 @@ fi
 # straight to invoking the C++ compiler.
 if [ "$main" != "python_ast" ]; then
   if [ "$lto" = "lto" ]; then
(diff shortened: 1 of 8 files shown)


lib/codegen/CreateTerm.cpp
```diff
@@ -1127,7 +1127,7 @@ bool makeFunction(
       debugName, debugName,
       getDebugFunctionType(getDebugType(returnCat, Out.str()), debugArgs),
       definition, applyRule);
-  if (fastcc) {
+  if (tailcc) {
     applyRule->setCallingConv(llvm::CallingConv::Fast);
```
Baltoli, line 1131: This will need to be changed as well!

bin/llvm-kompile-clang
```diff
@@ -149,7 +149,7 @@ fi
 # straight to invoking the C++ compiler.
 if [ "$main" != "python_ast" ]; then
   if [ "$lto" = "lto" ]; then
-    flags+=("-flto" "-Wl,-mllvm,-tailcallopt")
+    flags+=("-flto" "-Wl,-mllvm")
```
Baltoli, line 152:
```suggestion
    flags+=("-flto")
```

The flags `-Wl,-mllvm,-tailcallopt` are all connected and should therefore be removed together. When passed to `clang++`, `-Wl` states "pass everything following a comma to the linker's command line", and similarly when the _linker_ sees `-mllvm` it passes everything following the comma to the LLVM-internal options processing.

The net effect of this nested flag is to pass `-tailcallopt` to the LLVM-internal components that previously dealt with translating `fastcc` calls to `tailcc`.

include/kllvm/codegen/CreateTerm.h
```diff
@@ -47,20 +47,24 @@ class CreateTerm {
...
+      - args: the arguments to pass to the functgion
+      - sret: if true, this is a function that returns a struct constant via the
+              C abi, ie, the function actually returns void and the return value
+              is via a pointe. Note that this can be set to true even if the
+              function does not return a struct, in which case its value is
+              ignored.
+      - load: if the function returns a struct via sret, then if load is true,
+              we load the value
+      - tailcc: true if we should use the tailcc calling convention returned
+                from the function before returning it.*/
```
Baltoli, line 64:
```suggestion
     certain ABI and calling convention details:
      - name: the name of the function to call
      - returnCat: the value category of the return type of the function
      - args: the arguments to pass to the function
      - sret: if true, this is a function that returns a struct constant via the
              C ABI (that is, the function actually returns void and the return value
              is via a pointer). Note that this can be set to true even if the
              function does not return a struct, in which case its value is
              ignored.
      - load: if the function returns a struct via sret and load is true,
              we load the value on return.
      - tailcc: true if the call should be made via the tailcc calling convention.
*/
```

include/kllvm/codegen/CreateTerm.h
```diff
@@ -47,20 +47,24 @@ class CreateTerm {
      specified definition, and returns the value itself, along with a boolean
      indicating whether the resulting term could be an injection. */
   std::pair<llvm::Value *, bool> operator()(KOREPattern *pattern);
+
   /* creates a call instructin calling a particular llvm function, abstracting
```
Baltoli, line 51:
```suggestion
  /* creates a call instruction calling a particular LLVM function, abstracting
```

lib/codegen/Decision.cpp
```diff
@@ -1070,6 +1070,7 @@ void makeStepFunction(
   codegen(dt);
 }
 
+// TODO: As we're not using fastcc, we must check if we can remove this wrapper
```
Baltoli, line 1073:
```suggestion
```

I suspect that the same issues you experienced with `fastcc` will occur again when using `tailcc`. The LLVM [language specification](https://llvm.org/docs/LangRef.html#calling-conventions) states that `tailcc` is equivalent to `fastcc` except with the addition of guaranteed tail calls. The ABI changes that break GDB et. al. will therefore still be present with your changes here.

Baltoli (approved):
A couple of minor comments; please feel free to merge when they're applied. Thanks @Robertorosmaninho!

PR #803, opened 2023-07-14
Garbage Collection for range maps

mariaKt:
This PR implements garbage collection for range maps.

The current implementation of the range map's underlying data structure, the red-black tree, is using shared pointers to the internal nodes and leafs of the tree (i.e., the pointer to the root and pointers to left and right children are `shared_ptr`). This means that we do not need to manage those internal nodes and leaf structures as part of the llvm backend's garbage collection. However, the contents of the map (i.e., the key and value `KItem`s) should be migrated during garbage collection when we migrate an object that contains a range map as a child. We do that by traversing said range map at the time of migration of the parent object, and migrating every such key or value `KItem` while updating the pointer to it on the corresponding internal node of leaf structure.

We include a KORE test that is based on extending IMP with keywords for certain range map operations. The inputs are currently small IMP programs with a few different interleavings of update and remove operations. We will use the C semantics (whose memory cell, currently a map, will be replaced by a range map) and its test suite for more extensive testing after we complete that implementation.

Note: This approach may lead to memory fragmentation/leaks for large programs. Data pointed to by memory managed by the C++ runtime (shared pointers) may not be deleted/deallocated, but instead be overwritten after the llvm garbage collection is completed. This means that reference counting will not be updated for these shared objects, thus they may never be returned to the system.
If we observe this to be an issue, we can address it in a future PR by changing the implementation of the red-black tree to not rely on shared pointers but instead use memory managed by the llvm backend (kore-alloc) for the internal nodes and leafs as well, and migrate these additionally during garbage collection.


runtime/collect/migrate_collection.cpp
```diff
@@ -116,3 +116,20 @@ void migrate_map(void *m) {
   migrate_collection_node((void **)&impl.root);
   migrate_champ_traversal(impl.root, 0, migrate_map_leaf);
 }
+
+using treemap = rb_tree::RBTree<rng_map::Range<KElem>, KElem>;
+void migrate_treemap(treemap t) {
```
Baltoli, line 121: Looking back at the definition of `RBTree`, this copy is safe because the only state is a shared pointer - would this signature be more clear as `treemap& ` perhaps?

Baltoli (approved):
One minor comment; tests + code look great otherwise. Nice work @mariaKt!

runtime/collect/migrate_collection.cpp
mariaKt: Given that at the call sites
https://github.com/runtimeverification/llvm-backend/blob/e75ea3cadbfaa836c6775c28f1e37158ddd01360/runtime/collect/migrate_collection.cpp#L129
and
https://github.com/runtimeverification/llvm-backend/blob/e75ea3cadbfaa836c6775c28f1e37158ddd01360/runtime/collect/migrate_collection.cpp#L130
and
https://github.com/runtimeverification/llvm-backend/blob/e75ea3cadbfaa836c6775c28f1e37158ddd01360/runtime/collect/migrate_collection.cpp#L134
the function is called with a temporary value, I would have to assign it to a local and then pass that by reference. So effectively, we wouldn't be avoiding any copies.
If you still prefer that, let me know and I can go ahead and make the change.

dwightguth (approved):
pretty simple. I like it! As you say, if there are issues with performance regarding the shared pointers we can create a followup PR to deal with those.

PR #822, opened 2023-08-09
Test `kore_simplify` with KEVM's `definition.kore` on a real-world input

geo2a:
This is somewhat minimized example that should exhibit a memory leak.

in KLLVM when used via the C bindings.

## Getting the input

Please download [`llvm_calls_private.c`](https://drive.google.com/file/d/1jCn68dK10xULpJtY2jjQNDqIMgEse0GH/view?usp=drive_link) from RV Shared Google Drive, as it contains confidential bytecode and place it instead of the stub `llvm_calls.c` file. The stub file and the real file only differ in the bytecode token. 

## Running the test

The test should be compiled and ran as follows:

```
lit -av test/c/test_evm_leak.kore
```

Currently, this fails with

```
test/c/Output/test_evm_leak.kore.tmp.dir/libtest.so: undefined symbol: hook_KRYPTO_ecdsaPubKey/home/geo2a/Workspace/RV/llvm-backend/test/c/Output/test_evm_leak.kore.script: line 4: 2105633 Aborted                 (core dumped) 
```

because the call to `llvm-kompile-testing` needs to be changed to include hooks form `blockcahin-k-plugin`, and I don't have the expertise to do it in this test harness' setup.

## Context

Within `kore-rpc-booster`, we make hundreds of calls like this one while processing one execution request.

### DIFF
--- a/test/c/Inputs/evm.c
+++ b/test/c/Inputs/evm.c
@@ -1,0 +1,15 @@
+#include "api.h"
+
+#include <stdio.h>
+
+int main(int argc, char **argv) {
+  if (argc <= 1) {
+    return 1;
+  }
+
+  struct kllvm_c_api api = load_c_api(argv[1]);
+
+  api.kllvm_init();
+
+#include "llvm_calls.c"
+}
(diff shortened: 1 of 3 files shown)


geo2a:
The issue has been resolved with https://github.com/runtimeverification/hs-backend-booster/pull/263. Thanks everyone involved!

PR #826, opened 2023-08-16
fix llvm-kompile-compute-loc and llvm-kompile-compute-ordinal

dwightguth:
We were previously incorrectly assuming that all axioms in a kore file were indented exactly two spaces, which is not correct. This led to the ordinals and locations being generated being incorrect. This ought to fix that.


Robertorosmaninho (approved):

Scott-Guest (approved):
LGTM

Issue #824, opened 2023-08-11
Python binding for `StringPattern.contents` is broken


tothtamas28:
Consider the following small script:

```python3
from pyk.kllvm import load
from pyk.kllvm.parser import parse_text

ast = parse_text(r'"\x80"')
print(ast.contents)
```

This raises the following exception:

```
Traceback (most recent call last):
  File "/home/ttoth/git/pyk/test.py", line 5, in <module>
    print(ast.contents)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x80 in position 0: invalid start byte
```

A quick solution would be to expose `contents` as `bytes` (under perhaps a different name like `raw` or `raw_contents`).
Ideally though, the bindings should take care of the conversion to a `str`, otherwise, this has to be done on the Python side, which is obviously much slower.

Without this change, a workaround in `pyk` is to parse the result of `__str__()`.

Scott-Guest:
The changes I'm making for Unicode support in the LLVM backend (#3344) will fix this without having to change the bindings.

Specifically,`\xHH` will always be stored as as (the UTF-8 encoding of) `U+00HH` rather than the raw byte `0xHH`.

Scott-Guest:
Closing this as a duplicate of runtimeverification/k#3344

PR #828, opened 2023-08-17
Improve handling of Unicode vs Byte strings

Scott-Guest:
Part of runtimeverification/k#3344

Fixes the conflation between String and Bytes with regards to how escapes are handled. Specifically, now

- `\xHH` is always treated as `U+00HH` at parse time / when stored in a `KOREStringPattern`. 
    - For Bytes, we later produce the actual `char[]` of bytes when the runtime token is constructed
    - This ensures `KOREStringPattern::contents` is always a UTF-8 encoded string and aligns with how Pyk represents Bytes (resolving runtimeverification/llvm-backend#824)
- Created a separate `SortCategory::Bytes` and `BYTES_LAYOUT`
- The runtime representation for strings now uses the bit just before the length to mark whether it represents a UTF-8 encoded string or a byte string.
- We correctly print Unicode escapes (`\x`, `\u`, `\U`) versus individual byte escapes (only `\x`) depending on the sort.

A few other minor fixes along the way:
- Fixed some undefined behavior in `sfprintf` by ensuring `va_list`s are copied rather than re-used
- Renamed `set_len` to `init_with_len` because it clears also all non-length bits (got bit by this :smile:)
- Re-ordered `SortCategory` to ensure `MInt` is last so that `valueType.cat + valueType.bits` uniquely identifies a `ValueType`

Note that this just fixes the infrastructural issues - a few String hook algorithms still need to be updated to give correct results with non-ASCII characters.


cmake/RuntimeConfig.cmake
```diff
@@ -18,6 +18,7 @@ set(VARIABLE_BIT  0x8000000000000)
 set(LAYOUT_OFFSET 54)
 set(TAG_MASK      0xffffffff)
 set(LENGTH_MASK   0xffffffffff)
+set(IS_BYTES_BIT  0x10000000000)
 
 if(CMAKE_BUILD_TYPE STREQUAL "GcStats")
   set(HDR_MASK -18013298997854209) # 0xffc000ffffffffff
```
dwightguth, line 24: Need to add the macro for the sort category further down in this file.

dwightguth (changes requested):
The code looks good but it's incomplete. I can tell you tried to update every place where we were switching on the sort category, but you didn't get them all. If you see my comment above, we are defining cmake variables corresponding to each sort category and then turning them into a macro in `config/macros.h`. You need to make sure you update every place where those macros occur also. I suspect if this PR were merged in its current form, it would trigger regressions in garbage collection. I would recommend trying to create a test that exercises that regression and adding it to the test suite. It should just be a matter of having a term of sort Bytes live in the configuration during a minor collection cycle.

dwightguth (approved):

PR #830, opened 2023-08-23
fix bug in kore_block_get_bool

dwightguth:
We discovered a bug in the kore_block_get_bool function. It was reading 8 bytes and then casting the result to a bool, but a block containing a boolean as a child has one byte that is determined and 7 unspecified bytes that contain padding. As a result, if the padding was nonzero, the function would return true even if the injection actually contained the boolean `false`. This should fix that bug by casting the pointer to `bool*` before reading.


Scott-Guest (approved):
LGTM, nice catch

PR #793, opened 2023-07-07
Add `--profile` option to `llvm-kompile`

Baltoli:
This PR makes some minor refactorings to the `llvm-kompile-*` scripts, and adds an additional profiling option that (in verbose mode) prints the time taken by each step of the compilation pipeline.

I will follow this PR up with additional profiling information in the code generator.

For a small example:
```console
$ llvm-kompile test-kompiled/definition.kore test-kompiled/dt main -v --profile
+ /home/bruce/code/llvm-backend/build/install/bin/llvm-kompile-codegen test-kompiled/definition.kore test-kompiled/dt/dt.yaml test-kompiled/dt 0
  time=0.034
+ /usr/lib/llvm-14/bin/opt -mem2reg -tailcallelim -tailcallopt tmp.gsYQm9W8RC -o tmp.CTS0WVHve7
  time=0.021
+ /home/bruce/code/llvm-backend/build/install/bin/llvm-kompile-clang tmp.CTS0WVHve7 main nolto -fno-stack-protector -v --profile -fuse-ld=lld --python-output-dir test-kompiled
  | + /usr/lib/llvm-14/bin/llc -tailcallopt tmp.CTS0WVHve7 -mtriple=x86_64-pc-linux-gnu -filetype=obj -O0 --relocation-model=pic -o tmp.H3lGmLF9OO/asm.o
  |   time=0.040
  | + /usr/lib/llvm-14/bin/llc -tailcallopt /home/bruce/code/llvm-backend/build/install/bin/../lib/kllvm//llvm/finish_rewriting.ll -mtriple=x86_64-pc-linux-gnu -filetype=obj -O0 --relocation-model=pic -o tmp.H3lGmLF9OO/finish_rewriting.ll.o
  |   time=0.010
  | + /usr/lib/llvm-14/bin/llc -tailcallopt /home/bruce/code/llvm-backend/build/install/bin/../lib/kllvm//llvm/fresh.ll -mtriple=x86_64-pc-linux-gnu -filetype=obj -O0 --relocation-model=pic -o tmp.H3lGmLF9OO/fresh.ll.o
  |   time=0.010
  | + /usr/lib/llvm-14/bin/llc -tailcallopt /home/bruce/code/llvm-backend/build/install/bin/../lib/kllvm//llvm/getTag.ll -mtriple=x86_64-pc-linux-gnu -filetype=obj -O0 --relocation-model=pic -o tmp.H3lGmLF9OO/getTag.ll.o
  |   time=0.009
[...]

PR #833, opened 2023-09-11
Replacing date/gdate with perl program

Robertorosmaninho:
Fixes https://github.com/runtimeverification/k/issues/3610
This PR replaces the use of `date` and `gdate`, depending on the user system, to get the current timestamp in milliseconds with a simple `Perl` program to improve compatibility across Unix-like systems.
 
This modification is needed due to the following:
- `date` in MacOS is a FreeBSD version that doesn't support (return) timestamps in milliseconds as we need in `llvm-kompile --profile` introduced by #793.
- `gdate` isn't a default package nor a standalone package to be installed in MacOS. Instead, it's commonly installed with `coreutils`. We do not want to ask our users to install all `coreutils` when we only need `gdate` from it and to use it in an optional flag.
- Other options were considered as `python time` and `bash ${EPOCHREALTIME}`, but this was the simplest to use and with common default dependencies in Unix-like systems.

### DIFF
--- a/bin/utils.sh
+++ b/bin/utils.sh
@@ -5,14 +5,9 @@ export DEPTH=0
 verbose=${verbose:-false}
 profile=${profile:-false}
 
-if [[ "$OSTYPE" == "darwin"* ]]; then
-  DATE_CMD="gdate"
-else
-  DATE_CMD="date"
-fi
 
 time_now_ms () {
-  "$DATE_CMD" "+%s%3N"
+  perl -MTime::HiRes=time -e 'printf "%d\n", time*1000'
 }
 
 indented () {
--- a/nix/llvm-backend.nix
+++ b/nix/llvm-backend.nix
@@ -28,9 +28,6 @@ stdenv.mkDerivation {
       --replace 'extra_python_flags=()' \
                 'extra_python_flags=($(${python-env}/bin/pybind11-config --includes))'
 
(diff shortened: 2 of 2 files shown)


Robertorosmaninho:
I can probably delete these lines as well:
https://github.com/runtimeverification/llvm-backend/blob/f3e29231a9a8bc4373fbbc74866f96df454cbc7e/nix/llvm-backend.nix#L31-L32
They were added in https://github.com/runtimeverification/llvm-backend/pull/797 to fix an issue of using `gdate` on MacOS Homebrew.

radumereuta (commented):
Is Perl available by default in Ubuntu and Macs?
Do we need to add it as a dependency to the Dockerfiles?

Robertorosmaninho:
Yes, Perl seems to be default in Ubuntu Focal (v5.30), Jimmy (v5.34), and MacOS 13 (v5.30)! 
It is also default in other distros and macOS versions; I just don't have access to them!

radumereuta (approved):
Ok, let's see what happens.

goodlyrottenapple:
Seems ok if we are reasonably sure perl is always present on a system. Otherwise could always add perl as a runtime dependency for the nix built package here: https://github.com/runtimeverification/llvm-backend/blob/f3e29231a9a8bc4373fbbc74866f96df454cbc7e/nix/llvm-backend.nix#L17

Issue #834, opened 2023-09-11
`coreutils` as llvm-backend dependence in MacOS


Robertorosmaninho:
`coreutils` is currently a [dependence](https://github.com/runtimeverification/llvm-backend/blob/master/INSTALL.md#macos--homebrew) used to get [`gdate`](https://github.com/runtimeverification/llvm-backend/blob/f3e29231a9a8bc4373fbbc74866f96df454cbc7e/bin/utils.sh#L9) and [`uname`](https://github.com/runtimeverification/llvm-backend/blob/f3e29231a9a8bc4373fbbc74866f96df454cbc7e/nix/llvm-backend.nix#L35C28-L35C37) to use in bash scripts. However, both aren't necessary tools as we have other alternatives that don't require them. 

For instance:
- `gdate` -> Perl: #833
- [`uname`](https://github.com/runtimeverification/llvm-backend/blob/f3e29231a9a8bc4373fbbc74866f96df454cbc7e/bin/llvm-kompile-clang#L41) -> [`$OSTYPE`](https://github.com/runtimeverification/llvm-backend/blob/f3e29231a9a8bc4373fbbc74866f96df454cbc7e/bin/utils.sh#L8)

The last substitution has been used across the years by some open-source [projects](https://github.com/dylanaraps/neofetch/issues/433).

PR #835, opened 2023-09-13
Deleting coreutils dependence and replacing its `uname` by `$OSTYPE` bash var

Robertorosmaninho:
Fixes: #834


Baltoli:
This seems like a potentially much more fragile solution than having a dependency on coreutils. What's the motivation for removing the dependency?

Robertorosmaninho:
> This seems like a potentially much more fragile solution than having a dependency on coreutils. What's the motivation for removing the dependency?

There are two reasons to delete `coreutils` dependence:
1. It is only used by `uname` and only once. Many examples in the source code use `$OSTYPE` instead of `uname`.
2. MacOS has a built-in `uname` binary; we can use it if necessary, right? 


I can't really tell the reason why a shell variable would be much more fragile than the system call, but if that's necessary, I can do the other way around in a future PR and replace all `$OSTYPE` comparisons with `uname` ones.

gtrepta:
Some more instances of coreutils that haven't been removed:

https://github.com/runtimeverification/llvm-backend/blob/11b4a5727858f2a6a452673cbd97906dfc1000d4/nix/llvm-backend.nix#L1

https://github.com/runtimeverification/llvm-backend/blob/11b4a5727858f2a6a452673cbd97906dfc1000d4/INSTALL.md?plain=1#L34-L38

Robertorosmaninho:
Thanks @gtrepta!

radumereuta:
I don't feel confident reviewing this since I don't own a Mac.
I think @goodlyrottenapple uses one.

dwightguth (approved):

PR #836, opened 2023-09-15
sortCollections fixes

gtrepta:
For https://github.com/runtimeverification/k/issues/3530

This fixes a couple of issues where the way kprint was sorting AC symbols involved spurious whitespaces and a problematic ordering of operations.


gtrepta:
https://github.com/runtimeverification/llvm-backend/blob/11b4a5727858f2a6a452673cbd97906dfc1000d4/lib/ast/AST.cpp#L845-L878

@dwightguth Can you explain what this is trying to do? It seems like it's trying to find integers in the string and compare them by value. But I think `thisMarker` and `thatMarker` can drift away from each other and compare integers at different offsets, seems dubious. I'm tempted to replace it all with `return s1.compare(s2) < 0;`

dwightguth:
It's meant to be an implementation of comparison that compares strings containing numbers by comparing the numbers numerically rather than lexicographically. Are you experiencing some specific problem with the comparison you're trying to fix?

Robertorosmaninho (approved):
LGTM! Thanks for your work @gtrepta!

PR #837, opened 2023-09-19
Set `perl` as runtime dep in llvm-backend.nix

goodlyrottenapple:


Robertorosmaninho (approved):
If it is possible, we would like to use the host's Perl executable, but right now, this suits the case! Thanks for the help, @goodlyrottenapple!

PR #838, opened 2023-09-25
collect ensures clause in Parser.scala

dwightguth:
Here we add code to the scala codebase to collect into the AxiomInfo the ensures clause of axioms. This is not actually used by the llvm backend because the llvm backend does not support `ensures`, but we need to collect it because this code path is shared with the Maude backend.


Robertorosmaninho (approved):
LGTM! Looking forward to seeing the next steps!

PR #839, opened 2023-09-27
Relative position of function symbols in proof hints

gtrepta:


gtrepta:
The `createHook` function has become even larger with this change. There's lots of repitition there, I can probably cut its size down by quite a bit with some C macros.

lib/codegen/CreateTerm.cpp
```diff
@@ -743,28 +618,33 @@ llvm::Value *CreateTerm::createHook(
...
   std::vector<llvm::Value *> args;
   auto returnSort = dynamic_cast<KORECompositeSort *>(
       pattern->getConstructor()->getSort().get());
   auto returnCat = returnSort->getCategory(Definition);
   int i = 0;
   for (auto sort : pattern->getConstructor()->getArguments()) {
     auto concreteSort = dynamic_cast<KORECompositeSort *>(sort.get());
-    llvm::Value *arg
-        = createAllocation(pattern->getArguments()[i++].get()).first;
+    llvm::Value *arg = createAllocation(
```
dwightguth, line 643: we can use  the macro here too, right?

dwightguth (approved):
I'm approving with one minor comment. Feel free to address or not based on your assessment.

PR #842, opened 2023-09-28
First wave of and/or changes to the llvm backend

dwightguth:
Blocked on https://github.com/runtimeverification/k/pull/3676

This PR implements the changes to the scala code in the llvm backend to support multi-ary and and or. In order to do this without the second wave of changes to the frontend, we rely on a slightly messy approach that involves breaking the abstraction barrier of the scala code in the frontend. this will be cleaned up by a later PR that will revert the first of the two commits in this PR.

### DIFF
--- a/matching/pom.xml
+++ b/matching/pom.xml
@@ -18,14 +18,14 @@
     <repository>
       <id>runtime.verification</id>
       <name>Runtime Verification Repository</name>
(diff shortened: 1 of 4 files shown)


Robertorosmaninho (approved):
LGTM!

PR #845, opened 2023-09-29
Revert "Improve handling of Unicode vs Byte strings (#828)"

Scott-Guest:
The Unicode changes caused a regression in the C-semantics. Reverting the PR until this is investigated.


dwightguth:
Since this is not a straight no-conflict revert, it would seem, it would help to get a summary of exactly the difference between this PR and the one it reverts.

dwightguth:
In particular: did the revert commit apply cleanly? If not, what conflicts did you have to solve? What is the motivation behind the second two commits?

lib/codegen/CreateTerm.cpp
```diff
@@ -294,6 +293,16 @@ sptr<KORESort> termSort(KOREPattern *pattern) {
...
+std::string escape(std::string str) {
+  std::stringstream os;
+  os << std::setfill('0') << std::setw(2) << std::hex;
+  for (char c : str) {
+    unsigned char uc = c;
+    os << (int)uc;
+  }
+  return os.str();
+}
+
```
Scott-Guest, line 305: Had to manually resolve a conflict here by just choosing to include both versions (`ALLOC_ARG` is in master from someone else's changes, `escape` is something I'm reverting)
```
<<<<<<< HEAD                                                                                                                                                                                                                                  
#define ALLOC_ARG(x)                                                           \                                                                                                                                                              
  createAllocation(                                                            \                                                                                                                                                              
      pattern->getArguments()[x].get(),                                        \                                                                                                                                                              
      fmt::format("{}:{}", locationStack, x))                                  \                                                                                                                                                              
      .first                                                                                                                                                                                                                                  
=======                                                                                                                                                                                                                                       
std::string escape(std::string str) {                                                                                                                                                                                                         
  std::stringstream os;                                                                                                                                                                                                                       
  os << std::setfill('0') << std::setw(2) << std::hex;                                                                                                                                                                                        
  for (char c : str) {                                                                                                                                                                                                                        
    unsigned char uc = c;                                                                                                                                                                                                                     
    os << (int)uc;                                                                                                                                                                                                                            
  }                                                                                                                                                                                                                                           
  return os.str();                                                                                                                                                                                                                            
}                                                                                                                                                                                                                                             
>>>>>>> parent of 4c9d010 (Improve handling of Unicode vs Byte strings (#828))
```

lib/codegen/CreateTerm.cpp
```diff
@@ -907,7 +916,7 @@ CreateTerm::createAllocation(KOREPattern *pattern, std::string locationStack) {
                               ->getCategory(Definition)
                               .cat;
                symbolDecl->getAttributes().count("sortInjection")
-               && (cat == SortCategory::Symbol || cat == SortCategory::Bytes)) {
+               && (cat == SortCategory::Symbol)) {
```
Scott-Guest, line 919: Resolved a conflict here. Kept the version from master but just dropped `|| cat = SortCategory::Bytes`. The conditional is then equivalent to the pre-Unicode-changes commit, just slightly restructured.
```
<<<<<<< HEAD                                                                                                                                                                                                                                  
    } else if (auto cat = dynamic_cast<KORECompositeSort *>(                                                                                                                                                                                  
                              symbol->getArguments()[0].get())                                                                                                                                                                                
                              ->getCategory(Definition)                                                                                                                                                                                       
                              .cat;                                                                                                                                                                                                           
               symbolDecl->getAttributes().count("sortInjection")                                                                                                                                                                             
               && (cat == SortCategory::Symbol || cat == SortCategory::Bytes)) {                                                                                                                                                              
      std::pair<llvm::Value *, bool> val = createAllocation(                                                                                                                                                                                  
          constructor->getArguments()[0].get(),                                                                                                                                                                                               
          fmt::format("{}:0", locationStack));                                                                                                                                                                                                
=======                                                                                                                                                                                                                                       
    } else if (                                                                                                                                                                                                                               
        symbolDecl->getAttributes().count("sortInjection")                                                                                                                                                                                    
        && dynamic_cast<KORECompositeSort *>(symbol->getArguments()[0].get())                                                                                                                                                                 
                   ->getCategory(Definition)                                                                                                                                                                                                  
                   .cat                                                                                                                                                                                                                       
               == SortCategory::Symbol) {                                                                                                                                                                                                     
      std::pair<llvm::Value *, bool> val                                                                                                                                                                                                      
          = createAllocation(constructor->getArguments()[0].get());                                                                                                                                                                           
>>>>>>> parent of 4c9d010 (Improve handling of Unicode vs Byte strings (#828)) 
```

Scott-Guest:
I did have to resolve two conflicts in `CreateTerm.cpp`, but they were very straightforward. Left comments showing where they were.

The first commit "Fix incorrect revert of CreateTerm.cpp" was just fixing a dumb mistake where I committed the merge without actually resolving the conflict how I wanted.

The second commit "Update kool-static test output for revert" was to update a test output to use `\x` escapes instead of `\"` or `\n`, which is how it was done pre-Unicode-changes.

dwightguth (approved):

dwightguth (changes requested):
Don't revert the rename of set_len

dwightguth (approved):

PR #846, opened 2023-09-29
Add ZeroORMore option oto debug cl::opt

Robertorosmaninho:
This modification allows multiple `--debug` flags to be accepted by `llvm-kompile-codegen`


Baltoli:
LGTM if you confirm that you've reproduced the original failure case!

Baltoli (approved):

ehildenb:
I'm trying now with KEVM, but I suspect it will fix that first issue.

ehildenb:
I've confirmed that this makes it further on KEVM, thank you @Robertorosmaninho !

Robertorosmaninho:
Awesome! You’re welcome @ehildenb!

PR #849, opened 2023-09-30
Add test for `Bytes2String` infinite loop

### LINKED ISSUE #847: Infinite loop for Bytes2String
As far as I can tell, `Bytes2String(b"\xa9OSt\xfc\xe5\xed\xbc\x8e*\x86\x97\xc1S1g~n\xbf\x0b\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00")` enters an infinite loop when running with the LLVM backend.

Version:
```
$ kompile --version
K version:    v6.0.119-0-g015e5720dc-dirty
Build date:   Vi sep 29 20:18:24 EEST 2023
```

Input files:
a.k:
```
module A-SYNTAX
    imports INT-SYNTAX

    syntax KItem  ::= a(Int)
endmodule

module A
    imports BYTES
    imports A-SYNTAX

    rule a(_) => Bytes2String(b"\xa9OSt\xfc\xe5\xed\xbc\x8e*\x86\x97\xc1S1g~n\xbf\x0b\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00")
endmodule
```
start:
```
a(0)
```
Command line:
```
$ kompile a.k
$ krun start
```

Scott-Guest:
Closes #847.

The infinite loop was caused by the Unicode changes which have since been reverted, so this just adds a regression test.

### DIFF
--- a/nix/llvm-backend-matching.mavenix.lock
+++ b/nix/llvm-backend-matching.mavenix.lock
@@ -246,15 +246,15 @@
       "sha1": "fdec6f2d2514787039928bcb781f9e67f4738899"
     },
     {
-      "path": "com/runtimeverification/k/kore/1.0-SNAPSHOT/kore-1.0-20230929.174637-28.jar",
-      "sha1": "eb69f46a691abd910e46a6f37b721066e9adc804"
+      "path": "com/runtimeverification/k/kore/1.0-SNAPSHOT/kore-1.0-20230930.025149-29.jar",
+      "sha1": "04972642a285d27d20767761d593840798a5ffd5"
     },
     {
-      "path": "com/runtimeverification/k/kore/1.0-SNAPSHOT/kore-1.0-20230929.174637-28.pom",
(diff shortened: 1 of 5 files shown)


ehildenb:
@Scott-Guest is this the same regression that is represented here: https://github.com/runtimeverification/blockchain-k-plugin/pull/158

Maybe that one should be pulled to over here too?

Scott-Guest:
Hmm, the output is here is wrong because the `DecisionTree` uses the frontend parser which treats `\xHH` as `U+00HH` then emits it as UTF-8 to the YAML. 

For some reason though, this particular test does not behave incorrectly when run through `kompile` and `krun` from the K source instead of `llvm-kompile-testing` (the dt folders are different and it doesn't contain the offending file).

Regardless, re-merging the Unicode changes will fix this, so I'll hold off on adding this test until then.

Issue #840, opened 2023-09-28
[GDB] Create script and tests for GDB to run in CI


Robertorosmaninho:

Robertorosmaninho:
- [x] Check if it is possible to automatize GDB/LLDB executions to run then in CI
 > Yes, it is. Both tools have a `batch` mode, and we can pass a list of commands on a file to be executed by GDB/LLDB.
- [ ] What do we want to test? Which programs and commands do we want to verify?
- [ ] How to create a standard output between systems that don't contain any source directory or platform-specific info? We need to pipe all outputs to this standard format.
- [ ] Create the `Debugger Test` directory containing each test in a subfolder. Each test should contain:
  -  The test file
  -  A possible input
  - The debugger commands in a file
  - The expected output
- [ ] Integrate the test-suite with `lit`

Robertorosmaninho:
Blocked on: #858

PR #851, opened 2023-10-03
Creating `debug-command` flag on `llvm-krun`

Robertorosmaninho:
This PR introduces the `--debug-command` to the `llvm-krun` flag that enables the use of GDB and LLDB in a non-iterative mode by passing a file with the debugger commands appropriate for the OS.

This PR is part of #840 and will enable testing the debugger on the CI.


radumereuta (approved):
I would like it if someone more familiar with the LLVM backend approves this also.
But otherwise, lgtm.

dwightguth (approved):

PR #857, opened 2023-10-13
[Debugger] Introducing lldb on for `llvm-krun` script

### LINKED ISSUE #852: [Debugger] `llvm-krun` doesn't support `lldb`
The default debugger for `llvm-krun` is assumed to be `gdb` instead of testing if it's in a Linux or Mac environment and then deciding which debugger to use.

Robertorosmaninho:
Fixes #852 

This simple PR checks if the user is on a `macOS` system before executing the interpreter in debug mode, and if that's the case, the debug command is now modified to call `lldb` with the corrected supported flags instead of calling `gdb`.

It also fixes a bug on `--debug-batch`.


bin/llvm-krun
```diff
@@ -14,7 +14,7 @@ cleanup () {
   rm -rf "$input_file" "$expanded_input_file" "$output_file" "$parser_file" "${temp_inputs[@]}"
 }
 trap 'cleanup' INT TERM EXIT
-
+set -x
```
Baltoli, line 17: I don't think you want to leave this in globally!

bin/llvm-krun
```diff
@@ -279,6 +281,16 @@ for name in "${params[@]}"; do
...
 
+replace_gdb_by_lldb() {
+  if [ -n "$debug" ]; then
+    debug="${debug/gdb/lldb}"
+    debug="${debug/--args /--}"
+    debug="${debug/-x /-s }"
+  fi
+}
+
+#replace_gdb_by_lldb
```
Baltoli, line 292: Comment left in here...

bin/llvm-krun
```diff
@@ -425,7 +437,7 @@ if $expandMacros; then
   if [ "$verbose" -ge 2 ]; then
     set -x
   fi
-  "$(dirname "$0")/kore-expand-macros" "$dir" "$input_file" > "$expanded_input_file"
+ # "$(dirname "$0")/kore-expand-macros" "$dir" "$input_file" > "$expanded_input_file"
```
Baltoli, line 440: ...and here

bin/llvm-krun
```diff
@@ -454,6 +466,10 @@ if $dryRun; then
   exit 0
 fi
 
+if [[ "$OSTYPE" == "darwin"* ]]; then
+  replace_gdb_by_lldb
```
Baltoli, line 470: I would prefer to have the command chosen up front, rather than performing string replacements after the fact, I think

bin/llvm-krun
```diff
@@ -167,10 +167,12 @@ do
       debug="${debug/--args /} -x ${debug_command} --args "
     fi
     shift;
+    shift
     ;;
 
     --debug-batch)
     debug="gdb --batch --args "
```
Baltoli, line 174: Prefer arrays to strings with spaces for command / argument lists, generally

Baltoli (changes requested):
A few bash-related comments

bin/llvm-krun
Robertorosmaninho: You're right! I forgot to clean it up! Sorry and thanks for reminding me!

bin/llvm-krun
Robertorosmaninho: Here, we expect the path of the file, which contains the command we want to execute. I don't think we expect spaces on it.

Baltoli: To elaborate a bit here @Robertorosmaninho, the issue is that building a command line as a string with spaces in it means that you need an unquoted expansion (`$debug`) later in the program. Shellcheck normally doesn't like this (as it can lead to unexpected behaviour re. globbing), but in some cases it will attempt to do some rough [data-flow analysis](https://www.shellcheck.net/wiki/SC2248) to prove that such an expansion is actually safe; your `replace_gdb_by_lldb` function has slightly vexed Shellcheck here I think. I may open a PR later on to tighten up our settings.

See https://www.shellcheck.net/wiki/SC2086 (under "Exceptions") for the full rationale here.

bin/llvm-krun
```diff
@@ -472,7 +484,7 @@ set +e
 if [ "$verbose" -ge 1 ]; then
   set -x
 fi
-$debug "$dir"/interpreter "$expanded_input_file" "$depth" "$output_file" "${interpreter_flags[@]}"
+debug+=("${dir}/interpreter" "$expanded_input_file" "$depth" "$output_file" "${interpreter_flags[@]}")
```
Baltoli, line 487: I don't understand this line. Previously, it was _invoking_ the debugger but here it's appending to an array. Where is the debugger actually executed here?

Baltoli (changes requested):

bin/llvm-krun
Robertorosmaninho: Good point! Sorry for the mistake, I've just updated it!

Baltoli (approved):

Issue #810, opened 2023-07-25
[GDB] `printValueOfType` is not compatible with opaque pointers


Baltoli:
This function performs an explicit comparison against the string value of pointer types; in LLVM 16 and onwards these values are all just `ptr` and so the check breaks.

The fix will be to use the `ValueType` available at the callsite, rather than the string values.

https://github.com/runtimeverification/llvm-backend/blob/b1f1e9c41322a7e3303eb8b973f379f82462d05c/runtime/util/ConfigurationPrinter.cpp#L343-L367C2

radumereuta:
@Baltoli (if you have the time)
can you comment on this with more details about what needs to be done further here?
Is this related to GDB support?

Baltoli:
Yes, this is indeed related to GDB support. The code in question is responsible for pretty-printing a runtime term from the backend; to do this, we need to know what type of term we're looking at (e.g. is the term a block pointer, a leaf value, an MPZ integer...). The representation used by the backend doesn't let you figure that out intrinsically, so we need to carry around some extrinsic information about what terms represent when we dispatch based on their type.

Typically in the backend we carry around a value of type [`ValueType`](https://github.com/runtimeverification/llvm-backend/blob/4c9d0101951dc7fa4f07b0b03f5afd84331a40f7/include/kllvm/ast/AST.h#L132-L142) for this purpose. The problematic code here does not do this, but rather dispatches on the textual representation of the LLVM IR type of the term pointer to be printed (is it a `%block *`, `%mpz_t *`...). In LLVM 16, a breaking change was made that collapses the type of pointer values into a single `%ptr` type, which means that the incorrect branch is selected because the textual values concerned have changed.

The fix for this issue is to amend the call site for this function to keep track of a `ValueType` and pass that information through, rather than checking textual representations of LLVM IR (the latter approach is likely to be fragile in the future for other reasons as well, and I'd consider it a minor code smell generally).

Let me know if any of this doesn't make sense; while I'm still off @Robertorosmaninho probably makes the most sense to fix this as the original author of the code IIRC.

Robertorosmaninho:
Thanks for the clarification, Bruce! I'll try to investigate and fix this issue!

Baltoli:
@Robertorosmaninho I presume that this would be usefully fixed in relation to your changes to enable batched testing of the debugger support?

Robertorosmaninho:
Yes! That would be very helpful!

PR #861, opened 2023-10-16
Fix type printing in LLVM >= 16

Baltoli:
This PR fixes https://github.com/runtimeverification/llvm-backend/issues/810, the last outstanding bit of migration that needed to happen for us to support LLVM 16+ opaque pointers. The full context for this change is in the original issue.

To fix the issue, we need to delay the conversion from `ValueType` to `llvm::Type *` until a deeper point in the call stack when emitting calls to `addMatchFunction`. The changes here are designed to minimise the surface area of modified code; we can get away without changing the `MatchLog` data structure or the `printValueOfType` function by just converting `ValueType`s to an _imitation_ of the legacy LLVM type strings. We might want to revisit this architecture in the future, but for now I'm happy with the design.

The changes made here are tested by the tests for `k-rule-apply`; on platforms with LLVM 16+ we had marked these tests as expectedly failing, and we just remove that label here. In the second commit you can see CI failing because the `XFAIL` is incorrect.

### DIFF
--- a/include/kllvm/codegen/Decision.h
+++ b/include/kllvm/codegen/Decision.h
@@ -201,7 +201,7 @@ public:
 class FunctionNode : public DecisionNode {
 private:
   /* the list of arguments to the function. */
-  std::vector<var_type> bindings;
+  std::vector<std::pair<var_type, ValueType>> bindings;
   /* the name of the variable to bind to the result of the function. */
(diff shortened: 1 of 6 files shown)


Robertorosmaninho (approved):
LGTM

Issue #860, opened 2023-10-16
New proof hint features


Baltoli:
The proof checker team require several new features from the LLVM backend; see https://github.com/runtimeverification/proof-checker/issues/258 for the original context.[^1]

Those features are, with elaborated context from @dwightguth:
* [x] ~~Move where we emit the event that a rule applied to before the call to the apply_rule function rather than immediately before that function returns. Some work on this has been done on [hint_emitPlace]~~(https://github.com/runtimeverification/llvm-backend/compare/hint_emitPlace)
    * What is needed here: right now we emit an event that a rule applied in the apply_rule function generated by CreateTerm.cpp. We want to move this event to being emitted in Decision.cpp in the LeafNode code generation. The event remains otherwise identical.
    * Implemented in #877 
* [x] #881
    * What is needed here: remove the substitution from the event that is emitted when a rule applies. Emit an event at the beginning of the side_condition function generated by CreateTerm.cpp that contains the ordinal and the substitution. This will require ensuring that all variables are passed to the side_condition function even if the side condition does not use them. If a rule does not have a side condition, emit the identical event immediately prior to emitting the event associated with the rule being applied (after task 1 is complete, this will be in Decision.cpp in LeafNode)
* [x] ~~We need to emit the return value of hooks into the trace~~
    * What is needed here: In CreateTerm.cpp, when we emit a call to a hook, emit an event after the hook returns that logs the return value of the hook. This event also needs to be emitted in EmitConfigParser.cpp in the evaluateFunction function.
    * Implemented in #873 

[^1]: The issue can't be transferred directly as it belongs to a private repository.

Baltoli:
Currently the documented format appears to be the same as that accepted by the proof checker's parser; this is _not_ the same as the format being generated by the LLVM backend, and so we should make sure to get these sources in agreement before forking the checker's parser and proceeding to implement these features.

@dwightguth please provide an update with any changes made to the trace format since the [documentation](https://github.com/runtimeverification/llvm-backend/blob/master/docs/proof-trace.md) was written.

Baltoli:
@gtrepta will be working on part 3 of the above list.

Baltoli:
Closing as all three sub-tasks have been implemented.

PR #862, opened 2023-10-17
Refactor generation of proof hint "rule applied" event

Baltoli:
This PR packages up @gtrepta's initial work towards part (1) of #860; a refactoring of the emission of rewrite / rule match events in the proof trace.

At a high level, the code changes here cause the interpreter to emit "rule matched" events for _all_ rules, not just top-level rewrite rules; in the trace below we can see function rules emitting these events. Currently, the term produced by a function event is not yet emitted into the trace, but it could be recovered by evaluating the function on its bound arguments, which are in the trace.

To evaluate the current state of the trace format, consider the following definition:
```k
module TEST
  imports INT

  syntax Int ::= "inc" Int [function, total]
  rule inc I => I +Int 1

  syntax Foo ::= foo(Int) | bar(Int)
  rule foo(I) => bar(inc I)
endmodule
```

When run with `--proof-output`, the sequence of events that is generated in the trace is as follows (with lengthy KORE terms omitted here for my own sanity when extracting this information):
```
function: _|->_
function: .Map
function: _Map_
function: initGeneratedTopCell
rewrite :
  ordinal: 162 (rule initGeneratedTopCell(Init) => ...)
  VarInit -> ...
function: initKCell
rewrite:
  ordinal: 163 (rule initKCell(Init) => ...)
  VarInit -> ...
function: Map:lookup
function: project:KItem
rewrite:
  ordinal: 207 (rule project:KItem(K) => ...)
  VarK -> ...
function: initGeneratedCounterCell
rewrite:
  ordinal: 161 (rule initGeneratedCounterCell() => ...)
<initial configuration>
rewrite:
  ordinal: 126 (rule foo(I) => ...)
  _DotVar0 -> ...
  _DotVar1 -> ...
  I -> ...
function: inc__TEST_Int_Int
rewrite:
  ordinal: 160 (rule inc__TEST_Int_Int(I) => ...)
  I -> ...
function: _+Int_
<next configuration>
```

The things lacking from this trace are currently:
* Return values from hooked functions (#873, #860)
* Side conditions (#860)

Additionally, the parser will need to be updated to reflect this new format as previously.

I would also like to clean up the code a little now that I better understand the trace format and the events being emitted.


lib/codegen/Decision.cpp
```diff
@@ -557,10 +558,117 @@ void LeafNode::codegen(Decision *d) {
...
+  }
+
+  llvm::BranchInst::Create(MergeBlock, TrueBlock);
+  d->CurrentBlock = MergeBlock;
+  auto Call = llvm::CallInst::Create(applyRule, args, "", d->CurrentBlock);
+
+  // How do you emit the resulting config from here?
+
   setDebugLoc(Call);
   Call->setCallingConv(llvm::CallingConv::Tail);
```
gtrepta, line 673: This still needs to be implemented. The call instruction is meant to be tail recursive, I tried changing that but from what I remember it wasn't that simple.

gtrepta (changes requested):

lib/codegen/Decision.cpp
Baltoli: Once this is implemented, the call that's left in `CreateTerm` to emit the config on _rewrites_ can be removed, right? i.e. that it's there and not here is why we have configs and not function returns in the trace.

gtrepta: I believe so, yes.

PR #864, opened 2023-10-17
Remove old interpreter input when running tests

### LINKED ISSUE #863: Tests can receive stale input
The `lit` test suite can leave stale input around on disk when interpreters are run; in most cases this is fine as the parsers for the text and binary formats are largely self-limiting. However, in a handful of tests, the specific content that gets written produces binary KORE files that cause the deserializer to try going off the end of the valid content.

Example output:
```
Failed Tests (3):
  llvm-backend :: defn/test-gc-alwaysgc.kore
  llvm-backend :: defn/test28.kore
  llvm-backend :: defn/test3.kore
```

After wiping the test output directories and re-running, these tests pass.

Baltoli:
This PR makes a small change to the `lit` lines we use to run checks on interpreter output; currently they can receive partially stale input if the test suite has previously been run. The solution is simply to remove any old output from previous interpreter runs before we run them again.

Fixes #863

### DIFF
--- a/test/lit.cfg.py
+++ b/test/lit.cfg.py
@@ -129,8 +129,8 @@ config.substitutions.extend([
         done
     ''')),
 
-    ('%run-binary-out', '%t.interpreter %test-input -1 %t.out.bin --binary-output'),
-    ('%run-binary', '%convert-input && %t.interpreter %t.bin -1 /dev/stdout'),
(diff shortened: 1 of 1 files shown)


Scott-Guest (approved):
LGTM, thanks for fixing this!

PR #866, opened 2023-10-17
Memset hook

mariaKt:
This PR adds a hook that implements `memset` for the llvm backend. We intend to use this to perform copies while avoiding the overhead of an extra buffer copy in the C semantics.

### DIFF
--- a/runtime/strings/bytes.cpp
+++ b/runtime/strings/bytes.cpp
@@ -194,6 +194,11 @@ SortBytes hook_BYTES_memset(SortBytes b, SortInt start, SortInt count, SortInt v
         ustart, ucount, input_len);
   }
   int v = mpz_get_si(value);
+  if ((v < -128) || (v > 127)) {
+    KLLVM_HOOK_INVALID_ARGUMENT(
+        "Not a valid value for a byte in memset: v={}",
+        v);
+  }
   memset(b->data + ustart, v, ucount);
   return b;
 }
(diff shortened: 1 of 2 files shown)


runtime/strings/bytes.cpp
```diff
@@ -182,6 +182,27 @@ SortBytes hook_BYTES_replaceAt(SortBytes b, SortInt start, SortBytes b2) {
   return b;
 }
 
+SortBytes
+hook_BYTES_memset(SortBytes b, SortInt start, SortInt count, SortInt value) {
+  uint64_t ustart = get_ui(start);
+  uint64_t ucount = get_ui(count);
+  uint64_t uend = ustart + ucount;
```
Baltoli, line 189: Perhaps implement this in terms of `hook_INT_add`? It's a _really_ unlikely corner case, but it would mean we get an overflow error message rather than just a silent failure.

Baltoli (approved):
One very pedantic comment but otherwise this looks great - thanks @mariaKt

runtime/strings/bytes.cpp
dwightguth: I don't think it makes sense to use arbitrary precision arithmetic here, but we should  write a check for overflow since it would be undefined behavior. We can just throw an exception if the addition would overflow.

Baltoli: Yeah, that's all I'm getting at - my thought was you'd catch the overflow when lowering the result of the MPZ addition back to a regular integer.

Baltoli (approved):
Looks good to me!

PR #814, opened 2023-07-26
Inline invocations of `llc` for main module

Baltoli:
This follows #794 in the pursuit of improving the performance of the backend's compilation pipeline; the philosophy of the change is identical (remove places where we materialize code in the compilation pipeline, and instead use the LLVM libraries to do it all in-memory).

There are a few more changes required in this PR in order to expose the full generality of the options we previously supported for `llc`, but these are largely boilerplate.

The default behaviour of `llvm-kompile` is now to use the new pipeline, but it's possible to opt back out of it for debugging.

PR #869, opened 2023-10-19
Hotfix for K LLVM backend update

Baltoli:
This PR is a hotfix for a [broken](https://github.com/runtimeverification/k/pull/3728) backend update job.

When using the new [inlined version of `llvm-kompile`](https://github.com/runtimeverification/llvm-backend/pull/814), we were incorrectly using the result of `sys::getDefaultTargetTriple()` as our target for machine code emission, rather than the consistent value selected by running Clang on an empty input file.

On newer versions of macOS, this produces warnings like:
```
ld: warning: object file (tmp.1YutNb2k5B.o) was built for newer macOS version (14.0) than being linked (10.12)
```
when running `llvm-kompile`. These warnings were silently absorbed by CI as they are just warnings, but they are a) noisy and b) break the K test suite when grepping for error messages in failed compilations (`kompile ... 2>&1 | grep ...`).

The fix is just to use our existing technique for target triple selection in the one place we weren't previously.

Additionally, the PR makes two small related changes:
* Add conditional compilation guards for some LLVM headers that give deprecation warnings in LLVM 17
* Catch future errors of this kind in the backend CI by ensuring that `llvm-kompile` is silent and doesn't pollute `stderr` with extra output when it runs successfully

### DIFF
--- a/lib/codegen/ApplyPasses.cpp
+++ b/lib/codegen/ApplyPasses.cpp
@@ -7,10 +7,17 @@
 #endif
 
 #include <llvm/IR/LegacyPassManager.h>
+
+#if LLVM_VERSION_MAJOR >= 17
+#include <llvm/TargetParser/Host.h>
+#include <llvm/TargetParser/SubtargetFeature.h>
+#else
 #include <llvm/MC/SubtargetFeature.h>
+#include <llvm/Support/Host.h>
+#endif
+
 #include <llvm/Pass.h>
 #include <llvm/Support/CommandLine.h>
-#include <llvm/Support/Host.h>
 #include <llvm/Target/TargetMachine.h>
 #include <llvm/Target/TargetOptions.h>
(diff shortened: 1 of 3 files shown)


lib/codegen/ApplyPasses.cpp
```diff
@@ -73,7 +80,7 @@ void generate_object_file(llvm::Module &mod, llvm::raw_ostream &os) {
   }
 #endif
 
-  auto triple = sys::getDefaultTargetTriple();
+  auto triple = "@BACKEND_TARGET_TRIPLE@";
```
dwightguth, line 83: I'm starting to think this should maybe be in config/macros.h. what do you think?

Baltoli: Addressing in a followup PR.

dwightguth (approved):
@Baltoli agreed to address my comment in a follow-up pr

PR #872, opened 2023-10-22
Function event proof hint documentation

gtrepta:
This documents the function event in the proof hints that were added in an earlier commit.

### DIFF
--- a/docs/proof-trace.md
+++ b/docs/proof-trace.md
@@ -38,7 +38,7 @@ rewrite_trace := rule_ordinal rule_arity variable* delimited_serial_kore
 
 initial_config := delimited_serial_kore
 
-proof_trace := initial_config (function_event|rewrite_trace)*
+proof_trace := function_event* initial_config (function_event|rewrite_trace)*
 ```
 
 ## Notes
@@ -46,3 +46,5 @@ proof_trace := initial_config (function_event|rewrite_trace)*
 - The `rule_arity` should be used to determine how many variable substitutions to read
 - The serialized term for a variable substitution does not begin with the sentinel delimiter.
   This is because the null terminated variable name can act as the sentinel.
+- The `function_event`s at the beginning of the proof_trace are related to configuration initialization.
+- The `relative_position` is a null terminated string of positive integers separated by `:` (ie. `0:1:1`)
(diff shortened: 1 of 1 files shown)


Baltoli (approved):
Approving as this BNF will let me figure out how to extend the existing proof event parser for the time being; can you please add a textual description of what this part of the format does, and how to interpret the `relative_position` fields wrt. a configuration?

docs/proof-trace.md
```diff
@@ -25,6 +25,9 @@ Here is a BNF styled description of the format
 delimited_serial_kore := 0xffffffffffffffff serialized_term 0xcccccccccccccccc
 
 null_terminated_name := <c-style null terminated string>
+relative_position := [0-9]+(:[0-9]+)* 0x00
```
dwightguth, line 28: It's not clear how this is actually formatted. Is it just a string? We can maybe make that clearer.

Baltoli: My understanding from reading the code and looking at the generated files, it's just a null-terminated string that's made up of a series of non-negative integers separated by `:`

Baltoli (changes requested):
In fact, I realise now that this description does not actually capture accurately what's being generated by the backend. If I run one of the examples from the proof checker repo, I see that at the beginning of the file there are some function events when I'd expect to see an initial term:
```console
$ xxd ./generation/proof-hints/peano/mul_3_5.peano.hints | head -5
00000000: dddd dddd dddd dddd 4c62 6c27 556e 6473  ........Lbl'Unds
00000010: 5069 7065 272d 272d 4754 2d55 6e64 7327  Pipe'-'-GT-Unds'
00000020: 7b7d 0030 00dd dddd dddd dddd dd4c 626c  {}.0.........Lbl
00000030: 2753 746f 7027 4d61 707b 7d00 3000 dddd  'Stop'Map{}.0...
00000040: dddd dddd dddd 4c62 6c27 556e 6473 274d  ......Lbl'Unds'M
```

gtrepta:
@Baltoli Oh, good catch. I think those are function events for configuration initialization.

docs/proof-trace.md
gtrepta: Yeah, it's a null terminated string like `0:1:1`. I'll try to make it clearer.

gtrepta:
Letting #873 subsume this one.

PR #873, opened 2023-10-23
Hook event for proof hints

gtrepta:


lib/codegen/CreateTerm.cpp
```diff
@@ -916,11 +922,25 @@ CreateTerm::createAllocation(KOREPattern *pattern, std::string locationStack) {
...
+                                                    ->getArguments()[0]
+                                                    .get());
+        std::string name = strPattern->getContents();
+
+        ProofEvent p1(Definition, CurrentBlock, Module);
+        CurrentBlock = p1.hookEvent_pre(name);
+        llvm::Value *val = createHook(
+            symbolDecl->getAttributes().at("hook").get(), constructor,
+            locationStack);
+        ProofEvent p2(Definition, CurrentBlock, Module);
```
Baltoli, line 940: Why do we need two separate `ProofEvent`s here?

lib/codegen/ProofEvent.cpp
```diff
@@ -0,0 +1,198 @@
...
+        val, llvm::Type::getInt8PtrTy(Module->getContext()));
+    ir->CreateCall(
+        getOrInsertFunction(
+            Module, "serializeRawTermToFile",
+            llvm::Type::getVoidTy(Module->getContext()),
+            llvm::Type::getInt8PtrTy(Module->getContext()),
+            llvm::Type::getInt8PtrTy(Module->getContext()),
+            llvm::Type::getInt8PtrTy(Module->getContext())),
+        {outputFile, val, sortptr});
+  }
```
Baltoli, line 124: This particular idiom to safely serialize a term given its sort and type gets repeated below; could we pull it out into its own function? The changes in #862 will need the same code as well, so it makes sense to extract it.

lib/codegen/ProofEvent.cpp
```diff
@@ -0,0 +1,198 @@
...
+
+  llvm::BranchInst::Create(MergeBlock, TrueBlock);
+
+  llvm::BinaryOperator::Create(
+      llvm::Instruction::Add,
+      llvm::ConstantInt::get(llvm::Type::getInt8Ty(Module->getContext()), 0),
+      llvm::ConstantInt::get(llvm::Type::getInt8Ty(Module->getContext()), 0),
+      "nop", MergeBlock);
```
Baltoli, line 132: This is commented elsewhere, but the change is repeated a couple of times so perhaps it makes sense to extract the code. We could also audit callers to see if the "insert after last instruction" idiom could be changed; you will know better than me how viable that is.

Baltoli (approved):
Approving as there are only a couple of minor changes I can see

lib/codegen/ProofEvent.cpp
gtrepta: Yep, this block here appears in multiple places and should definitely be factored into its own function.

lib/codegen/ProofEvent.cpp
gtrepta: Yeah, this can be extracted.

"insert after last instruction" is kind of the standard way of building up instructions in an `llvm::BasicBlock`, so I don't think we want to change that. In this case I think we may have hit a bug in llvm, I'll try opening and issue for it on their project.

Issue #868, opened 2023-10-18
Deprecate LLVM < 15


Baltoli:
We have dropped support for Ubuntu Focal (20.04) in K, and so we no longer have any platforms where the newest version is < 15. We can therefore safely deprecate these versions in the backend.
* [x] Update any documentation that still recommends an installation of LLVM 14 or older.
* [x] Identify any conditional compilation macro usages that we can fold away.
* [x] Update the build system and test matrix appropriately.

PR #874, opened 2023-10-30
Drop LLVM <15

Baltoli:
Having dropped support for Ubuntu Focal in K, we no longer need to support versions of LLVM older than 15; this PR makes the required changes to the backend to drop support for those older versions. I will follow up with a K PR to update the corresponding documentation.

Fixes https://github.com/runtimeverification/llvm-backend/issues/868

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1,11 +1,11 @@
 # Dependencies
 
-## Ubuntu 20.04:
+## Ubuntu 22.04:
 
 ```shell
 sudo apt update
 sudo apt install      \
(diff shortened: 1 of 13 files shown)


dwightguth (approved):
Looks fine, but some of these functions existed primarily to support multiple LLVM versions and we may want to inline and delete them.

Robertorosmaninho (approved):
I'm very happy with these modifications and how they're simplifying the code. Besides Dwight's comment, the PR LGTM!

PR #875, opened 2023-11-03
Use domain value argument sort if possible

Baltoli:
This PR implements a feature requested by @gtrepta whereby `getSort()` calls on _domain value_ patterns that are part of a definition will return the first formal sort argument (e.g. `SortX{}` in `\dv{SortX{}}("...")` rather than `nullptr`. Doing so simplifies printing and serialization logic in some parts of the backend.

As well as CI passing, I have tested this on upstream K and it does not appear to break anything; my suspicion is that we just weren't relying on these sorts being `nullptr` previously.

The PR also pins a specific version of `clang-format`, which we weren't doing previously.

### DIFF
--- a/.github/workflows/clang-format-check.yml
+++ b/.github/workflows/clang-format-check.yml
@@ -12,9 +12,10 @@ jobs:
     steps:
     - uses: actions/checkout@v3
     - name: Check that C and C++ code is correctly formatted
-      uses: jidicula/clang-format-action@v4.8.0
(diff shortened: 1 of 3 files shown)


Baltoli:
Merging as @gtrepta has essentially replicated these changes on his own branch: https://github.com/runtimeverification/llvm-backend/pull/873/files#diff-c7e4888f3b570630e2b50cba88f36b8a002aa9cc637f9bc6391559f1b2c38c49

PR #877, opened 2023-11-09
Move rewrite event emission

Baltoli:
This PR is a reworking of the changes initially made by @gtrepta in #862, to include the cleaned-up proof infrastructure in #876.

The changes here are pretty minimal; we need to split the "rewrite" event into a pre and post part so that the arity / ordinal and resulting configuration can be emitted separately. Other than that, it's just a question of reusing some of the boilerplate in #862 that assembles the substitution information in the event.

~~There is one major issue with the change here, which @gtrepta also identified in his initial version of this change. We can only emit resulting configurations for top-level rewrites (that is, we cannot print the resulting _term_ from calling a function). Doing so breaks TCO for the generated interpreter badly enough that some backend tests crash; I have not dug into this in detail as I'm reasonably sure that's the reason things break.~~
* @nishantjr confirms that we can safely omit the result of function calls from the trace.

## Outstanding Issues

I am working my way through the example below to ensure that the format being generated is precisely what we expect to see from the documentation. The issues I have identified are as follows:
* [x] Duplicated KORE comments below are caused by function events not being bracketed as rewrites and hooks are; we need to pull the header out to the front, allocate args, then do the existing event.
* [x] We need to handle binary KORE delimitation more consistently; at the moment it's wavy whether we emit header / footer or not. Should probably standardise on using the 1.2.0 self-limiting format.

## Worked Example

Similarly to #862, I have generated a trace for the program `foo(0)` in the following definition:
```k
module TEST
  imports INT

  syntax Int ::= "inc" Int [function, total]
  rule inc I => I +Int 1

  syntax Foo ::= foo(Int) | bar(Int)
  rule foo(I) => bar(inc I)
endmodule
```

The trace looks like:
```
hook: MAP.element
  function: Lbl'UndsPipe'-'-GT-Unds'{} (0)
    arg: kore[60]
    arg: kore[109]
hook result: kore[207]
hook: MAP.unit
  function: Lbl'Stop'Map{} (0)
hook result: kore[51]
hook: MAP.concat
  function: Lbl'Unds'Map'Unds'{} (0)
    arg: kore[51]
    arg: kore[207]
hook result: kore[207]
function: LblinitGeneratedTopCell{} (0)
  arg: kore[207]
rule: 162 1
  VarInit = kore[207]
function: LblinitKCell{} (0:0)
  arg: kore[207]
rule: 163 1
  VarInit = kore[207]
function: Lblproject'Coln'KItem{} (0:0:0)
  hook: MAP.lookup
    function: LblMap'Coln'lookup{} (0:0:0:0:0)
      arg: kore[207]
      arg: kore[60]
  hook result: kore[109]
  arg: kore[129]
rule: 207 1
  VarK = kore[109]
function: LblinitGeneratedCounterCell{} (0:1)
rule: 161 0
config: kore[237]
rule: 126 3
  Var'Unds'DotVar0 = kore[61]
  VarI = kore[50]
  Var'Unds'DotVar1 = kore[10]
function: Lblinc'UndsUnds'TEST'Unds'Int'Unds'Int{} (0:0:0:0:0:0)
  arg: kore[50]
rule: 160 1
  VarI = kore[50]
hook: INT.add
  function: Lbl'UndsPlus'Int'Unds'{} (0)
    arg: kore[50]
    arg: kore[50]
hook result: kore[50]
config: kore[237]
```

And was generated with the following (really horrible, quick and dirty) Python recogniser:
<details><summary>Recogniser</summary>

```python
#!/usr/bin/env python3

import sys
import struct

def load(fn):
    with open(fn, 'rb') as f:
        return f.read()

def word(byte):
    return byte * 8

class Parser:
    def __init__(self, data):
        self.input = data
        self.depth = 0

    def print(self, *args):
        print(f'{"  " * self.depth}', end='')
        print(*args)

    def peek(self, cst):
        return self.input[:len(cst)] == cst

    def eof(self):
        return len(self.input) == 0

    def skip(self, n):
        self.input = self.input[n:]

    def read_constant(self, cst):
        if self.input[:len(cst)] != cst:
            print(self.input[:40])
            assert False
        self.input = self.input[len(cst):]

    def read_word(self, byte):
        assert len(byte) == 1
        self.read_constant(byte * 8)

    def read_until(self, constant):
        index = self.input.find(constant)
        ret = self.input[:index]
        self.input = self.input[index:]
        return ret
    
    def read_string(self):
        data = self.read_until(b'\x00')
        self.read_constant(b'\x00')
        return str(data, 'ascii')

    def read_uint64(self):
        index = 8
        raw = self.input[:index]
        self.input = self.input[index:]
        little_endian_long_long = '<Q'
        return struct.unpack(little_endian_long_long, raw)[0]

    def read_arg(self):
        self.depth += 1
        if self.peek(b'\x7F'):
            l = self.read_kore()
            self.print(f'arg: kore[{l}]')
            self.depth -= 1
            return True
        elif self.peek(word(b'\xDD')):
            self.read_function()
            self.depth -= 1
            return True
        elif self.peek(word(b'\xAA')):
            self.read_hook()
            self.depth -= 1
            return True
        elif self.peek(word(b'\x11')) or self.peek(word(b'\xBB')):
            self.depth -= 1
            return False
        else:
            self.read_rule()
            self.depth -= 1
            return True

    def read_hook(self):
        self.read_word(b'\xAA')
        name = self.read_string()
        self.print(f'hook: {name}')
        while True:
            if not self.read_arg():
                break
        self.read_word(b'\xBB')
        result = self.read_kore()
        self.print(f'hook result: kore[{result}]')

    def read_function(self):
        self.read_word(b'\xDD')
        label = self.read_string()
        loc = self.read_string()
        self.print(f'function: {label} ({loc})')
        while True:
            if not self.read_arg():
                break
        self.read_word(b'\x11')

    def read_kore(self):
        self.read_constant(b'\x7FKORE')
        self.skip(6)
        pattern_len = self.read_uint64()
        self.skip(pattern_len)
        return pattern_len

    def read_rule(self):
        ordinal = self.read_uint64()
        arity = self.read_uint64()
        self.print(f'rule: {ordinal} {arity}')
        for i in range(arity):
            self.read_variable()

    def read_variable(self):
        name = self.read_string()
        kore_len = self.read_kore()
        self.print(f'  {name} = kore[{kore_len}]')
        self.read_word(b'\xCC')

    def read_config(self):
        self.read_word(b'\xFF')
        l = self.read_kore()
        self.print(f'config: kore[{l}]')
        self.read_word(b'\xCC')

    def read_trace(self):
        while not self.eof():
            if self.peek(word(b'\xAA')):
                self.read_hook()
            elif self.peek(word(b'\xDD')):
                self.read_function()
            elif self.peek(word(b'\xFF')):
                self.read_config()
            else:
                self.read_rule()
            
if __name__ == '__main__':
    data = load(sys.argv[1])
    parser = Parser(data)
    parser.read_trace()
```
</details>

I have verified by hand that this recogniser accepts all the proof traces from the proof generation repo's master branch, if they are regenerated with the LLVM backend pointed at this PR (`kup install k --override llvm-backend hint_emitPlace_2`).

### DIFF
--- a/docs/proof-trace.md
+++ b/docs/proof-trace.md
@@ -19,38 +19,42 @@ have rendered this unnecessary, but the change hasn't been implemented yet.
 
 ## Grammar
 
-Here is a BNF styled description of the format
-
+Here is a BNF styled description of the format:
 ```
-delimited_serial_kore := 0xffffffffffffffff serialized_term 0xcccccccccccccccc
+proof_trace ::= event*
 
-null_terminated_name := <c-style null terminated string>
-relative_position := [0-9]+(:[0-9]+)* 0x00
+event       ::= hook
+              | function
+              | rule
+              | config
 
-function_event := 0xdddddddddddddddd null_terminated_name relative_position
+argument    ::= hook
+              | function
+              | rule
+              | kore_term
 
-hook_event := 0xaaaaaaaaaaaaaaaa null_terminated_name hook_arg* 0xbbbbbbbbbbbbbbbb serialized_term
+name        ::= string
+location    ::= string
+function    ::= WORD(0xDD) name location arg* WORD(0x11)
 
-hook_arg := serialized_term
-          | function_event
-          | hook_event
+hook        ::= WORD(0xAA) name arg* WORD(0xBB) kore_term
 
-variable := null_terminated_name serialized_term 0xcccccccccccccccc
+ordinal     ::= uint64
+arity       ::= uint64
+variable    ::= name kore_term WORD(0xCC)
+rule        ::= ordinal arity variable*
 
-rule_ordinal := <64-bit unsigned little endian integer>
-rule_arity := <64-bit unsigned little endian integer>
+config      ::= WORD(0xFF) kore_term WORD(0xCC)
 
-rewrite_trace := rule_ordinal rule_arity variable* delimited_serial_kore
-
-initial_config := delimited_serial_kore
-
-proof_trace := (function_event|hook_event)* initial_config (function_event|hook_event|rewrite_trace)*
+string      ::= <c-style null terminated string>
+uint64      ::= <64-bit unsigned little endian integer>
 ```
 
 ## Notes
 
-- The `rule_arity` should be used to determine how many variable substitutions to read
-- The serialized term for a variable substitution does not begin with the sentinel delimiter.
-  This is because the null terminated variable name can act as the sentinel.
-- The `function_event|hook_event`s at the beginning of the proof_trace are related to configuration initialization.
-- The `relative_position` is a null terminated string of positive integers separated by `:` (ie. `0:1:1`)
+- The `rule_arity` should be used to determine how many variable substitutions
+  to read.
+- Events at the beginning of the trace (i.e. before the first `config` event)
+  are related to configuration initialization.
+- The `relative_position` is a null terminated string of positive integers
+  separated by `:` (ie. `0:1:1`)
--- a/include/kllvm/codegen/ProofEvent.h
+++ b/include/kllvm/codegen/ProofEvent.h
@@ -102,11 +102,14 @@ public:
       llvm::Value *val, KORECompositeSort *sort,
       llvm::BasicBlock *current_block);
 
-  [[nodiscard]] llvm::BasicBlock *rewriteEvent(
-      KOREAxiomDeclaration *axiom, llvm::Value *return_value, uint64_t arity,
+  [[nodiscard]] llvm::BasicBlock *rewriteEvent_pre(
+      KOREAxiomDeclaration *axiom, uint64_t arity,
       std::map<std::string, KOREVariablePattern *> vars,
       llvm::StringMap<llvm::Value *> const &subst,
       llvm::BasicBlock *current_block);
+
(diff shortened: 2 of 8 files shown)


F-WRunTime:
Something has happened in a post job that has stalled out the runner for 4hrs, this needs to be canceled and rerun

dwightguth:
I believe I mentioned the issue with TCO to the proof object generation team and they said it should be fine to skip emitting a resulting term for functional rules.

dwightguth:
It's worth reaching out to confirm that is still true based on their current understanding, but there's not much that can be done regardless; TCO is a fundamental part of correctness in bounded amounts of memory in a functional language, and there is no way to emit that information in a way that doesn't break TCO.

Baltoli:
That was my _vague_ recollection as well, but I didn't want to assume I was correct. If that is indeed the case then I'll clean up the commented code and this can be considered properly ready for review. Thanks Dwight!

gtrepta (approved):
LGTM

Baltoli:
@malturki has OKed the grammar for his work on the checker's parser.

malturki (approved):
LGTM

PR #879, opened 2023-11-10
Flags to turn on/off proof hint instrumenation and output

theo25:
This PR adds two flags to our scripts to turn the proof hint instrumentation and output on and off:
- a flag `--proof-hint-instrumentation` is added to the `llvm-kompile` script. When the flag is passed the generated llvm code is instrumented with proof hint generation code.
- a flag `--proof-hints` is added to the `llvm-krun` script. When the flag is passed to an instrumented interpreter binary, the binary will output proof hints.

### DIFF
--- a/lib/codegen/CreateTerm.cpp
+++ b/lib/codegen/CreateTerm.cpp
@@ -21,6 +21,13 @@
 #include "llvm/Support/SourceMgr.h"
 #include "llvm/Support/raw_ostream.h"
 
+extern llvm::cl::OptionCategory CodegenCat;
+
+llvm::cl::opt<bool> ProofHintInstrumentation(
+    "proof-hint-instrumentation",
+    llvm::cl::desc("Enable instrumentation for generation of proof hints"),
+    llvm::cl::cat(CodegenCat));
+
 namespace kllvm {
 
 using namespace fmt::literals;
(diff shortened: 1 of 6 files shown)


gtrepta (commented):
Maybe it would be simpler for `ProofEvent` to read the ProofHintInstrumentation flag and turn its code emitting functions into effective no-ops if it's false? That way nothing needs to be changed across the multiple places it gets used.

lib/codegen/CreateTerm.cpp
```diff
@@ -21,6 +21,13 @@
 #include "llvm/Support/SourceMgr.h"
 #include "llvm/Support/raw_ostream.h"
 
+extern llvm::cl::OptionCategory CodegenCat;
+
+llvm::cl::opt<bool> ProofHintInstrumentation(
+    "proof-hint-instrumentation",
+    llvm::cl::desc("Enable instrumentation for generation of proof hints"),
+    llvm::cl::cat(CodegenCat));
```
Baltoli, line 29: Yep, I basically agree with Guy's suggestion - let's push this option into the `ProofEvent` class. Otherwise the plumbing code etc. all looks good!

Baltoli (approved):

Baltoli (changes requested):
One organisational comment as Guy notes also; otherwise all good!

gtrepta (changes requested):
You'll need to add a lit command to kompile with the proof instrumentation and have `tests/defn/imp-proof.kore` use it

https://github.com/runtimeverification/llvm-backend/blob/0e24009e09ac3f96b796a9bb11d078dc3b49fdc3/test/lit.cfg.py#L80-L87

Baltoli (approved):
Looks good @theo25; feel free to merge!

Baltoli:
@gtrepta you'll need to dismiss your review here if you're happy with the state of the PR

gtrepta (approved):

PR #885, opened 2023-11-14
Version header for proof hints

gtrepta:
fixes: #883

### DIFF
--- a/docs/proof-trace.md
+++ b/docs/proof-trace.md
@@ -21,7 +21,9 @@ have rendered this unnecessary, but the change hasn't been implemented yet.
 
 Here is a BNF styled description of the format:
 ```
-proof_trace ::= event*
+proof_trace ::= header event*
+
+header      ::= "HINT" <4-byte version number>
 
 event       ::= hook
               | function
--- a/runtime/util/util.cpp
+++ b/runtime/util/util.cpp
@@ -28,4 +28,13 @@ block *constructKItemInj(void *subject, const char *sort, bool raw_value) {
   auto args = std::vector<void *>{add_indirection ? (void *)&subject : subject};
   return static_cast<block *>(constructCompositePattern(tag, args));
 }
+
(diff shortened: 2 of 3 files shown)


docs/proof-trace.md
```diff
@@ -21,7 +21,9 @@ have rendered this unnecessary, but the change hasn't been implemented yet.
 
 Here is a BNF styled description of the format:
 ```
-proof_trace ::= event*
+proof_trace ::= header event*
+
+header      ::= "HINT" <4-byte version number>
```
Baltoli, line 26: Could we have this be a semantic version like in the binary KORE format (which uses 3 16-bit unsigned integers for `major . minor . patch`)

runtime/util/util.cpp
```diff
@@ -28,4 +28,13 @@ block *constructKItemInj(void *subject, const char *sort, bool raw_value) {
   auto args = std::vector<void *>{add_indirection ? (void *)&subject : subject};
   return static_cast<block *>(constructCompositePattern(tag, args));
 }
+
+void printProofHintHeader(char *output_file) {
+  unsigned int version = 1;
```
Baltoli, line 33: https://github.com/runtimeverification/llvm-backend/blob/master/include/kllvm/binary/version.h maybe pull this class upwards out of the `binary` library and reuse it here?

docs/proof-trace.md
Baltoli: See comment below, in fact

gtrepta: @dwightguth asked for the single integer over the semantic version. I think I agree with doing it this way, the format shouldn't really go through any radical changes, and there probably wont be any widespread use of it outside of a couple of our tools.

Baltoli: In that case it's fine by me!

Baltoli (approved):

PR #888, opened 2023-11-15
Move some stray install locations into the `kllvm` subdirectories

gtrepta:
for https://github.com/runtimeverification/k/issues/3675


Baltoli:
Have you tested this with downstream K / semantics to make sure nothing is incorrectly depending on the old paths?

gtrepta:
@Baltoli I ran the K package tests for Jammy locally and they passed. The nix and MacOS tests will take too long to figure out, but I feel pretty confident that this change wont break anything.

Baltoli:
Yep, I also tested the C semantics so this looks good to go.

Baltoli (approved):

PR #891, opened 2023-11-16
Missing cases for proof instrumentation flag

Baltoli:
We were incorrectly emitting proof instrumentation when these event functions were called; this PR adds the same flag check to their preludes as the other events have. I have verified that there is no proof-related generated code in the module produced by the code generator after this change.


theo25 (approved):
Looks good to me

gtrepta (approved):

PR #893, opened 2023-11-17
Bump hint version for hook locations

gtrepta:


Baltoli (approved):

runtime/util/util.cpp
```diff
@@ -30,7 +30,7 @@ block *constructKItemInj(void *subject, const char *sort, bool raw_value) {
 }
 
 void printProofHintHeader(char *output_file) {
-  unsigned int version = 1;
+  unsigned int version = 2;
```
theo25, line 33: this is a nit pick, but could we use `uint32_t` here. It is not guaranteed that `unsigned int` is 4 bytes long in every architecture.

Baltoli, line 33:
```suggestion
  uint32_t version = 2;
```

Baltoli (approved):

PR #897, opened 2023-11-21
Add term simplification to Python bindings

Baltoli:
This PR exposes the concrete term simplification routines already used by the booster to the Python bindings so that they can be called from Pyk. Specifically, it adds two free functions to the `runtime` module:
```python
simplify_pattern(Pattern, Sort) -> Pattern
simplify_bool_pattern(Pattern) -> bool
```

These are exposed as free functions because they depend on the runtime library and so can't be automatically bound to the `Pattern` class as methods; I suggest Pyk probably wants to monkey-patch them into its `Pattern` class along with some kind of dynamic check for whether the `runtime` module has been loaded or not.

Previously, the C and Python bindings didn't share a great deal of code; the simplification routine is the first bit of non-trivial behaviour that they need to share. I have therefore started a reorganisation of the bindings code to extract as much functionality to a common library, which has the additional benefit of reducing the amount of memory-leak prone handling of C wrapper structures. To keep the diff in this PR reasonable I have not completed this migration, but this is future work that I will complete in a future PR.

A summary of the changes in this PR is as follows:
* Build a common bindings library that the C and Python bindings link against.
* Extract the minimal set of features from the existing C bindings that the Python bindings require to perform term simplification, and port the C bindings over to calling the extracted common features.
* Add a test case to the Python bindings test suite that performs a few small simplifications of sort `Int` and `Bool`; these are really just smoke tests and Pyk should test the feature more thoroughly.

### DIFF
--- a/bindings/c/lib.cpp
+++ b/bindings/c/lib.cpp
@@ -1,5 +1,7 @@
 #include <kllvm-c/kllvm-c.h>
 
+#include "bindings/core/core.h"
+
 #include <kllvm/ast/AST.h>
 #include <kllvm/binary/serializer.h>
 #include <kllvm/parser/KOREParser.h>
@@ -52,7 +54,6 @@ extern "C" {
 block *take_steps(int64_t, block *);
 void initStaticObjects(void);
 void freeAllKoreMem(void);
-void *constructInitialConfiguration(const kllvm::KOREPattern *);
 }
 
 extern "C" {
@@ -185,14 +186,8 @@ kore_pattern *kore_pattern_new_token_with_len(
 
 kore_pattern *kore_pattern_new_injection(
     kore_pattern const *term, kore_sort const *from, kore_sort const *to) {
-  auto inj_sym = kore_symbol_new("inj");
-  kore_symbol_add_formal_argument(inj_sym, from);
-  kore_symbol_add_formal_argument(inj_sym, to);
-
-  auto inj = kore_composite_pattern_from_symbol(inj_sym);
-  kore_composite_pattern_add_argument(inj, term);
-
-  kore_symbol_free(inj_sym);
+  auto inj = new kore_pattern;
+  inj->ptr_ = kllvm::bindings::make_injection(term->ptr_, from->ptr_, to->ptr_);
   return inj;
 }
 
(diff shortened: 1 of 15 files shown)


Baltoli:
I have also tested the booster's integration tests with this change applied and there is no breakage to the C bindings.

tothtamas28 (commented):
Looks good in terms of functionality.

Python tests: https://github.com/runtimeverification/pyk/pull/736

dwightguth (changes requested):
It's very difficult to follow these changes because they are a mix of a refactoring that moves a lot of code around, and new features. I'm concerned that despite my best efforts, I still wasn't really able to determine what code was "new" and review it effectively.

I think you need to refactor the history of this PR to more obviously break it into two components: the moving around of existing code, and the addition of new code.

bindings/c/lib.cpp
```diff
@@ -185,14 +186,8 @@ kore_pattern *kore_pattern_new_token_with_len(
...
-  auto inj_sym = kore_symbol_new("inj");
-  kore_symbol_add_formal_argument(inj_sym, from);
-  kore_symbol_add_formal_argument(inj_sym, to);
-
-  auto inj = kore_composite_pattern_from_symbol(inj_sym);
-  kore_composite_pattern_add_argument(inj, term);
-
-  kore_symbol_free(inj_sym);
+  auto inj = new kore_pattern;
+  inj->ptr_ = kllvm::bindings::make_injection(term->ptr_, from->ptr_, to->ptr_);
```
Baltoli, line 189: @dwightguth As an example of how to look at the changes in this diff, the `kllvm::bindings::make_injection` function contains code that is _equivalent_ to the C on the LHS here, but using the underlying C++ structures directly. All the changes in this commit are of that form.

bindings/core/src/core.cpp
```diff
@@ -1 +1,83 @@
...
+    std::shared_ptr<KORESort> to) {
+  auto inj_sym = KORESymbol::Create("inj");
+
+  inj_sym->addFormalArgument(from);
+  inj_sym->addFormalArgument(to);
+
+  auto inj = KORECompositePattern::Create(std::move(inj_sym));
+  inj->addArgument(term);
+
+  return inj;
```
Baltoli, line 26: The restated C++ code from my other comment is here.

dwightguth (approved):

PR #905, opened 2023-11-27
Expose function evaluation to bindings

Baltoli:
This PR adds a new feature to the backend's binding API that allows for function symbols to be evaluated on a set of AST pattern arguments, with the result returned as a pattern also. This is motivated by the proof generation project; the execution trace format generated by the backend to summarise a concrete execution can contain the labels and arguments for function call events, but not the functions' return values without breaking tail-call optimisation.

Most of the prerequisite work for this PR was done in #908 and #911; the actual changes here are quite small:
* Adds a sort-aware term-to-pattern reification method that uses the new `rawTerm{}` wrapper.
* Marshals arguments and returns between pattern and term level in the core bindings API.
* Adds a test case demonstrating a few simple cases for this feature; as ever the bulk of the testing ought to be performed in Pyk.


include/kllvm/bindings/core/core.h
```diff
@@ -37,6 +37,10 @@ std::shared_ptr<kllvm::KOREPattern> simplify(
 bool is_sort_kitem(std::shared_ptr<kllvm::KORESort> const &sort);
 bool is_sort_k(std::shared_ptr<kllvm::KORESort> const &sort);
 
+std::shared_ptr<kllvm::KOREPattern> evaluate_function(
+    std::string const &label,
+    std::vector<std::shared_ptr<kllvm::KOREPattern>> const &args);
```
tothtamas28, line 42: Did you consider passing a `CompositePattern`?

tothtamas28: Right, since part of the input is already structured (`args`), I guess having a single structured parameter would make this a bit cleaner.

Robertorosmaninho (approved):
The C++ code looks good to me!

test/python/test_evaluate.py
```diff
@@ -0,0 +1,50 @@
...
+
+class TestEvaluate(unittest.TestCase):
+
+    def test_ctor(self):
+        left = kllvm.parser.Parser.from_string('Lblfoo{}(\\dv{SortInt{}}("23"))').pattern()
+        right = kllvm.parser.Parser.from_string('Lblbar{}(\\dv{SortInt{}}("56"))').pattern()
+
+        call = kllvm.ast.CompositePattern('Lblf')
+        call.add_argument(left)
+        call.add_argument(right)
```
tothtamas28, line 19: Now with the change to the API the input can be parsed in one step.

tothtamas28 (commented):
Python API looks good.

Tests in `pyk`: https://github.com/runtimeverification/pyk/pull/784

include/kllvm/bindings/core/core.h
```diff
@@ -37,6 +37,9 @@ std::shared_ptr<kllvm::KOREPattern> simplify(
 bool is_sort_kitem(std::shared_ptr<kllvm::KORESort> const &sort);
 bool is_sort_k(std::shared_ptr<kllvm::KORESort> const &sort);
 
+std::shared_ptr<KOREPattern>
+evaluate_function(std::shared_ptr<KORECompositePattern> term);
```
tothtamas28, line 41: Btw, what's the difference between this function and `simplify_pattern`? Is this just a special case thereof?

Baltoli: It's a special case; simplification works also for terms like `c(f())` where `c` is a constructor - this API is intended _only_ to directly evaluate functions and so it skips a couple of checks in the backend to get directly to the right code path.

PR #907, opened 2023-11-28
Add raw term symbol to definitions

Baltoli:
This is an alternative formulation to https://github.com/runtimeverification/k/pull/3829; that PR contains the original context for why this change is necessary. Rather than adding a regular symbol to the K standard library source and using a check in the frontend to ensure that user code doesn't mention it, we directly emit a symbol into each definition with a name that the frontend cannot reference.

We also bump the version number for the proof hint format so that this non-standard symbol must be explicitly accepted by the Python client code.


Baltoli:
I have tested this against the downstream K regression test suite and as expected, this does not break anything.

PR #911, opened 2023-12-01
Implement emitted return sort table

Baltoli:


Baltoli:
> LGTM! One question, though, as we're emitting more code, how does this modification affect the performance of the uses of our generated runtime lib?

I'd be very surprised if this caused any kind of meaningful performance change; the new code-generation steps is one new O(symbols) among a set of similar steps and the data getting emitted is only a single pointer for each symbol (the sort names are already present in the binary).

Issue #788, opened 2023-06-26
LLVM backend failing to evaluate function


virgil-serbanuta:
For the example below, the LLVM backend ends execution with an error, outputting an unevaluated function, without any obvious reason for not evaluating it.

```
[Error] krun: ./a-kompiled/interpreter 
/tmp/.krun-2023-06-26-17-49-17-Oq7lqoxprx/tmp.in.zDrsIYS8SO -1 
/tmp/.krun-2023-06-26-17-49-17-Oq7lqoxprx/result.kore
g ( ListItem ( wrap ( b"a" ) ) )
```

All or most of the elements in the semantics below seem to be needed for reproducing the issue (e.g. using a normal Map, or passing a list directly to `g`, without producing it with `Set2List(keys(...))` do not generate the error above).

To reproduce:
a.k
```
module A
    imports BYTES
    imports COLLECTIONS
    imports K-EQUAL-SYNTAX
    imports LIST

    syntax List ::= f ( MapBytesToBytes ) [function, symbol, klabel(myF)]
                  | g ( List )            [function, symbol, klabel(myG)]
    rule f(M) => g(Set2List(keys(M)))

    rule g(.List) => .List
    rule g(ListItem(_:WrappedBytes) L:List) => g(L:List)

    syntax KItem  ::= a(MapBytesToBytes)
                    | b(List)

    rule a(M) => b(f(M))

    syntax WrappedBytes ::= wrap(Bytes)  [symbol, klabel(wrapBytes)]

    syntax MapBytesToBytes  [hook(MAP.Map)]
    syntax MapBytesToBytes ::= MapBytesToBytes MapBytesToBytes
          [ left, function, hook(MAP.concat), klabel(_MapBytesToBytes_),
            symbol, assoc, comm, unit(.MapBytesToBytes), element(_Bytes2Bytes|->_),
            index(0), format(%1%n%2)
          ]
    syntax MapBytesToBytes ::= ".MapBytesToBytes"
          [ function, total, hook(MAP.unit),
            klabel(.MapBytesToBytes), symbol, latex(\dotCt{MapBytesToBytes})
          ]
    syntax MapBytesToBytes ::= WrappedBytes "Bytes2Bytes|->" WrappedBytes
          [ function, total, hook(MAP.element),
            klabel(_Bytes2Bytes|->_), symbol,
            latex({#1}\mapsto{#2}), injective
          ]

[...]

PR #917, opened 2023-12-05
Add user map element symbol

Baltoli:
https://github.com/runtimeverification/llvm-backend/issues/788


virgil-serbanuta:
The solution should allow the semantics to also work in the Haskell backend.

PR #918, opened 2023-12-05
Parser for proof hint traces

theo25:
This PR modifies the validator tool and class for proof hint traces into a parser. The parser both validates a given binary proof hint trace and generates a data structure  for said trace that reflects the one used by the [Python parser found here](https://github.com/runtimeverification/proof-checker/blob/main/generation/src/proof_generation/llvm_proof_hint.py).

In a future PR, we will add Python bindings for the data structures and the parser API.


gtrepta (commented):
It makes sense to rename `ProofTraceValidator` to `ProofTraceParser` now.

Can we test this somehow? What if the verbose option in `kore-proof-trace` printed the trace from the parsed `LLVMRewriteTrace` structure after parsing instead of the parser printing it during parsing, and then we diff that against an expected output?

theo25:
@gtrepta I have made the proposed changes. I modified the lit test to check the verbose output but I haven't modified the parser to print the trace after (as opposed to while) parsing since it does not make any difference in checking the output. If you prefer the printing to happen after parsing anyway, I can make that change as well.

theo25:
I have made the change of printing the whole data structure after parsing. I have also replaced an unordered map with an ordered map to avoid non-determinism in the order of printing the map's elements.

include/kllvm/binary/ProofTraceParser.h
```diff
@@ -0,0 +1,626 @@
+#ifndef PROOF_TRACE_VALIDATOR_H
+#define PROOF_TRACE_VALIDATOR_H
```
Baltoli, line 2:
```suggestion
#ifndef PROOF_TRACE_PARSER_H
#define PROOF_TRACE_PARSER_H
```

include/kllvm/binary/ProofTraceParser.h
```diff
@@ -0,0 +1,626 @@
...
+
+namespace detail {
+
+constexpr uint64_t word(uint8_t b) {
+  uint64_t w = uint64_t(b);
+  for (auto i = 0; i < 7; ++i) {
+    w = (w << 8u) | w;
+  }
+  return w;
+}
```
Baltoli, line 22: Can we unify these two implementations? https://github.com/runtimeverification/llvm-backend/blob/e8c7319292ad82f7ca93e0faa4aa99656104cc17/lib/codegen/ProofEvent.cpp#L25-L32

Baltoli (changes requested):
A couple of minor comments, and one top-level one - can we add an integration test that actually validates some simple properties of the parsed trace? The traces are pretty big, but we could add a test case for a simpler rewrite system. For example:
```k
syntax Foo ::= a() | b() | c()
rule a() => b()
rule b() => c()
```
then the test would validate that for the program `a()` we see precisely 3 configurations, and that the top-level rewrites for these rules occur once each in the right order.

The best way to do this is probably to add a new executable that links against the parser library, then reads a trace file and tests a set of hard-coded properties like these.

include/kllvm/binary/ProofTraceParser.h
theo25: good catch! I pushed a commit to unify them.

theo25:
I have addressed all the comments, @Baltoli if you could take another look

tools/kore-proof-trace-test/main.cpp
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  // check that the trace after the initial configuration is 4 events long
+  if (Trace.trace.size() != 4u) {
+    return 1;
+  }
+
+  // check that the first event is the rewrite a() => b()
+  const auto *Rule1 = dynamic_cast<LLVMRuleEvent const *>(
+      Trace.trace[0].getStepEvent().get());
```
Baltoli, line 39: Prefer `std::dynamic_pointer_cast` rather than `dynamic_cast` on the contained pointer

tools/kore-proof-trace-test/main.cpp
```diff
@@ -0,0 +1,68 @@
...
+    cl::Positional, cl::desc("<input file>"), cl::Required,
+    cl::cat(KoreProofTraceTestCat));
+
+int main(int argc, char **argv) {
+  cl::HideUnrelatedOptions({&KoreProofTraceTestCat});
+  cl::ParseCommandLineOptions(argc, argv);
+
+  ProofTraceParser Parser(false, 3u);
+  LLVMRewriteTrace Trace;
+  if (!Parser.parse_proof_trace(InputFilename, Trace)) {
```
Baltoli, line 22: Does this API need to be a pass-by-ref-and modify? Can we have `parse_proof_trace` return a `std::optional<LLVMRewriteTrace>` instead?

lib/binary/ProofTraceParser.cpp
```diff
@@ -0,0 +1,97 @@
+#include <kllvm/binary/ProofTraceParser.h>
+
+namespace kllvm {
+
+#define INDENT_SIZE 2u
```
Baltoli, line 5:
```suggestion
constexpr auto indent_size = 2u;
```
+ the downstream name changes to references to `INDENT_SIZE`

lib/binary/ProofTraceParser.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+void LLVMRewriteEvent::printSubstitution(
+    std::ostream &Out, unsigned indent) const {
+  std::string Indent(indent * INDENT_SIZE, ' ');
+  for (const auto &p : substitution) {
+    Out << Indent << p.first << " = kore[" << p.second.second << "]"
+        << std::endl;
```
Baltoli, line 12: My preference for this kind of code - at least for the moment - is to continue _printing_ to an ostream, but to use `fmt` to do the actual formatting. Something like:
```c++
Out << fmt::format("{}{} = kore[{}]\n", Indent, p.first, p.second.second);
```

Baltoli (approved):
A few nitpicks but this seems largely very good to me; feel free to merge once my comments are addressed! Thanks Theo

PR #920, opened 2023-12-07
Adopt scalafmt for Scala code style

Baltoli:
In https://github.com/runtimeverification/k/pull/3841, we adopted the `scalafmt` tool to auto-format the Scala code in the K frontend. This PR adopts a precisely identical formatting setup for the LLVM backend's Scala code.


Baltoli:
Blocked on the maven dependency update job being out of date: #921

Robertorosmaninho (approved):
Overall LGTM! I would expect @Scott-Guest to approvals for a more precise review of the config files!

matching/pom.xml
```diff
@@ -31,6 +31,14 @@
...
   </repositories>
 
+  <properties>
+    <scala.majorVersion>2.12</scala.majorVersion>
+    <scala.minorVersion>18</scala.minorVersion>
+    <scala.version>${scala.majorVersion}.${scala.minorVersion}</scala.version>
+    <spotless.version>2.41.1</spotless.version>
+    <scalafmt.version>3.7.17</scalafmt.version>
+  </properties>
+
```
Scott-Guest, line 41: As far as I understand, all the properties are automatically inherited from the parent POM, so we don't need to redeclare them here.

matching/pom.xml
```diff
@@ -59,6 +67,16 @@
...
+    <dependency>
+      <groupId>com.diffplug.spotless</groupId>
+      <artifactId>spotless-maven-plugin</artifactId>
+      <version>${spotless.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.scalameta</groupId>
+      <artifactId>scalafmt-core_${scala.majorVersion}</artifactId>
+      <version>${scalafmt.version}</version>
+    </dependency>
```
Scott-Guest, line 79: Same thing here - dependencies will automatically be pulled in from the parent POM and don't need to be redeclared

Scott-Guest (approved):
LGTM overall, but see my comments about the property and dependency declarations

gtrepta (approved):

Baltoli:
Thanks @Scott-Guest - I didn't realise Maven worked like that!

Baltoli:
I have addressed the duplicated configuration elements; going to merge this now.

Scott-Guest:
@Baltoli Hmm, I am pretty sure this is how Maven works, and it passes locally for me when removing the re-declarations. The Maven documentation explicitly states that properties and dependencies are inherited: https://maven.apache.org/pom.html#inheritance

I see that CI is failing due to the undefined properties, so I might be missing something though!

Baltoli:
I seem to have found the correct incantation to get Maven to inherit the properties and configure the plugin now!

PR #924, opened 2023-12-11
Basic setup of clang-tidy

Baltoli:
This PR continues our thread of work on improving code quality and automated tooling by introducing a new `clang-tidy` check that runs on the backend's C++ code. Because `clang-tidy` has so many available checks, this initial PR sets up only the basic `clang-analyzer-*` set. In future PRs, we will be able to introduce and fix more checks incrementally using this setup.

The structure of the changes in this PR are as follows:
1. In https://github.com/runtimeverification/llvm-backend/pull/924/commits/7bc21d9f393e821f69447933ed4813fe8d14e756, a new CI workflow is set up to run the tool inside a Dockerized runner. All warnings generated are silenced explicitly in the config file. Additionally, a runner script is added to make it more convenient to run the tool locally.
2. In the subsequent commits (https://github.com/runtimeverification/llvm-backend/pull/924/commits/ce5538a66abea11ffaff3aeacaa763818994b5fa ... https://github.com/runtimeverification/llvm-backend/pull/924/commits/d8aff83af93b50e031cb88ac5d0481f5693371ce), one check per commit is fixed. The diff to `.clang-tidy` in each commit identifies the check being fixed, and the changes to the code in each case are pretty self-explanatory.

Some notes:
* We pin the version of the tool run in CI to `clang-tidy-15` as that's what's available in the Ubuntu Jammy package repos. I _think_ newer versions should be backwards compatible, in that any warning thrown by an older version will also be thrown by the newer one.
* I've kept the local runner script pretty minimal; it should work on macOS with `llvm@15` installed, or Ubuntu with `clang-tidy-15` installed. Give it a go on your own systems and make sure it works for you!
* There is one warning thrown inside `immer`; I have left the suppression on for this warning, and in general I'm disinclined to fix this explicitly given it's in library code and doesn't seem to be breaking anything. I will probably add a `NOLINT` directive to our fork of `immer` down the line.
* I'm obviously happy to bikeshed future checks being added or deleted, but I think the default set in this particular PR are  pretty uncontroversial.


Baltoli:
Note that CI is failing because of https://github.com/runtimeverification/llvm-backend/issues/926; this is otherwise ready to review.

Baltoli:
I will wait until this is merged to do so, but the following sets of check suppressions can be used to start off applying a few useful families of checks we may want to apply later on:

```
  modernize-*
  -modernize-avoid-c-arrays
  -modernize-concat-nested-namespaces
  -modernize-deprecated-headers
  -modernize-loop-convert
  -modernize-macro-to-enum
  -modernize-raw-string-literal
  -modernize-redundant-void-arg
  -modernize-return-braced-init-list
  -modernize-use-auto
  -modernize-use-default-member-init
  -modernize-use-emplace
  -modernize-use-nullptr
  -modernize-use-trailing-return-type
  -modernize-use-using
```

Scott-Guest (approved):
LGTM, thanks for getting this set up!

Baltoli:
Merging this as the consensus in the K / LLVM meeting was positive; future checks might need more bikeshedding.

Robertorosmaninho (commented):
LGTM! Sorry for the late review!

Issue #916, opened 2023-12-05
Use `rawTerm` rather than injections in term simplification


Baltoli:
Currently, we inject non-block symbols into `KItem` at the bindings' interface to the term simplifier so that they can be printed and serialized. This is not sound in general, as clients cannot tell with 100% certainty when such an injection should be stripped off a resulting term.[^1]

In #907, we added a new symbol that replaces these spurious injections (`rawTerm{}`). This new feature is used to implement the "evaluate function symbol" binding (https://github.com/runtimeverification/llvm-backend/pull/905), which has a very similar interface to the term simplifier. We should update the term simplifier to use `rawTerm{}` rather than injections.

[^1]: I believe the booster has some logic that looks symbols up in the definition to see when the injection can be stripped.

PR #931, opened 2023-12-12
`rawTerm` for term simplification

gtrepta:
fixes: #916


Baltoli:
The change here is pretty much exactly what I was expecting to see, thanks Guy - can you run the Pyk and Booster integration test suites to see what (if anything) will need to be updated? My recollection is that:
* there are some Pyk tests that will need to be updated similarly to the tests you've changed here
* the booster _should_ just work but I don't know precisely the deserialization code path that it uses; it might be the case that some Haskell code is needed to strip the `rawTerm{}`

cc @goodlyrottenapple @tothtamas28

gtrepta:
The `pyk` integration tests did indeed have a case that needed to be updated. I was able to fix it very quickly, so that won't be an issue.

The booster backend is failing as well, but I'm not sure how to fix it. It plainly states `Unknown symbol "rawTerm"`. So, it needs to use some functionality to strip it there. A grep for `rawTerm` in the repo doesn't show anything, though, so I guess it doesn't exist yet.

Baltoli:
Blocked on: https://github.com/runtimeverification/hs-backend-booster/issues/431

PR #933, opened 2023-12-18
Enable clang-tidy modernization checks

Baltoli:
This follows up on #924 by enabling the modernization set of `clang-tidy` checks. As in the original PR, each individual commit largely mechanically fixes simple issues related to a single check. The changes here are stylistic ones that ensure we are using modern C++ features appropriately in the backend's code, and aren't leaning on legacy / C-style practices.

There are a couple of commits that are not directly mechanical changes:
- https://github.com/runtimeverification/llvm-backend/pull/933/commits/ed23fddbd90898e7c54e1ed37d4352689b294fef allows the script that drives `clang-tidy` to receive additional CLI args; doing so enables the use of the auto-fixer.
- https://github.com/runtimeverification/llvm-backend/pull/933/commits/b74fe6ac4456c0025975d4aed6ba9216dec01051 refactors some formatting code to use FMT rather than legacy printing functions.
- https://github.com/runtimeverification/llvm-backend/pull/933/commits/e13d666f501856d0b38bd77a2ad307b7fd1f2b50 corrects the formatting of code generated by the auto-fixer.

As ever, I'm open to bikeshedding the specific changes introduced here - it's really easy to change whether a specific check is enabled or disabled. For instance, I have left the [trailing return type](https://clang.llvm.org/extra/clang-tidy/checks/modernize/use-trailing-return-type.html) check disabled as it's not the code style we're using.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -2,7 +2,6 @@ Checks:
   clang-analyzer-*
   -clang-analyzer-cplusplus.PlacementNew
   modernize-*
-  -modernize-avoid-c-arrays
   -modernize-use-trailing-return-type
 
 WarningsAsErrors:
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -3,7 +3,6 @@ Checks:
   -clang-analyzer-cplusplus.PlacementNew
   modernize-*
   -modernize-avoid-c-arrays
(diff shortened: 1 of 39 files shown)


Scott-Guest (approved):
Great work, strongly in favor of all these changes

PR #935, opened 2023-12-18
Enable clang-tidy bugprone checks

Baltoli:
This is another PR in my ongoing series to enable more clang-tidy checks for the codebase. As previously, each commit in this PR fixes the issues flagged by a single check, and can / should be reviewed independently.

### DIFF
--- a/runtime/io/io.cpp
+++ b/runtime/io/io.cpp
@@ -617,6 +617,7 @@ list hook_KREFLECTION_argv() {
 
   list l{};
 
+  // NOLINTBEGIN(*-sizeof-expression)
   for (int i = 0; i < llvm_backend_argc; i++) {
     stringbuffer *buf = hook_BUFFER_empty();
     buf = hook_BUFFER_concat_raw(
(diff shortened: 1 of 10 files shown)


runtime/strings/strings.cpp
```diff
@@ -17,8 +17,6 @@
 
 extern "C" {
 
-#define KCHAR char
```
Scott-Guest, line 20: Nice!

Scott-Guest (approved):

runtime/strings/strings.cpp
Baltoli: There's more of this to be removed but I'm going to do it incrementally

runtime/io/io.cpp
```diff
@@ -628,6 +629,7 @@ list hook_KREFLECTION_argv() {
     memcpy(b->children, &str, sizeof(str));
```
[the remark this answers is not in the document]
Baltoli: Good point; addressed!

PR #936, opened 2023-12-18
Enable clang-tidy readability checks

Baltoli:
This is another PR in my ongoing series to enable more clang-tidy checks for the codebase. As previously, each commit in this PR fixes the issues flagged by a single check, and can / should be reviewed independently.

As ever, I'm happy to bikeshed any of the specific changes in this PR. Note that for [`readability-function-cognitive-complexity`](https://github.com/runtimeverification/llvm-backend/pull/936/commits/c2b0de149776d1af54a519e81ecf53923dfd1ec1), I've not actually refactored any existing code in the interest of keeping this diff minimal, but the check will prevent us from introducing new code that exceeds the complexity threshold.

### DIFF
--- a/runtime/util/match_log.cpp
+++ b/runtime/util/match_log.cpp
@@ -52,8 +52,9 @@ void addMatchFunction(char *debugName, char *function, void *result, ...) {
   std::vector<void *> args;
   while (true) {
     void *arg = va_arg(ap, void *);
-    if (!arg)
+    if (!arg) {
       break;
+    }
     args.push_back(arg);
   }
 
--- a/runtime/util/match_log.cpp
+++ b/runtime/util/match_log.cpp
@@ -12,7 +12,7 @@ extern "C" void *getStderr(void) {
 static std::vector<MatchLog> matchLog;
 
 void **getMatchFnArgs(MatchLog *log) {
-  return &log->args[0];
+  return log->args.data();
 }
 
 extern "C" {
@@ -21,7 +21,7 @@ void resetMatchReason(void) {
 }
 
 MatchLog *getMatchLog(void) {
-  return &matchLog[0];
+  return matchLog.data();
 }
 
 size_t getMatchLogSize(void) {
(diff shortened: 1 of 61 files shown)


runtime/util/match_log.cpp
```diff
@@ -40,12 +40,13 @@ void addMatchSuccess(void) {
        nullptr});
 }
 
-void addMatchFailReason(void *subject, char *pattern, char *sort) {
+void addMatchFailReason(void *subject, const char *pattern, const char *sort) {
```
Scott-Guest, line 43: Looks like we use both `const char *` and `char const*` in various places.

We should consider setting set `QualifierAlignment: Right` to clean this up: https://clang.llvm.org/docs/ClangFormatStyleOptions.html#qualifieralignment

Scott-Guest (approved):
LGTM! Left a few comments that would be nice to address in a follow-up PR.

I'm also strongly in favor of refactoring all the `NOLINTNEXTLINE(*-cognitive-complexity)` functions at some point.

Baltoli:
Yep, I'll fix those comments in follow-ups. The naming convention in particular needs some bikeshedding first, but I plan to get round to it.

runtime/util/match_log.cpp
Scott-Guest: @gtrepta `const char *` and `char const *` are equivalent - both are a pointer to a constant `char`. I think the common lingo here is "west const" versus "east const". 

(Both are distinct from `char * const` though - a constant pointer to a `char`.)

The warning is worth bikeshedding about, but IMO it's not a big concern:
- I've used it on large production codebases before with no issues
- Looking through the git blame history that adds this option `clang-format`, AFAICT the only known issue is that lower-case macros may be mis-identified as types, meaning that it could rewrite
```
#define intptr int *
const intptr a;
```
to
```
intptr const a;
```
which is a semantic change. This doesn't seem to appear anywhere in our codebase, and if it does, we should just change it to a typedef instead of a macro anyways.

Baltoli: I think that `clang-tidy` will actually catch usages of lower-case macros like that anyway, so I think we can make this change safely.

PR #939, opened 2023-12-18
Python bindings for the proof trace API

theo25:
This PR adds python bindings for the proof trace parser.

The PR also contains two more changes, that became apparent while I was working on the bindings:
- The `LLVMRewriteTrace` object becomes a `class` rather than a `struct` and we provide bindings for that class.
- The `ProofTraceParser` object does not accept the expected hints version as an argument to its constructor, but rather stores this information in a static constant. This made more sense since the parser currently supports only the latest hints version. In the future, if we want the parser to support older versions we can revisit this.


Baltoli (approved):
LGTM; thanks Theo!

Issue #922, opened 2023-12-08
Potential memory leak in Python binding for `simplify_pattern`


tothtamas28:
Running `simplify_pattern` iteratively results in monotonically increasing memory consumption. 

[test_simplify.py.txt](https://github.com/runtimeverification/llvm-backend/files/13614189/test_simplify.py.txt)

Copy the file to `pyk` repo root, then run

```
poetry run python3 test_simplify.py
```

Baltoli:
We fixed a couple of minor memory leaks in https://github.com/runtimeverification/llvm-backend/pull/924; I confirmed that this particular reproduction still happens after doing so.

Baltoli:
I have created a smaller C reproduction of the issue that avoids Python entirely in the interests of making it easier to use traditional memory profiling tools; this more minimal reproduction still exhibits the same monotonic increase in memory usage. It needs a couple of paths to be set up but isn't too onerous to run.

`simplify.c`
```c
#include <kllvm-c/kllvm-c.h>
#include <stdio.h>
#include <stdlib.h>

int main() {
  kore_pattern *pat = kore_pattern_parse(
      "Lbl'UndsPlus'Int'Unds'{}(\\dv{SortInt{}}(\"1\"),\\dv{SortInt{}}(\"2\")"
      ")");

  kore_sort *sort = kore_composite_sort_new("SortInt");

  for (int i = 0;; ++i) {
    if (i % 10000 == 0) {
      fprintf(stderr, ".");
    }

    char *data;
    size_t size;
    kore_simplify(pat, sort, &data, &size);
    free(data);
  }

  kore_sort_free(sort);
  kore_pattern_free(pat);
```

Baltoli:
Rewrite step reproduction; from the Pyk root as usual:

`src/tests/integration/k-files/count.k`
```k
module COUNT
  imports INT
  imports BOOL
  
  syntax C ::= count(Int)
  rule count(I) => count(I -Int 1) requires I =/=Int 0
  rule count(I) => true requires I ==Int 0
endmodule
```

`init.kore`
```
[...]

PR #945, opened 2023-12-20
Enable fixes for memory leaks in bindings

Baltoli:
This PR fixes a long-standing issue in the backend's binding code, which is that the call to `initStaticObjects()` when the Python module is _bound_ does not affect the process that actually runs the backend interpreter. A consequence of this is that GMP is never told to use our GC'ed memory allocators, and so any code using integer operations ends up leaking lots of memory when it runs for a long time.

To fix Python code that uses the bindings:
* Call `init_static_objects` on application startup
* If an application _only_ does function evaluation or simplification, you may also want to call `free_all_gc_memory` occasionally as the backend will not do so autonomously unless a rewrite step is taken.

Relevant:
* #770
* #922 (has reproduction instructions)

### DIFF
--- a/bindings/python/runtime.cpp
+++ b/bindings/python/runtime.cpp
@@ -40,7 +40,8 @@ PYBIND11_DECLARE_HOLDER_TYPE(T, raw_ptr<T>, true);
  */
 
 extern "C" {
-void initStaticObjects();
+void initStaticObjects(void);
+void freeAllKoreMem(void);
 block *take_steps(int64_t, block *);
 void *constructInitialConfiguration(const KOREPattern *initial);
 }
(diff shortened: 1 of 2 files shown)


bindings/python/runtime.cpp
```diff
@@ -97,7 +101,5 @@ void bind_runtime(py::module_ &m) {
 }
 
 PYBIND11_MODULE(_kllvm_runtime, m) {
-  initStaticObjects();
-
```
tothtamas28, line 101: If I understand correctly, the root cause of all these memory leaks is that this is not the right place to call this initializer function, but instead it has to be called from the imported module in Python. Is there a simple explanation for why that is?

tothtamas28 (commented):
Tested:

* `InternalTerm.step`
* `simplify_pattern` (with `free_all_gc_memory` in each iteration)
* `evaluate_function` (with `free_all_gc_memory` in each iteration)

And the memory leak seems to be gone. :rocket:

tothtamas28 (approved):

bindings/python/runtime.cpp
Baltoli: I don't have a satisfactory explanation for that, unfortunately. I suspect the answer probably lies somewhere in the internals of how Pybind11 actually works 🤷

Baltoli: (but yes, your understanding matches mine)

PR #946, opened 2023-12-21
Use python rather than perl for profiling

Baltoli:
In https://github.com/runtimeverification/llvm-backend/pull/833 we changed from the built-in date command to perl for getting the system time in ms; this caused some weird interactions with the OS when installed via `kup`: https://github.com/runtimeverification/kup/issues/80

This PR drops our use of `perl` and instead uses an equivalent `python3` command, which we already depend on when building the backend.

### DIFF
--- a/bin/utils.sh
+++ b/bin/utils.sh
@@ -7,7 +7,7 @@ profile=${profile:-false}
 
 
 time_now_ms () {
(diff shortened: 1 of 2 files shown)


anvacaru (approved):
With this change I no longer get the warnings. Approved!

PR #947, opened 2024-01-11
Update to C++-20

Baltoli:
Core support for C++20 is now good enough across the versions of Clang that we support building the backend with that I'm happy to set it as the default version. Doing so will get us some nice refactoring opportunities in the future.


Baltoli:
There are a couple more fixes to work through `clang-tidy` before this is ready: https://github.com/runtimeverification/immer/pull/3

PR #954, opened 2024-01-17
Set right-alignment for qualifiers

Baltoli:
As suggested in https://github.com/runtimeverification/llvm-backend/pull/936, this PR sets the appropriate `clang-format` option to right-align qualifiers in the backend (that is, writing `T const&` rather than `const T&`).

This PR is fully mechanical and introduces no manual changes to the code.


.clang-format
```diff
@@ -20,4 +20,5 @@ SpaceBeforeCpp11BracedList: false
 DerivePointerAlignment: false
 PointerAlignment: Right
 ReflowComments: false
+QualifierAlignment: Right
```
Scott-Guest, line 23: Looking at the [source code](https://github.com/llvm/llvm-project/blob/fca6992be1f272f5e997bd510ca03c9389550c13/clang/lib/Format/Format.cpp#L1072C30-L1072C60), setting 
```
QualifierAlignment: Right
```
is equivalent to
```
QualifierAlignment: Custom
QualifierOrder: ['type', 'const', 'volatile']
```
meaning it doesn't enforce a particular order for qualifiers like `static`, `inline`, `constexpr`, etc.

I would personally be in favor of using `Custom` and fully-specifying the `QualifierOrder` for all supported qualifiers listed here: https://clang.llvm.org/docs/ClangFormatStyleOptions.html#qualifierorder

Scott-Guest (approved):
I'm okay if we want to merge this as-is, but see my comment.

.clang-format
Baltoli: How do you feel about `static constexpr inline friend type const volatile restrict`?

Scott-Guest: Works for me!

PR #955, opened 2024-01-17
Interpret C++ exceptions in C bindings

### LINKED ISSUE #953: Interpret exceptions in C bindings
The booster is experiencing difficulties where exceptions that are thrown in C++ hook implementations (e.g. division by zero, logarithm of zero, ...) bring the entire process down when the exception crosses the C++-C boundary.

We should add error handling code to the simplification bindings (`kore_simplify[_bool]`) that catches these exceptions and propagates some C-compatible error information to the caller.

Baltoli:
This PR amends the interface of the simplification routines in the C bindings to allow for error information to be safely returned to calling clients without crashing the host process.

The changes are as follows:
- Introduce a new opaque `kore_error` type to the C bindings, with appropriate bindings.
- **Breaking change** to the interfaces of `kore_simplify`, `kore_simplify_bool` and `kore_simplify_binary` to take an optional additional `kore_error` parameter. If a C++ exception is thrown during these bindings, it will be caught and used to populate the error parameter's fields. The booster will need to be updated as part of its K release job down the line.
- Add a test that catches an error that imitates the original reporting of this issue (invalid numeric value in the hooked arithmetic library).

Closes #953


Scott-Guest (approved):
LGTM!

PR #956, opened 2024-01-18
Minor cleanups to C bindings

Baltoli:
While working on these bindings, I noticed that we were not taking best advantage of C++ and were writing some quite C-like code in places. This PR cleans up the following things:
- Use aggregate constructors to build internal types, rather than assigning to members
- Use `unique_ptr`s to manage allocated intermediate objects
- Use a helper function to manage C strings


gtrepta (approved):
LGTM

Issue #957, opened 2024-01-18
Refactor side condition and rewrite events in proof traces


Baltoli:
Per @nishantjr on [Slack](https://runtimeverification.slack.com/archives/C05MF3KEWNM/p1705407588543239), there is an identified redundancy between the substitutions emitted into proof traces for rewrite and side condition events. Specifically, the substitution emitted for a side condition will _always_ be a subset of the substitution for the corresponding rule.

Nishant proposes the following changes to the proof trace format to eliminate the redundancy:
* Emit a side condition event for _every_ rule that matches (i.e. for rules without side conditions, a constant `true` result will be emitted).
* Do not emit the substitution and ordinal for side condition events.

The result of making this change will be that every rule event has a corresponding side condition event, and no substitutions are repeated redundantly.

The C++ parser for the trace format will need to be updated, along with a version bump to the format.

nishantjr:
Thanks Bruce! Note that this will greatly simplify #727

theo25:
@Baltoli @nishantjr It seems to me that the rule ordinal and the substitution are the only two pieces of information that we emit as part of a side condition hint. See here:
https://github.com/runtimeverification/llvm-backend/blob/2fdcb5953b4c40198a1ecaa8f91fae5b47d7100b/include/kllvm/binary/ProofTraceParser.h#L83-L95
https://github.com/runtimeverification/llvm-backend/blob/2fdcb5953b4c40198a1ecaa8f91fae5b47d7100b/include/kllvm/binary/ProofTraceParser.h#L42-L68

If that is the case indeed then isn't the task here equivalent to completely remove the generation of side condition hints and just assume by default that every rule event contains a corresponding side condition?

nishantjr:
Indeed. The way I’m looking at it now, is that they are “try-rewrite”, and
“rewrite-succeeded” events.

In the previous version, since the side condition only included the partial
substitution in the side condition event, it would have forced us to apply
the substitution separately from the rewrite-event substitution when
proving rule application. This would have made the proof generation more
fragile and force us to have a separate optimization/clean up phase.

PR #960, opened 2024-01-23
Removing emission of side condition hints

theo25:
This PR removes all the information previously carried by side condition events from the proof hint trace  because it has been determined that said information is redundant. It also bumps the proof hint trace version to 4. This is done according to the discussion in #957 .

### DIFF
--- a/bindings/python/ast.cpp
+++ b/bindings/python/ast.cpp
@@ -370,7 +370,7 @@ void bind_proof_trace(py::module_ &m) {
 
   [[maybe_unused]] auto llvm_side_condition_event = py::class_<
       LLVMSideConditionEvent, std::shared_ptr<LLVMSideConditionEvent>>(
-      proof_trace, "LLVMSideConditionEvent", llvm_rewrite_event);
+      proof_trace, "LLVMSideConditionEvent", llvm_step_event);
 
   py::class_<LLVMFunctionEvent, std::shared_ptr<LLVMFunctionEvent>>(
       proof_trace, "LLVMFunctionEvent", llvm_step_event)
--- a/include/kllvm/binary/ProofTraceParser.h
+++ b/include/kllvm/binary/ProofTraceParser.h
@@ -80,15 +80,14 @@ public:
   virtual void print(std::ostream &Out, unsigned indent = 0u) const override;
 };
 
-class LLVMSideConditionEvent : public LLVMRewriteEvent {
+class LLVMSideConditionEvent : public LLVMStepEvent {
 private:
-  LLVMSideConditionEvent(uint64_t _ruleOrdinal)
-      : LLVMRewriteEvent(_ruleOrdinal) { }
+  LLVMSideConditionEvent()
+      : LLVMStepEvent() { }
 
 public:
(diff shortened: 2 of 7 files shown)


Baltoli:
After this change, does the trace still contain the _result_ of checking the side condition of each rule?

theo25:
If I understand correctly, the check of a side condition is a function call and as such the code generator should instrument the code to generate the appropriate function call event for the trace. I am not sure if that is what you are asking.

Baltoli:
> You suggest that the rules that previously where paired with a side condition should be flagged as such

Not quite; I mean that every "rule matched" event should have a corresponding "side condition _result_" event - that event in many cases will just be `true`.

Baltoli:
Sorry, had to remind myself of the context and got mixed up - I realised that previously we had events indicating "side condition attempted but failed". Those events were _side condition events with no associated rule event_. I guess @nishantjr can confirm, but I presume the team only need the side conditions for rules that actually match?

theo25:
Now that you mention it, I think we still have side condition events with no associated rule event. At least, I can see such events in the [`imp-proof`](https://github.com/runtimeverification/llvm-backend/blob/master/test/output/imp-proof.out.diff) test we have in the repo. Do we still need to generate those? And if yes, do these events need the rule ordinal and substitution information?

nishantjr:
@theo25 

1. Its OK to have side condition events (let's call these `try-rule` events) without a corresponding rule event (let's call these `apply-rule` events) when the side condition returns false.
2. If rule doesn't match it's better to not emit both events. However, if there are simplification events that are emitted (e.g. to simplify the side condition to false) but we require that the `try-rule` event is emitted. Otherwise we do not know that we need to throw away the simplifications.

My understanding of the LLVM backend this is what I expect to be easiest to implement.

1. If a rule fails to match, neither event will be emitted
2. If a rule matches but the side condition simplifies to false, then a `try-rule` along with some simplifications will be emitted. No `apply-rule` event will be emitted.  
4. If a rule matches and the side condition simplifies to true, then `try-rule`, followed by some simplification events, followed by `apply-rule` will be emitted.

 In an ideal world the llvm backend wouldn't even emit hints for `2`. However, this would involve caching the hints  for the simplification in memory until the we know the result, and then emitting it. My understanding is that we don't want to do this. Instead, the proof generation will remove this hint.

theo25:
@nishantjr I agree with your analysis. So what kind of information should the `try-rule` event carry in cases `2` and `3` above. Given the discussion above, it seems that you do not need any extra information (i.e. the rule ordinal and substitution are both redundant and these were the only pieces of information for a side condition event). I will go ahead and implement it this way, but could you please confirm that this is the case?

nishantjr:
> @nishantjr I agree with your analysis. So what kind of information should the `try-rule` event carry in cases `2` and `3` above. Given the discussion above, it seems that you do not need any extra information (i.e. the rule ordinal and substitution are both redundant and these were the only pieces of information for a side condition event). I will go ahead and implement it this way, but could you please confirm that this is the case?

In cases 2 and 3, the `try-rule` event should have both the rule-ordinal and the substitution. I think the `apply-rule` will have no information attached. It will act as a sentinel/flag just indicating that the previous `try-rule` succeeded.

PR #962, opened 2024-01-26
Add `--use-k-semantics` flag to disable possibly unsound optimizations

### LINKED ISSUE #930: Unsound/invalid rewrite in the LLVM backend
The LLVM backend changes cells that should not change when rewriting.

Version:
```
kompile --version
K version:    v6.1.37-0-g9a8e220ade
Build date:   Jo dec 07 19:00:09 EET 2023
```

To reproduce:
a.k
```
module A-SYNTAX
    syntax KItem  ::= "a"
endmodule

module A
    imports A-SYNTAX
    imports BYTES
    imports INT-SYNTAX

    configuration
      <test>
        <k> $PGM </k>
        <mem> .Bytes </mem>
        <stuff> .Bytes </stuff>
      </test>

    syntax KItem ::= "b" | "c" | "d" | "e"

    rule <k> a => b </k>
        <mem> _ => b"alice" </mem>
    rule <k> b => c </k>
        <mem> B </mem>
        <stuff> _ => B </stuff>
    rule <k> c => d </k>
        <mem> B => replaceAtBytes(B, 0:Int, b"bob__") </mem>
endmodule
```

start:
```
a
```

command line:
```
kompile a.k && krun start
```

Output:
```
<test>
  <k>
    d ~> .
  </k>
  <mem>
    b"bob__"
  </mem>
  <stuff>
    b"bob__"
  </stuff>
</test>
```

Expected:
```
<test>
  <k>
    d ~> .
  </k>
  <mem>
    b"bob__"
  </mem>
  <stuff>
    b"alice"
  </stuff>
</test>
```

Baltoli:
This PR adds a flag to `llvm-kompile` that disables semantics used by the LLVM backend that are unsound in certain scenarios, but are broadly useful for performance. In the initial PR, the only such change is to make objects of sort `Bytes` copy-on-write rather than mutable.

The numerous edge cases from my initial attempt in #950 are retained as test cases here.

I will follow up with a PR to the frontend that passes this flag through correctly.

Fixes #930


Scott-Guest (approved):
The implementation LGTM, but I pretty strongly disagree with the defaults here.

The whole purpose of the backends is to implement K semantics, and the default assumption is that different backends agree on these semantics. A semantics-breaking optimization like mutable `Bytes` should only be available with an explicit flag, not be the default behavior.

If we really want mutable `Bytes` by default, we should introduce it as a separate `MutableBytes` type (or better a `Ref{T}`or `Buffer{T}` type).

Baltoli:
I agree with you philosophically in terms of the role of the backend in implementing K semantics; the issues I have with making the default immutable are:
* A reasonably small change for the one case we currently know about (Virgil's example in the WASM semantics) suddenly becomes a large breaking change for _every_ semantics that currently exists; anyone using the `BYTES` module will need to opt-in to a `--disable-k-semantics` flag lest their behaviour silently change.
* Identifying breakage is much harder if the issue is "my use of the `BYTES` module is slower than it should be on some programs", than if it's "I see an unsound rewrite", so I'm not super keen on making the new implementation a silent default.
* As you identify, a pure-K solution with a parametric type might be a better solution in the long run.

I'm open to being convinced otherwise, but currently my internal compass swings a bit closer to "keep the surface area of this change small" than it does to "keep K+backends philosophically clean". Certainly not an obvious decision though!

gtrepta (approved):
LGTM.

Re the defaults, I agree that it does make sense to make semantics breaking optimizations opt-in, but also that we should keep this change in a small surface area. Making a sizeable change to the defaults should involve a longer process of some sort (deprecation warnings, major/minor version bump to K, etc.)

Scott-Guest:
@Baltoli I don't want to waste too much time on this, so feel free to merge either way at your discretion. 

If you care to bikeshed about this though, while you point out some fair considerations, K's limited user base makes them less convincing to me. I think the effort / change here is actually pretty small:

- The only real-world K semantics where performance matters are the handful developed here at RV. We can easily submit a small PR adding a `--llvm-mutable-bytes` flag to each of them.
- If we don't use this flag for a semantics, or if someone outside RV says "BYTES is suddenly slow", it shouldn't be inordinately difficult to debug - just profile and see which commit introduced the regression.

I'd emphasize that one of the big goals mentioned in previous K planning meetings is "Turning around perception from long-term community members that K is unreliable, difficult, not useful in practice." In my experience getting ramped up on K, the "unreliable" perception isn't entirely wrong - while the core of the language works well, there are a number of buggy, dark corners like this which are minimally documented and/or only avoided due to folk-knowledge at RV. 

Choosing to make our tools unsound by default because of historical performance hacks definitely does not help this perception - particularly given that K is a formal verification tool where soundness is the whole point. We have a small user base currently, so we should take that as an opportunity to aggressively clean up K and deal with a bit of pain now to do the "philosophically correct" thing, rather than baking in weird corner cases like this which will surprise new users and experienced developers alike (as evidenced by @virgil-serbanuta submitting this issue in the first place) and become increasingly difficult to remove as time goes on.

Regardless of what we go with for now, we should definitely remove this flag and provide a separate type at some point.

@gtrepta I'm fine if we want to decide some way to roll out any change to the default slowly, but I don't know if it's 100% required given that we just need to add a flag to a few semantics here.

Baltoli:
Good points well made @Scott-Guest[^1]; thanks for spending the time thinking this all through - I'm convinced! The next steps are:
* We should introduce a no-op frontend flag to specify that mutable `BYTES` semantics are being used, and propagate it to the larger semantics (EVM, C, ...?).
* Flip the logic in this change, and add an extra function to the C bindings that exposes the boolean flag so the booster can warn in the presence of mutable `BYTES`.

[^1]: In particular, the point about the booster diverging is key. I can't be bothered to properly spin up a demo, but I think we're both imagining roughly the same reproducing example whereby you could get it to produce an unsound result via LLVM simplification.

Baltoli:
This PR is ready to merge now that the corresponding change has gone through to K.

Issue #963, opened 2024-01-29
Ignore `inj` symbols when emitting locations in proof hints


fiedlr:
*Problem*

Kore is sorted and the `inj` symbol allows to cast a subsort to its supersort wherever needed. The `inj` symbol is taken into account when generating (sub)term locations in proof hints. 

We are generating proofs in a sorted layer of AML, which is itself unsorted. Whenever a sort is a subsort of another sort, the term already can be used as is where the supersort is expected. The `inj` symbols are therefore redundant in AML. We just need to make sure to do sort checking whenever we do function applications.

Since `inj` symbols are redundant in AML with certain considerations and measures, we want to simply remove them. This will break the locations as given in proof hints.

*Request*

Could the backend ignore injections when generating locations for proof hints?

PR #964, opened 2024-01-30
Ignore injection symbols when tracking locations in the proof trace

gtrepta:
fixes: #963


Baltoli (approved):
The substance of the change looks like what I was expecting; could we maybe factor the repeated logic out though? Thanks guy!

Issue #925, opened 2023-12-11
Expose "safely evaluate partial function" binding


Baltoli:
It would be useful for the [booster to be able to use the LLVM backend to evaluate partial functions where all the function's arguments are concrete](https://github.com/runtimeverification/hs-backend-booster/issues/415); however, the current design of the function evaluator in the backend just ends up calling `abort` when a partial function is evaluated, which crashes the host process (`kore-rpc`).

We should implement a refactored version of [`abortWhenStuck`](https://github.com/runtimeverification/llvm-backend/blob/e8c7319292ad82f7ca93e0faa4aa99656104cc17/lib/codegen/Decision.cpp#L798-L848) that instead of crashing the process, returns a sentinel value[^1] to the caller. Note that the existing behaviour would be subsumed by this; we can implement it by calling the partial-safe version, then checking for the sentinel value and aborting if it's returned.

Clients that want the abort-safe behaviour can then have the internal evaluator exposed as a new binding.

[^1]: I vaguely recall that `nullptr` isn't valid here; perhaps we want to distinguish a particular term representation that signals failure.

goodlyrottenapple:
This was another point, recently, where an error type would maybe be helpful: https://github.com/runtimeverification/llvm-backend/blob/eeccaebaed784a40f17971bf9f1028b07c510537/runtime/util/ConfigurationParser.cpp#L36-L39

This instance wasn't a pressing issue, because the cause was a mismatch between the llvm library version and the kore file loaded in the booster, which is probably a good reason to completely abort execution, but perhaps might be better if the calling code was allowed to handle this error.

Baltoli:
That one's trickier, because calls to `getTagForSymbolName` are so ubiquitous throughout the generated code; the bound code isn't usually _directly_ calling it (if that were the case, propagating the error would be easier). We would probably need to redesign big chunks of the backend to have a global error recovery mechanism to handle it.

goodlyrottenapple:
Fair enough, this one probably wouldn't be that important

Baltoli:
We can probably implement this now, at least for the booster via the C API: #955

PR #967, opened 2024-02-05
Allow partial function evaluation to be caught by bindings

Baltoli:
This PR extends the backend's ability to recover from errors when partial functions are evaluated in the context of a bindings library. Rather then unconditionally crashing the entire host process, we throw an exception that the bindings code can catch and translate to a C-ABI error structure for the backend to deal with.

Previously, we implemented this behaviour for hooks (https://github.com/runtimeverification/llvm-backend/pull/955); this PR does the same for general partial function evaluation. The changes are as follows:
* Refactor (as discussed previously) the implementation of `finish_rewriting` from LLVM IR into C++.
* Add a new code-generation flag to enable the new error behaviour.
* Modify `llvm-kompile` to pass this flag when compiling a bindings library.
* Read the flag to change behaviour in the appropriate places in the runtime library.
* Add a test that the C bindings can safely try to evaluate undefined partial functions and catch an error when the evaluation fails.

Fixes https://github.com/runtimeverification/llvm-backend/issues/925


gtrepta (approved):
LGTM

PR #968, opened 2024-02-05
Use FILE * instead of filenames when passing file paths to serialization and printing commands

dwightguth:
Previously we were storing a filename in a global variable and then passing it to a bunch of functions. Each of these functions then had to call fopen and fclose, which was quite expensive. This PR changes the behavior so instead we once open a file with fopen at the start of rewriting, then call a bunch of functions which take a `FILE *` as an argument, then call fclose once at the end. This gets much better buffering behavior and reduces the number of syscalls considerably.

When measuring the performance of  this code against a "sum to 10k" imp program with proof hint generation enabled, we saw a time reduction from 3.5m to 2m.


gtrepta (approved):

bindings/python/runtime.cpp
```diff
@@ -96,7 +96,9 @@ void bind_runtime(py::module_ &m) {
       .def(
           "_serialize_raw", [](block *term, std::string const &filename,
                                std::string const &sort) {
-            serializeRawTermToFile(filename.c_str(), term, sort.c_str());
+            FILE *file = fopen(filename.c_str(), "a");
+            serializeRawTermToFile(file, term, sort.c_str());
+            fclose(file);
```
Baltoli, line 101: Nitpick but this can be a `unique_ptr`:
```suggestion
            auto file = std::unique_ptr<FILE, decltype(fclose)>(fopen(filename.c_str(), "a"), fclose);
            serializeRawTermToFile(file.get(), term, sort.c_str());
```

Baltoli (approved):
LGTM once conflicts are fixed!

bindings/python/runtime.cpp
dwightguth: I don't know that I really like this idiom.

PR #972, opened 2024-02-06
Update immer dependency

Baltoli:
In https://github.com/runtimeverification/immer/pull/2, I made a tiny change to our immer dependency that will eventually enable us to update to C++20 when the time comes. This PR just bumps the backend submodule to use that new version.


dwightguth (commented):
it looks like your update is changing slightly more than you expected. I'm not entirely sure this is correct. You might want to take a closer look at the exact commit history and the exact commit we were previously using.

Baltoli:
Ah, well spotted - I presumed the submodule was up to date with the master branch of our fork but apparently not.

dwightguth (approved):

Issue #974, opened 2024-02-06
serializeConfiguration does not efficiently compute symbol name and arguments to emit


dwightguth:
Right now in `serializeConfigurationInternal`, it will get the tag of the block being emitted, then convert that tag to a KORE string containing name and arguments using `getSymbolNameForTag`. Then it will run this symbol name through a function which parses the symbol name into name and arguments using the KORE parser, and caches the result. This is pretty inefficient, especially if you are calling serializeConfiguration over and over again.

Since the tags are precomputed statically at compilation time, we ought to be able to precompute a table that is indexed by tag and each table entry contains the exact sequence of bytes we need to emit to serialize the name and arguments of that symbol. This will be much more efficient, at a modest space overhead in the size of the interpreter binary.

PR #978, opened 2024-02-08
Precompute serialized symbols 

Baltoli:
When a symbol has no formal sort parameters, we can directly look up its name in the interpreter's data tables and use a faster path for serialization (instead of parsing every symbol and doing a lookup in a hash table to get the sort arguments and name).

Fixes https://github.com/runtimeverification/llvm-backend/issues/974

### DIFF
--- a/include/runtime/header.h
+++ b/include/runtime/header.h
@@ -354,6 +354,8 @@ char const *getReturnSortForTag(uint32_t tag);
 char const **getArgumentSortsForTag(uint32_t tag);
 char const *topSort(void);
 
+bool symbolIsInstantiation(uint32_t tag);
+
 typedef struct {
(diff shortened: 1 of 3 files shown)


dwightguth (approved):
I see this only changes the behavior for non-parametrized symbols, but such symbols are common enough that since this is primarily being done for performance, it ought to substantially improve performance nonetheless. I am approving. Thanks!

Baltoli:
Yes, that's correct - for the sake of recording it here, it's not possible to do this optimisation for `inj` symbols specifically because they rely on the actual run-time sort being threaded through the serializer. Treating other parameterised symbols the same as injections keeps the code clean while - as you identify - making the common case faster.

PR #979, opened 2024-02-09
Follow rv-tools Nixpkgs version

Baltoli:
As discussed; the version pinned here is actually the version already being used by the backend.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -3,7 +3,9 @@
 
   inputs = {
     utils.url = "github:numtide/flake-utils";
-    nixpkgs.url = "github:nixos/nixpkgs";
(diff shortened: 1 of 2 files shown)


flake.nix
```diff
@@ -3,7 +3,9 @@
 
   inputs = {
     utils.url = "github:numtide/flake-utils";
-    nixpkgs.url = "github:nixos/nixpkgs";
+    rv-utils.url = "github:runtimeverification/rv-nix-tools/f3e8bab938db9800044a94b94e59c5eba25977bc";
```
goodlyrottenapple, line 6: i think its fine to just do
```suggestion
    rv-utils.url = "github:runtimeverification/rv-nix-tools";
```
since we fully control updates to this repo and it will make the update scripts a lot easier (we wont have to sed the hash inside the nix file and can just call `nix flake lock --update-input rv-utils`

PR #982, opened 2024-02-14
Adding a side condition exit event to the proof hint trace

theo25:
The new event carries the result of the side condition check.


Baltoli (approved):
Code looks good; will defer to @nishantjr for a final review

nishantjr (approved):
@theo25  looks good from what I can tell.
I'm finding it a bit difficult to review with the kore terms un-expanded in the pretty output. Would it be possible to expand them in `imp-proof.out`?
I think the python script has some option for this. Its fine to do this in a later PR of course.

Baltoli:
> Its fine to do this in a later PR of course

Yep, probably best to do that in a separate PR - will merge this given your approval @nishantjr.

PR #985, opened 2024-02-15
Serializer optimization

dwightguth:
This PR optimizes the serializer somewhat with the following refactorings:

* Add a flag to the serializer to disable interning.
* Disable interning when generating a proof trace
* Clean up and micro-optimize the process of emitting characters, strings, and integral data to the serializer


include/kllvm/binary/serializer.h
```diff
@@ -148,17 +132,10 @@ class serializer {
...
-  for (auto it = begin; it != end; ++it) {
-    emit(*it);
-  }
-}
-
 template <typename T, typename>
 void serializer::emit(T val) {
-  auto bytes = detail::to_bytes(val);
-  emit(bytes.begin(), bytes.end());
+  buffer_.append((char *)&val, sizeof(T));
```
Baltoli, line 137: I would rather these (and similar) were `reinterpret_cast`s

Baltoli (approved):
One minor nitpick but feel free to merge when that's resolved!

PR #986, opened 2024-02-16
Update to C++-20

Baltoli:
Core support for C++20 is now good enough across the versions of Clang that we support building the backend with that I'm happy to set it as the default version. Doing so will get us some nice refactoring opportunities in the future.

Rebased version of https://github.com/runtimeverification/llvm-backend/pull/947


gtrepta:
Is the `using enum attribute_set::key;` quality of life improvement for #983 able to be introduced in this PR?

Baltoli:
I'll probably do it in a separate cleanup PR; I wanted to keep this one to just the minimum changes we need to enable C++20

gtrepta (approved):

PR #988, opened 2024-02-16
Remove array namespace check

gtrepta:
For https://github.com/runtimeverification/k/issues/3976

There's a special case for a hook namespace in `ARRAY` which falls back to K semantics for those symbols. The frontend wont emit any hook attributes in that namespace for the ARRAY module anyway so it can safely be removed.


Baltoli (approved):
A quick grep suggests that there are no actual usages of the `ARRAY` module in the backend's integration test suite, and so we can safely remove this without updating the tests. Thanks @gtrepta!

PR #992, opened 2024-02-22
Rename `arenaAlloc` to fix collision with GHC RTS

Baltoli:
This fixes a rather unpleasant bug when using the C bindings from the Haskell backend, whereby [recent changes](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=246462) to the dynamic loader on macOS break symbol lookup from dynamic libraries.

The result of this change is that calls to the backend's `arenaAlloc` function were ending up in the Haskell RTS's function of the same name, meaning that we got garbage memory allocated all over the place. There's no sensible fix here other than to rename our function and avoid the collision by doing so.


goodlyrottenapple (approved):
This does indeed fix the issue when loading the llvm backend as a library on macos

PR #996, opened 2024-02-26
Run `clang-tidy` on headers

### LINKED ISSUE #995: Run `clang-tidy` on headers as well as source files
I realised that our current setup doesn't detect `clang-tidy` violations in any header files, only in source files that are compiled by CMake directly. We can fix this with `-header-filter 'include/(kllvm|runtime)'` to pull in only our own header files (and not dependencies, LLVM etc.). Doing so should be pretty mechanical, but might produce a large diff.

Baltoli:
This is a largely mechanical PR that extends our `clang-tidy` setup to run on header files as well as implementations (an oversight in our earlier introduction of the checker).

No new checks are added, so there shouldn't be anything controversial in the changes here.

Fixes #995

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -21,6 +21,7 @@ Checks:
   -bugprone-exception-escape
   -bugprone-implicit-widening-of-multiplication-result
   -bugprone-unchecked-optional-access
+  -bugprone-lambda-function-name
   cppcoreguidelines-*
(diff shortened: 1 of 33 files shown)


Scott-Guest (approved):

Scott-Guest (approved):
Gave it a quick skim, and everything looks reasonable

PR #997, opened 2024-02-27
Remove `zlib` dependency

Baltoli:
When setting CLion up to do some automatic refactoring, it complained about not being able to find this dependency. Some digging reveals it was added in the very first commit of the backend's build system and has not been touched since, so I think it's safe to remove.


dwightguth:
I believe it was added because it was required by something we were doing with llvm in a very early version of the backend. llvm depends on zlib. But it doesn't look like we actually need the development version of the package anymore.

Scott-Guest (approved):

Issue #994, opened 2024-02-23
Decide on and enforce house C++ style


Baltoli:
## Introduction

After https://github.com/runtimeverification/llvm-backend/pull/992 [^1], I think it's worth putting the effort in to properly clean up our C++ code with a house style. We use a bunch of different styles in the C++ code, and there is no strong consensus for what the correct solution will look like. This issue is an attempt at an opinionated set of defaults that we can bikeshed, then start to apply gradually to the code.[^2]

> [!NOTE]
> There are a few things that are out of scope here. We already have a consistent _structural_ style for the code that's enforced conveniently by `clang-format`, and similarly we have a lot of semantic best practices and conventions being enforced by `clang-tidy`. We can therefore restrict the bikeshedding here to naming aesthetics and consistency.

## Migration Strategy

Once we agree on a style, it should be easily achievable to migrate one stylistic element at a time using [`clang-tidy`](https://clang.llvm.org/extra/clang-tidy/checks/readability/identifier-naming.html).

[^1]: We'd have avoided this pain by enforcing a consistent naming prefix for our C code so as to remain hygienic when interfacing with the outside world. Nobody other than us is naming their functions `kllvm_arena_alloc`!
[^2]: We can hopefully clean up or improve some of the bigger, gnarlier functions that are currently exempted from `clang-tidy`'s `congitive-complexity` warnings when we do this.

ehildenb:
All for it! I think it's best to just pick a style, make the PR, and then let people nit at it. But I think general consensus is that a consistent style is more important than any particular choice.

Baltoli:
Indeed - it's blocked on #995 but once that's resolved I can use the tooling to just apply this style globally and see if there are nitpicks

PR #1000, opened 2024-03-01
Enforce style

Baltoli:
This PR is a mechanical[^1] fix for #994; it selects some default naming conventions for C++ code and applies them across the project. Open to bikeshedding if there are particular places anyone doesn't like, but as with all these changes the most important thing is having the tool enabled at all.

The changes might not be perfect; there's a lot of surface area to cover here but they should be pretty good. I've tested against the K integration test suite to make sure nothing is broken there.

[^1]: `clang-tidy` + a pile  of one-off `sed` scripts to cover edge cases; fortunately style is easier to enforce than to apply so `clang-tidy` will do on its own moving forwards.

## Summary by CodeRabbit

- **Refactor**
  - Standardized naming conventions across the codebase for improved consistency and readability. This includes converting camelCase to snake_case and adjusting enum values to PascalCase where applicable.
  - Renamed various entities (functions, variables, classes) to adhere to the new naming convention, enhancing code clarity.
- **Style**
  - Updated comments and documentation to reflect changes in naming conventions and code structure.
- **Chores**
  - Adjusted method signatures and variable names for consistency with the newly adopted naming standards.


include/kllvm/binary/serializer.h
```diff
@@ -28,10 +28,10 @@ class serializer {
...
   enum flags {
     NONE = 0,
-    DROP_HEADER = 1,
-    DROP_ARITY = 2,
-    DROP_BOTH = 3,
-    NO_INTERN = 4,
+    DropHeader = 1,
+    DropArity = 2,
+    DropBoth = 3,
+    NoIntern = 4,
```
Scott-Guest, line 34: Shouldn't this be `lower_case` based on  `readability-identifier-naming.EnumCase`?

include/kllvm/parser/KOREScanner.h
```diff
@@ -29,36 +29,36 @@ enum class token {
   COLONEQUAL,
   ID,
   STRING,
-  TOKEN_EOF,
```
Scott-Guest, line 32: All the other `token` enum values need to be updated to `CamelCase` as well

include/runtime/header.h
```diff
@@ -25,11 +25,11 @@
 #include <runtime/collections/rangemap.h>
 #include <unordered_set>
 
-struct MatchLog {
+struct match_log {
   enum { SUCCESS = 0, FUNCTION, FAIL } kind;
```
Scott-Guest, line 29: Again, shouldn't this be `lower_case` based on `readability-identifier-naming.EnumCase`? Or `CamelCase` based on `readability-identifier-naming.EnumConstantCase`? Not sure which one applies in this case.

Scott-Guest (approved):
LGTM, except a few of the enums need to be fixed up.

The style is a bit different from my personal preference (I like `CamelCase` types), but I'm happy enough with the current selection.

Scott-Guest:
Also oh my the noise from CodeRabbit is pretty bad here :grimacing: Not the biggest fan from what I've seen so far

Baltoli:
> few of the enums need to be fixed up

Will sort these out and merge!

> Also oh my the noise from CodeRabbit is pretty bad here 😬 Not the biggest fan from what I've seen so far

Yeah, not sure I love it really either. Will give it one more go on a more "semantic" PR but I'm inclined to turn it off.

Baltoli:
> I can probably actually just get it to PascalCase types without too much extra effort

In fact, I might do this and `using` / `typedef` in a follow-up change just so that I can get this initial large diff merged; it's becoming a bit unwieldy to work with

include/runtime/header.h
Baltoli: As above, it's `EnumConstantCase` - should be `CamelCased`

PR #1006, opened 2024-03-05
Miscellaneous Scala cleanup

Scott-Guest:
This PR makes a couple quick changes analogous to runtimeverification/k#4055:

- Update `scala-maven-plugin` to its latest version
- Set the Java version by relying on `<release>` in `maven-compiler-plugin` inherited from the parent POM, rather than manually setting `-source` and `-target` for `javac`
- Disable various Scala language extensions and enable `-Werror`
  - One warning about pattern matching on an erased type
  - Remaining warnings were exhaustiveness issues fixed with a `case _ => ???`. This is poor style and an abuse of `???` IMO, but it's already pervasive throughout our codebase.


Baltoli:
Will need the changes to Maven usage from: https://github.com/runtimeverification/k/pull/4108

Baltoli (approved):

Issue #999, opened 2024-02-29
Remove back-edge dependency on the K frontend


Baltoli:
Currently, the pattern-matching compiler depends on Scala code in the K frontend to parse KORE. This induces a backwards edge in the K ecosystem dependency graph that's previously been painful to deal with.[^1] As part of our general work to simplify and reorganise the K ecosystem code, we want to remove this backwards edge in the graph: https://github.com/runtimeverification/k/issues/4063#issuecomment-1971457051

Some thoughts on how we might achieve this:
- Pull the offending code out of the frontend and into the backend directly; the frontend will then need to depend on the backend to get the parser etc.
- Extract the KORE parser Scala code as a standalone library that lives in its own repository and can be consumed as a regular Maven dependency.
- _much more work, longer term_ do a rewrite of the matching compiler to no longer use Scala.

[^1]: It doesn't happen often because the KORE language is pretty stable, but when we last made changes (related to `\left-assoc` and `\right-assoc`) it was _extremely_ painful for a number of reasons: getting PRs merged in lockstep with each other, people's maven caches getting stale and producing confusing errors, etc.

Baltoli:
The packages that are backwards dependencies from the LLVM backend into the frontend:
```
org.kframework.attributes.HasLocation
org.kframework.attributes.Location
org.kframework.attributes.Source
org.kframework.kore.K
org.kframework.kore.KORE.KApply
org.kframework.kore.KORE.KList
org.kframework.mpfr._
org.kframework.parser.kore
org.kframework.parser.kore.CompoundSort
org.kframework.parser.kore.Sort
org.kframework.parser.kore.SymbolOrAlias
org.kframework.parser.kore.Variable
org.kframework.parser.kore._
org.kframework.parser.kore.implementation.{
org.kframework.parser.kore.parser.KoreToK
org.kframework.parser.kore.parser.TextToKore
org.kframework.unparser.ToKast
org.kframework.utils.errorsystem.KEMException
org.kframework.utils.errorsystem.KException```

Baltoli:
I think the first step to take here is to migrate the error-handling code out of the pattern matcher, and have the frontend do translation on the errors and warnings it sees.

PR #1007, opened 2024-03-06
Remove usages of K frontend code

Baltoli:
This PR finalises the removal of the LLVM backend's dependence on the K frontend, now that we have the shared data structures extracted out to the [`scala-kore` repository](https://github.com/runtimeverification/scala-kore). There are a few smaller changes that make up this PR, but the commit history is clean and can be reviewed commit-by-commit.

Once this is merged, the LLVM backend will no longer depend on the K Frontend and we will have eliminated the backwards edge in the K dependency graph!

Fixes: https://github.com/runtimeverification/llvm-backend/issues/999


goodlyrottenapple (approved):

matching/src/main/java/org/kframework/backend/llvm/matching/MatchingException.java
```diff
@@ -1,6 +1,6 @@
 package org.kframework.backend.llvm.matching;
 
-import org.kframework.kore.Pattern;
+import com.runtimeverification.k.kore.Pattern;
```
gtrepta, line 3: Just nitpicking here, what was the reason for going to this prefix instead of back to `org.kframework.parser`?

Baltoli: * `com.runtimeverification.k` better reflects our current organisational structure; the old code was written when this all would have lived under the legacy `kframework` organisation.
* The group ID set in the K POM and here is `com.runtimeverification`
* It's not just a parser; there's a parser subpackage but the data structures don't conceptually belong there
* Easier eye test; will make it much simpler to do the port when we get to the frontend because it's more obvious which version is being user

Scott-Guest (approved):

PR #1008, opened 2024-03-07
Set LLVM backend up to receive automatic updates from scala-kore

Baltoli:
This PR makes two conceptual sets of changes:
- Set up an instance of our standard release-tag dependency update workflow to receive new releases of the `scala-kore` library.
- Remove the backend's inheritance from the K parent POM; I had missed that this backwards dependency still existed, and only noticed when working on this PR that the Nix hash was still changing unexpectedly.[^1]

The actual changes here are small, and I've verified by running the CI action against this branch that the correct behaviour is induced. Once this PR is merged, I will enable the automated dependency updater for `scala-kore`.

[^1]: I've verified by inspecting the `.m2` directory that the fixed-output derivation creates that the only RV dependencies that end up cached are `scala-kore` and `mpfr-java`; prior to this PR we were still getting K `parent` snapshot POMs in the FOD, which break the hash on each K release.


matching/pom.xml
```diff
@@ -3,17 +3,23 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
 
   <modelVersion>4.0.0</modelVersion>
-  <parent>
-    <groupId>com.runtimeverification.k</groupId>
-    <artifactId>parent</artifactId>
-    <version>1.0-SNAPSHOT</version>
-    <relativePath>../../../../../..</relativePath>
```
gtrepta, line 10: Nice to see this relativePath cell go away!

gtrepta (approved):

PR #1009, opened 2024-03-07
Fix `kompile -O3`

dwightguth:
There exists a bug in `kompile -O3` which occurs if a specialized step function has a residual of collection sort (i.e., its rhs has a function that returns a collection) and garbage collection is triggered immediately after this rule applies. In this case, the collection in question has been allocated with koreAllocAlwaysGC, and the memory it is using gets freed. As a result, subsequent allocations can write to that memory and corrupt the term pointed to by that collection pointer.

The fix is to detect when this occurs just prior to invoking the garbage collector, and copying the collection pointer into a block with a block header, which then gets passed as the updated garbage collection root to the gc. GC keeps that memory alive, and then the post-gc address seen by the function that invoked the collector corresponds to the migrated pointer on the main kore heap.

We implement this with the following steps:

1. Create a family of functions `store_<cat>_for_gc` and `load_<cat>_for_gc` which handle the process of storing and loading the collection pointer to the heap when the GC is invoked.
2. Create new reserved symbols used to store these values for all collection sort categories present in the definition.
3. Ensure that the garbage collector knows that a root is a block containing a collection rather than the collection pointer itself.


lib/ast/definition.cpp
```diff
@@ -73,12 +73,30 @@ void kore_definition::add_module(sptr<kore_module> module) {
...
 
+  for (auto const &cat : hooked_sorts_) {
+    switch (cat.first.cat) {
+    case sort_category::Map:
+    case sort_category::List:
+    case sort_category::Set:
+    case sort_category::RangeMap:
+      decl = kore_symbol_declaration::create(
+          "rawCollection_" + std::to_string((int)cat.first.cat), true);
+      sort = cat.second;
```
Baltoli, line 91: I'd rather not redeclare `decl` and declare `sort` at the top here; it took me a moment to parse what was happening in this code. No reason not to just declare a fresh variable for each inside the body of the `case`, I don't think

lib/codegen/Decision.cpp
Baltoli, line 1: This is a lot of extra code being added into this function; could we pull the new code out into its own function?

lib/ast/definition.cpp
```diff
@@ -73,12 +73,30 @@ void kore_definition::add_module(sptr<kore_module> module) {
...
   mod->add_declaration(std::move(decl));
 
+  for (auto const &cat : hooked_sorts_) {
+    switch (cat.first.cat) {
+    case sort_category::Map:
+    case sort_category::List:
+    case sort_category::Set:
+    case sort_category::RangeMap:
+      decl = kore_symbol_declaration::create(
+          "rawCollection_" + std::to_string((int)cat.first.cat), true);
```
Baltoli, line 90: As we're using the same `rawCollection_i` schema elsewhere, let's pull the computation of the symbol name for each category out into a function.

runtime/collect/collect.cpp
```diff
@@ -326,11 +331,108 @@ void kore_collect(void **roots, uint8_t nroots, layoutitem *type_info) {
...
+
+bool store_rangemap_for_gc(void **roots, rangemap *ptr) {
+  if (get_arena_semispace_id_of_object(ptr) != ALWAYSGCSPACE_ID) {
+    *roots = ptr;
+    return false;
+  }
+  void *mem = kore_alloc(sizeof(blockheader) + sizeof(rangemap));
+  auto *hdr = (blockheader *)mem;
+  std::string name = "rawCollection_"
+                     + std::to_string((int)kllvm::sort_category::RangeMap)
```
Baltoli, line 398: See above

runtime/collect/collect.cpp
```diff
@@ -326,11 +331,108 @@ void kore_collect(void **roots, uint8_t nroots, layoutitem *type_info) {
...
+  auto *child = (rangemap *)(hdr + 1);
+  *child = std::move(*ptr);
+  *roots = mem;
+  return true;
+}
+
+map *load_map_for_gc(void **roots, bool is_block) {
+  void *mem = *roots;
+  if (is_block) {
+    return (map *)(((char *)mem) + 8);
```
Baltoli, line 410: 8 here is `sizeof(blockheader)` right?

Baltoli (changes requested):

include/runtime/collections/RBTree.h
```diff
@@ -130,6 +130,15 @@ class RBTree {
...
   }
 
+  RBTree(RBTree const &other) = default;
+
+  RBTree &operator=(RBTree const &other) = default;
+
+  RBTree &operator=(RBTree &&other) {
+    other.root_.swap(root_);
+    return *this;
+  }
```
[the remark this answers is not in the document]
dwightguth: I can add the move constructor, but this cannot be a defaulted operator because the defaulted move semantics for shared_ptr will move the contents of the shared_ptr rather than simply moving the pointer itself. This leads to a segmentation fault.

Baltoli (approved):

PR #1010, opened 2024-03-08
Revert "Set LLVM backend up to receive automatic updates from scala-kore"

dwightguth:
Reverts runtimeverification/llvm-backend#1008


dwightguth:
Not sure why this is failing CI but it's a direct revert of the tip of master so I'm going to go ahead and merge despite that.

Issue #1013, opened 2024-03-12
Run non-Nix CI path


Baltoli:
We have experienced issues recently because the only code path that's currently exercised in the backend CI is the Nix configuration; K itself is built "locally" in its own CI, which means there's a backend configuration that can be broken downstream without us noticing here.

To fix this, we should reinstate a docker-based Ubuntu build & test of the LLVM backend that matches the environment in which K is built.

PR #1014, opened 2024-03-12
Add CI job to build and test the backend outside Nix

Baltoli:
Previously, our CI for the LLVM backend was exclusively using the Nix flake setup to run tests. This meant that there was an edge case failure mode that could manifest in the K dependency update job, where an error that broke local builds of the backend would pass CI here then end up blocking the release job.

This PR addresses the issue by adding an extra CI job to smoke test the local build in a configuration that matches how the K job builds the backend.

Fixes https://github.com/runtimeverification/llvm-backend/issues/1013


.github/workflows/Dockerfile
```diff
@@ -49,7 +50,7 @@ RUN groupadd -g $GROUP_ID $GROUP && useradd -m -u $USER_ID -s /bin/sh -g $GROUP
 USER $USER:$GROUP
 
 RUN    pip3 install --user --upgrade pip  \
-    && ~/.local/bin/pip3 install --user lit
+    && ~/.local/bin/pip3 install --user lit pybind11
 
 ADD matching/pom.xml /home/user/.tmp-maven/
 RUN    cd /home/user/.tmp-maven \
```
gtrepta, line 56: Is this working? It looks like compilation of `matching` is still downloading dependencies on the test action: https://github.com/runtimeverification/llvm-backend/actions/runs/8263962179/job/22606511215#step:5:4858

Baltoli: Turns out this is expected behaviour; `dependency:go-offline` only caches build and test dependencies, not things like plugins.

gtrepta (approved):

PR #1016, opened 2024-03-18
Add source artifacts for Scala compiler bridge

Baltoli:
https://github.com/runtimeverification/llvm-backend/pull/1006 is failing because the Scala Maven plugin is trying to download the sources for the compiler bridge, but can't do so outside of a Nix FOD. This PR adds the bridge to the manually-downloaded Maven artifacts (as we did in https://github.com/runtimeverification/k/pull/4055)


Scott-Guest (approved):
LGTM, thank you!

PR #1019, opened 2024-04-02
Adding `get_requires()` to Python bindings

Robertorosmaninho:
This is a requirement from the Math Proof Generation Team of $Piˆ2$ to get easy access to the Pattern of a requires clause.

This PR must be merged before this [PyK PR](https://github.com/runtimeverification/pyk/pull/1056) that will properly expose this binding to the Pi2 project.

### DIFF
--- a/bindings/python/ast.cpp
+++ b/bindings/python/ast.cpp
@@ -179,7 +179,8 @@ void bind_ast(py::module_ &m) {
           py::arg("is_claim") = false)
       .def_property_readonly("is_claim", &kore_axiom_declaration::is_claim)
       .def("add_pattern", &kore_axiom_declaration::add_pattern)
-      .def_property_readonly("pattern", &kore_axiom_declaration::get_pattern);
+      .def_property_readonly("pattern", &kore_axiom_declaration::get_pattern)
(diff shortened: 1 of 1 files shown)


bindings/python/ast.cpp
```diff
@@ -179,7 +179,8 @@ void bind_ast(py::module_ &m) {
           py::arg("is_claim") = false)
       .def_property_readonly("is_claim", &kore_axiom_declaration::is_claim)
       .def("add_pattern", &kore_axiom_declaration::add_pattern)
-      .def_property_readonly("pattern", &kore_axiom_declaration::get_pattern);
+      .def_property_readonly("pattern", &kore_axiom_declaration::get_pattern)
+      .def("get_requires", &kore_axiom_declaration::get_requires);
```
Baltoli, line 183: This should be a read only property, and shouldn't have the get_ prefix

Baltoli (changes requested):

bindings/python/ast.cpp
Robertorosmaninho: Thanks, Bruce! I've just modified it as required!

Baltoli (approved):

PR #1020, opened 2024-04-03
Updates to hook and function event hints

theo25:
This PR implements two updates to the format of hook and function event hints.

1. The PR fixes an inconsistency in the hint format for hook hints, where the hooks that are hard-coded in the code generator were including their arguments as part of the hook hint, while the hooks that call an external function were not including their arguments as part of the hint. We know include the hook arguments in both cases.
2. The PR  removes the arguments from function events by default. This is done for performance reasons and given that the math proof team does not need these arguments for their proof automation. We add an additional flag to `llvm-kompile`, namely `--proof-hint-instrumentation-slow`, that if passed (instead of `--proof-hint-instrumentation`) will instrument the code in the previous way, i.e. with generation of arguments for function event hints.


lib/codegen/ProofEvent.cpp
```diff
@@ -174,7 +174,7 @@ proof_event::event_prelude(
 llvm::BasicBlock *proof_event::hook_event_pre(
     std::string const &name, llvm::BasicBlock *current_block,
     std::string const &location_stack) {
-  if (!proof_hint_instrumentation) {
```
dwightguth, line 177: I think it would be better if setting `--slow` also set the regular flag rather than having to make the code check both flags everywhere.

dwightguth (changes requested):

lib/codegen/ProofEvent.cpp
theo25: Good point, I pushed a commit to do that.

dwightguth (approved):

PR #1021, opened 2024-04-04
Don't emit non-exhaustiveness warning if `no-evaluators` is present

Scott-Guest:
Closes runtimeverification/k#4148

Updates `matching` to not emit a warning if the `no-evaluators` attribute is present.


tothtamas28 (approved):
Consider writing a test for this.

Scott-Guest:
> Consider writing a test for this.

The other exhaustiveness tests are in the frontend, so added one there: runtimeverification/k#4160

PR #1025, opened 2024-04-12
Omitting intemedaite configuration events in the proof trace

theo25:
This PR disables the generation of intermediate configuration events in the proof trace by default.  It is a followup of #1020 and aims to further improve the performance of generating hints. Similarly to #1020, we allow the user to get the old behavior of emitting every intermediate configuration event by passing the `--proof-hint-instrumentation-slow`  flag to `llvm-kompile`.


runtime/util/finish_rewriting.cpp
```diff
@@ -44,6 +45,10 @@ int32_t get_exit_code(block *);
     } else {
       print_configuration(output_file, subject);
     }
+  } else if (!error && !proof_hint_instrumentation_slow) {
+    write_uint64_to_file(output_file, 0xFFFFFFFFFFFFFFFF);
+    serialize_configuration_to_file(output_file, subject, true, false);
+    write_uint64_to_file(output_file, 0xCCCCCCCCCCCCCCCC);
   }
```
Robertorosmaninho, line 52: I just didn't understand this part. Can you elaborate on this?

lib/codegen/ProofEvent.cpp
```diff
@@ -288,7 +288,7 @@ llvm::BasicBlock *proof_event::rewrite_event_pre(
 llvm::BasicBlock *proof_event::rewrite_event_post(
     kore_axiom_declaration *axiom, llvm::Value *return_value,
     llvm::BasicBlock *current_block) {
-  if (!proof_hint_instrumentation) {
+  if (!proof_hint_instrumentation_slow) {
```
Robertorosmaninho, line 291: Is this the only place we need to modify this flag?

Robertorosmaninho (approved):
A few questions just for my understanding, otherwise LGTM!

runtime/util/finish_rewriting.cpp
gtrepta: I think what's happening here is that if `proof_hint_instrumentation_slow` is set, then the final configuration gets emitted by the final rewrite event. Without it, though, we still want to emit the final configuration, so we do that here. Is that correct, @theo25 ?

theo25: yes that is correct. When the slow flag has been passed we don't need to emit the final config here because it would have been emitted the same way that the rest of the intermediate configs would have been emitted. But when the slow flag has not been passed, we need to emit it here.

lib/codegen/ProofEvent.cpp
theo25: yes we want to emit the intermediate configs if the slow flag has been passed. If the normal flag has been passed or if no flag has been passed (in both of those cases (`proof_hint_instrumentation_slow` would be `false`) we want to skip this part.

gtrepta (approved):
LGTM

PR #1026, opened 2024-04-15
Store base pointer offset for map/list/set children of blocks and use this to migrate during garbage collection with -O3

dwightguth:
Previously there existed a bug that would occur if the following happened:

1. A garbage collection is triggered with at least two roots:
   a. A map/list/set which is a child of a symbol-layout block
   b. A symbol-layout block that recursively points to the symbol-layout block which is the parent of the map/list/set
2. Garbage collection does not copy the symbol which is the parent of the map/list/set when tackling root 1a above. It merely tackles the internal memory allocated by that map/list/set.
3. Garbage collection moves the block to the new allocation semispace. The root pointer to the map/list/set still points to the collection semispace.
4. Garbage collection ends, leaking a pointer into the collection semispace. This memory has been corrupted by storing forwarding pointers to the allocation semispace.
5. The backend tries to read the map/list/set, but its length field is corrupted by the forwarding pointer, leading to a segfault.

In order to fix this, we annotate each map/list/set child of a block with a "base pointer offset". This offset is an integer that, when added to the address of the map/list/set pointer which is 8 bytes following it, will return the address of the block header of the containing block. We do not try to modify `get_value_type` since that would be very intrusive as a change; instead we only modify the logic which selects children of blocks using llvm `getelementptr` instructions, as well as the code that creates map/list/set children in blocks in the first place.

Once these annotations have been applied, which is the subject of the first 5 commits in this PR, we modify the garbage collector to use them in order to reconstruct the base pointer when it receives a garbage collection root. We refactor out a `migrate_root` function which handles GC roots, and rename `migrate_roots` to `migrate_static_roots` to avoid confusion. Then we modify `migrate_root` to reconstruct the base pointer, migrate the base pointer, and then write the updated derived pointer back to the array of gc roots.

We no longer need to call `migrate_map`, `migrate_set`, or `migrate_list` on the resulting map/list/set because this will now occur for us during evacuation of the relevant parent block.

Code to handle the case where a map/list/set pointer is bare on the alwaysgcspace rather than a child of a block remains largely unchanged, except that we need to modify how we convert those collection to the kore heap in order to accommodate the new memory layout.

Finally, we add a test. I ran the test on the master branch of K and it failed.


dwightguth:
This is ready for review, but don't merge it yet. I'm still investigating why I seem to be getting some new test failures in KEVM. I want to make sure I haven't introduced a regression.

dwightguth:
Should be ready to merge now once it's been reviewed. I fixed all of the bugs I was able to find.

Baltoli (approved):
One minor comment but this looks well thought through!

PR #1028, opened 2024-04-22
Serialize header for new serialization format

dwightguth:
This PR adds a tool to serialize the header for the new serialization format that will be used by the proof hints. The new format focuses on fast serialization of llvm backend terms rather than generality. It should be quite fast to deserialize as well, however.

Not ready for review yet. I still have to:

* [x] write tests to ensure that the header emitted matches expectations
* [x] double check that this code doesn't break anything
* [x] write up documentation of the new serialization format


tools/kore-rich-header/main.cpp
```diff
@@ -0,0 +1,80 @@
...
+#include <cstdint>
+#include <iostream>
+
+using namespace kllvm;
+using namespace kllvm::parser;
+
+int main(int argc, char **argv) {
+  char *filename = argv[1];
+  kore_parser parser(filename);
+  ptr<kore_definition> definition = parser.definition();
```
Baltoli, line 13: Feel free not to given you're the only consumer of this tool @dwightguth, but consider setting up an LLVM command line argument parser:
- Keeps the signature of the tool documented with `-h` flags / usage message
- Makes it easier to add extra features / behaviour in the future if you need to

Baltoli (approved):
Organizationally, I'd prefer for the actual implementation of the header serialization to be a library function that dumps bytes into a buffer so that the actual command line tool becomes a thin wrapper on top of it.

Doing so would also allow you to bake this tool's behaviour into the code generator more easily (i.e. creating a serialization header for a definition can then be done with a single flag +  function call during code generation).

Approving as that will be a pretty minor change to make, and I'm very happy with the documentation / general structure of these changes.

PR #1029, opened 2024-04-22
Don't emit branch unnecessary check for injection 

Robertorosmaninho:
Fix https://github.com/Pi-Squared-Inc/pi2/issues/1323
The llvm backend currently emits a branch each time a new injection is created to check whether the term that is its child is itself an injection. We skip over some circumstances where such a check would not be necessary, but not all. 
If the sort being injected has no subsort, then the child cannot be an injection. This PR adds a condition to check if the current sort has no subsorts. If that's the case, we also skip the branch/new blocks emit.


dwightguth (changes requested):
You found the right place to modify the code, but the boolean condition is wrong. Right now this is testing whether the sort is parametric, not whether it has any subsorts. We may already have code to test subsorts in kore_definition. I'm unsure. But this is definitely not going to do it. I'd talk with Theo to clarify exactly what the boolean condition should look like.

theo25:
I think you need to utilize this method

https://github.com/runtimeverification/llvm-backend/blob/7e12c1d0dfb2853ee51dee5775919b8440709ee6/include/kllvm/ast/AST.h#L966

in order to check if the sort in question has any subsorts.

dwightguth (changes requested):
Still several problems here.

1. get_subsorts does not cache its result, and it's expensive to call. You need to change this so it reuses the computation on subsequent function calls.
2. You are checking whether any subsorts exist in the definition, not whether a given sort has subsorts.
3. according to the documentation, get_subsorts returns a map from each sort to the set of sorts greater than it. You actually want the inverse map: the map from each sort to the set of sorts less than it.
4. You need to confirm whether or not get_subsorts returns the reflexive or non-reflexive relation. Ie, does it map each sort to itself? If it does not, it should suffice to check emptiness of the set associated with that sort in the inverse relation. If it does, however, you instead need to check whether the size of the associated set is greater than 1.

I'd recommend doing the following:

1. Change get_subsorts to cache its result.
2. Create another function that uses the cached result of get_subsorts to create a cached inverse relation.
3. Call the inverse function you created, then look up the sort in the resulting map.

Baltoli:
https://github.com/runtimeverification/llvm-backend/blob/7e12c1d0dfb2853ee51dee5775919b8440709ee6/lib/ast/definition.cpp#L107-L122

The implementation of `get_subsorts` is not reflexive.

lib/ast/definition.cpp
```diff
@@ -118,7 +123,26 @@ SubsortMap kore_definition::get_subsorts() const {
...
+
+  auto inverted_subsorts = SubsortMap{};
+
+  for (auto const &[inner, outers] : get_subsorts()) {
+    for (auto *outer : outers) {
+      inverted_subsorts[outer].insert(inner);
+    }
+  }
+
+  subsorts_inverted_ = new SubsortMap(inverted_subsorts);
```
Robertorosmaninho, line 144: @theo25, I'm not sure if we need `transitive_closure` here. Could you please advise?

include/kllvm/ast/AST.h
```diff
@@ -914,6 +914,9 @@ class kore_definition {
 
   kore_symbol *inj_symbol_{};
 
+  SubsortMap *subsorts_ = nullptr;
+  SubsortMap *subsorts_inverted_ = nullptr;
```
Baltoli, line 918: Storing raw pointers here is error-prone; as currently written the class will leak the memory allocated to store these maps when it's destroyed. `std::optional` gives you the same interface but correct RAII behaviour.

lib/ast/definition.cpp
Baltoli: The `get_subsorts()` method has entries:
```
S |-> {T . S is a subsort of T}
```
That is, it tells you _every_ sort that's greater than `S`, including transitively.

I think in your specific case you only need the non-transitive entries:
```
S |-> {T . S is a direct supersort of T}
```
such that `S` having a non-empty entry tells us that it has at least one subsort.

For consistency, I'd probably prefer to make your implementation transitive as well, but if you choose not to do this please make sure to document its contract similarly to `get_subsorts()`.

lib/ast/definition.cpp
```diff
@@ -118,7 +123,26 @@ SubsortMap kore_definition::get_subsorts() const {
     }
   }
 
-  return transitive_closure(subsorts);
+  subsorts_ = new SubsortMap(transitive_closure(subsorts));
+  return *subsorts_;
+}
+
+SubsortMap kore_definition::get_inverted_subsorts() {
```
Baltoli, line 130: `get_supersorts`?

Baltoli (changes requested):

lib/ast/definition.cpp
Robertorosmaninho: Do you think that's a better name?

lib/ast/definition.cpp
Robertorosmaninho: I see, and it makes sense! The problem here is that the version I have cached contains the transitive case. I'm unsure if we gain a lot caching only this version as the `transitive_closure` function seems pretty expensive to me.

Robertorosmaninho: Should we cache both versions or just keep the non-transitive version, @Baltoli?

include/kllvm/ast/AST.h
```diff
@@ -914,6 +914,9 @@ class kore_definition {
 
   kore_symbol *inj_symbol_{};
 
+  std::optional<SubsortMap> subsorts_;
+  std::optional<SubsortMap> subsorts_inverted_;
```
Baltoli, line 918:
```suggestion
  std::optional<SubsortMap> supersorts_;
```

include/kllvm/ast/AST.h
```diff
@@ -963,7 +966,17 @@ class kore_definition {
...
+
+  /*
+   * Build this definition's inverted subsort relation from axioms that have the
+   * `subsort` attribute.
+   *
+   * The returned map is as follows:
+   *
+   * S |-> {T . S is a direct supersort of T}
+   */
+  [[nodiscard]] SubsortMap get_supersort();
```
Baltoli, line 979:
```suggestion
  [[nodiscard]] SubsortMap get_supersorts();
```

include/kllvm/ast/AST.h
```diff
@@ -963,7 +966,17 @@ class kore_definition {
    *
    *   S |-> {T . S is a subsort of T}
    */
-  [[nodiscard]] SubsortMap get_subsorts() const;
+  [[nodiscard]] SubsortMap get_subsorts();
+
+  /*
+   * Build this definition's inverted subsort relation from axioms that have the
```
Baltoli, line 972: My point here was that "supersort" captures precisely what you mean by "inverted subsort"; the sub- and supersort relations are indeed inverses of one another, but it's good to keep the naming consistent.

include/kllvm/ast/AST.h
```diff
@@ -963,7 +966,17 @@ class kore_definition {
...
+   * `subsort` attribute.
+   *
+   * The returned map is as follows:
+   *
+   * S |-> {T . S is a direct supersort of T}
```
Baltoli, line 977: This isn't the case in your implementation; my preference is to just cache the transitive closure result for consistency.

lib/ast/definition.cpp
```diff
@@ -104,7 +104,12 @@ void kore_definition::insert_reserved_symbols() {
   add_module(std::move(mod));
 }
 
-SubsortMap kore_definition::get_subsorts() const {
+SubsortMap kore_definition::get_subsorts() {
```
Baltoli, line 107: I don't follow the logic in this function; all it needs to do is compute the transitive closure _once_ and cache that result. Why are you recomputing it?

lib/ast/definition.cpp
```diff
@@ -118,9 +123,29 @@ SubsortMap kore_definition::get_subsorts() const {
...
+  auto inverted_subsorts = SubsortMap{};
+
+  for (auto const &[inner, outers] : get_subsorts()) {
+    for (auto *outer : outers) {
+      inverted_subsorts[outer].insert(inner);
+    }
+  }
+
+  subsorts_inverted_ = std::optional<SubsortMap>(
+      SubsortMap(transitive_closure(inverted_subsorts)));
```
Baltoli, line 145: See related suggestions for renaming this; you also don't need this pile of constructors to assign to a `std::optional`.
```suggestion
  subsorts_inverted_ = transitive_closure(inverted_subsorts);
```

Baltoli (changes requested):

lib/ast/definition.cpp
Robertorosmaninho: The problem here is that we need the non-transitive subsorts set to compute the supersorts. However, on second thought, we could just calculate it in the same way and cache everything right!

lib/ast/definition.cpp
```diff
@@ -118,7 +123,31 @@ SubsortMap kore_definition::get_subsorts() const {
...
+      auto const &inner_sort
+          = att->get_constructor()->get_formal_arguments()[0];
+      auto const &outer_sort
+          = att->get_constructor()->get_formal_arguments()[1];
+      supersorts[outer_sort.get()].insert(inner_sort.get());
+    }
+  }
+
+  supersorts_ = transitive_closure(supersorts);
+  return *supersorts_;
```
Baltoli, line 150: I realise you can actually simplify this a bit more, in fact - down to something like:
```
if (empty) {
  compute sub / supers;
  transitive closure;
}

return *subsorts_;
```

Baltoli (approved):
One more simplification I can see, and the comment re. direct supersorts still needs to be changed, but go ahead and merge once those are addressed.

lib/ast/definition.cpp
Robertorosmaninho: To be honest, I would like to do something like:
```
get_subsorts() {
  if (empty) {
    compute sub && compute super
    sub_ = transitive closure sub
    super_ transitive closure super
  }
  
  return *sub_;
}
```

Baltoli: I'm not sure that it makes sense to precompute both at the same time; there are a few existing call sites that use the subsorts relation but don't need the supersorts one.

lib/codegen/CreateTerm.cpp
```diff
@@ -952,12 +952,11 @@ std::pair<llvm::Value *, bool> create_term::create_allocation(
...
-    if (auto cat
-        = dynamic_cast<kore_composite_sort *>(symbol->get_arguments()[0].get())
-              ->get_category(definition_)
-              .cat;
+    if (auto *sort
+        = dynamic_cast<kore_composite_sort *>(symbol->get_arguments()[0].get());
         symbol_decl->attributes().contains(attribute_set::key::SortInjection)
-        && (cat == sort_category::Symbol)) {
+        && (sort->get_category(definition_).cat == sort_category::Symbol)
+        && definition_->get_supersorts()[sort].empty()) {
```
dwightguth, line 959: This doesn't seem quite right to me. I'm honestly a little surprised that the tests are passing. If I understand correctly, `get_supersorts` is returning a map from each supersort to the set of subsorts associated with that sort. In such a case, the condition in this case ought to be negated, right? The if block should succeed only if the sorts are non-empty.

Am I misunderstanding the behavior of the code somehow? And if not, can you add some kind of test that would detect this bug?

Robertorosmaninho: I think that what you're saying makes sense to me, Dwight. However, 10% of our tests (using lit) fail if I use the exact opposite condition. I'm blocked from understanding how this ought to work, then. For comparison, as it is implemented, it also passes all K regression tests, but the opposite condition also fails in some tests...

Baltoli (changes requested):
You will need to rebase this once your other PR is merged:
- Remove the versioning commits added while we were using the old versioning scheme.
- Remove the implementation of the supersorts relation as it will be added by the other PR.

PR #1031, opened 2024-04-24
Fix locations in proof hints

theo25:
This PR fixes 2 issues with the generation of locations for the hook and function proof hint events.

- The starting location stack is now the empty stack as opposed to the `"0"` stack. This is because the math team assumes that the first location in the stack corresponds to an argument of the top cell in the configuration and not the top cell constructor.
- The previous method for skipping locations that correspond to injections, namely skipping a location when the child of a term is an injection, was leading to errors when it combined with the injection compaction optimization. Specifically, the method could lead to skipping an extra location in the case where the algorithm needed to generate code for possibly applying an injection compaction. We change the algorithm for skipping locations, so that we skip a location for every child of an injection term.

### DIFF
--- a/lib/codegen/CreateTerm.cpp
+++ b/lib/codegen/CreateTerm.cpp
@@ -324,9 +324,6 @@ llvm::Value *create_term::alloc_arg(
     std::string const &location_stack) {
   kore_pattern *p = pattern->get_arguments()[idx].get();
   std::string new_location = fmt::format("{}:{}", location_stack, idx);
-  if (is_injection_symbol(p, definition_->get_inj_symbol())) {
-    new_location = location_stack;
-  }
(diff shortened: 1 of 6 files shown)


lib/codegen/CreateTerm.cpp
```diff
@@ -323,7 +323,9 @@ llvm::Value *create_term::alloc_arg(
     kore_composite_pattern *pattern, int idx, bool is_hook_arg,
     std::string const &location_stack) {
   kore_pattern *p = pattern->get_arguments()[idx].get();
-  std::string new_location = fmt::format("{}:{}", location_stack, idx);
+  std::string new_location = location_stack.size()
```
Baltoli, line 326: I have a weak preference for using `.empty()` and reversing the branches here

Baltoli (approved):

PR #1032, opened 2024-04-24
Updating `imp.md`, kore tests and `kore-proof-trace` to parse side condition events

Robertorosmaninho:
This PR fixes https://github.com/Pi-Squared-Inc/pi2/issues/1348. This issue reports that `kore-proof-trace` doesn't generate any hints when a `requires` clause is used in an example.

After investigating the issue, we realized that our regression tests didn't cover this case. Therefore, we updated the current `imp-*.kore` tests to contain it. We also updated `imp.md` to be compatible with K v7.0.15. To improve the reproducibility of our tests in the future, we made `imp.md` the source code of `imp.kore`,  `imp-proof.kore`, and `imp-slow-proof.kore`.

Then, we needed to update the Proof Hints version and the `kore-proof-trace` tool to include "side condition" events as arguments and parse them appropriately to hints for the `requires` clause and the rest of the test.


theo25 (commented):
Make sure to update the `proof-trace.md` document to reflect this change.

Robertorosmaninho:
> Make sure to update the `proof-trace.md` document to reflect this change.

I am not sure if this will be needed; the trace should be the same; the significant change is that we can now parse it with `kore-proof-trace.` I would like to check `BOOL.and`, which is not present on the hints right now.

theo25:
The current document does not list the side condition events as possible arguments in the grammar. I think it would be good to fix that as part of this PR.

theo25:
Also, why is there a need to update imp.md and the kore tests as part of this PR? Unless the update is required for the other changes in the PR, I would make another PR for that. If there is indeed a need to update them here, could you explain in some detail in the PR description?

Robertorosmaninho:
> Also, why is there a need to update imp.md and the kore tests as part of this PR? Unless the update is required for the other changes in the PR, I would make another PR for that. If there is indeed a need to update them here, could you explain in some detail in the PR description?

We need at least the update `imp-proof.kore` to contain the `requires` clause. But I can remove the others and address this in a following PR.

Baltoli:
> To improve the reproducibility of our tests in the future, we made imp.md the source code of imp.kore, imp-proof.kore, and imp-slow-proof.kore.

I'm fine with this being done inline with this PR.

Baltoli (commented):
Seems reasonable to me; will leave final approval to @theo25 though!

theo25 (approved):
It looks good to me too! I am approving it but it seems you will need to regenerate the expected outputs for the proof related tests because another PR merged in the meantime that changed them.

PR #1039, opened 2024-05-02
Open run tests for develop PRs

Baltoli:


Baltoli:
The base branch here needs to be `master` so that all the CI bits exist from Github's perspective; we can then branch off + protect `develop` from the commit that merges this PR into `master`.

dwightguth (approved):

PR #1040, opened 2024-05-03
Replace `LLVMREwriteEvent` with `llvm_rewrite_event`

Robertorosmaninho:
A simple PR to modify the CamelCase in `AST.cpp` to snake_case!


Baltoli:
This change _is_ consistent with what's in https://github.com/runtimeverification/k/blob/master/pyk/src/pyk/kllvm/hints/prooftrace.py, but these type names should actually _all_ be camel case in idiomatic Python.

Looking at the code, the issue with converting them is that all the Pybind types are being used as members of Python classes, which have the equivalent camel-cased names. Those Python wrapper classes just directly pass their methods through to the underlying bindings classes. Why do we implement it that way? Is there a good reason we couldn't just do what we do [here](https://github.com/runtimeverification/k/blob/master/pyk/src/pyk/kllvm/ast.py) and just import / re-export the bindings from the module?

cc @theo25

Baltoli:
(FWIW this is on me for not catching this in https://github.com/runtimeverification/k/pull/4288 - just didn't twig that we were doing that wrapping of the bindings until I took a second look at it!)

Baltoli (approved):
Approving to remove the annoying typo; we can revisit the design of the bindings later.

PR #1046, opened 2024-05-07
Improve test harness and test coverage for proof hints

dwightguth:
Previously the tests for proof hints were rather anemic, not to mention somewhat non-standard. We make the following changes in order to canonify a technique for testing proof hints, and also expand the test coverage of this code path:

1. We remove the kore-proof-hint-test binary which appears to be written with the intent of being used in one specific integration test. This seems to be duplicated logic with the test_proof_hint.py python tests, and I don't see the need to maintain a specific C++ binary just for one test. We also remove the test associated with this binary since it appears to be redundant.
2. We clean up the infrastructure for how we test proof hints. We no longer test proof hints without passing --expand-terms, since this exposes information that is not particularly meaningful and makes the tests flaky, like the number of bytes in each kore term in the trace, which may be subject to change.
3. We codify a set of techniques for testing proof hints. Any test which uses `%interpreter` and `%check-diff` can be extended by adding the lines `%proof-interpreter` and `%check-proof-out` to the exact same kore definition. This allows us to reuse kore files and input files and test the same tests both with and without the proof hint.
4. We add the proof hint testing harness to a small selection of more complex tests in the tests directory. If regressions occur in the future, we can extend the number of tests we do this with, and we can selectively choose to enable this test coverage on new integration tests that we create very easily. We do not enable this on all tests because it does not scale well to long execution traces due to the fact that this would require very large expected output files to be stored in source control.


bin/llvm-kompile-clang
```diff
@@ -256,7 +256,7 @@ if $link; then
       mkdir -p "$python_path"
 
       script_name="$(basename "$output_file").py"
-      cp "$LIBDIR/lldb/klldb.py" "$python_path/$script_name"
+      cp -f "$LIBDIR/lldb/klldb.py" "$python_path/$script_name"
```
gtrepta, line 259: What did you need to make this change for?

gtrepta (approved):
LGTM.

bin/llvm-kompile-clang
Baltoli: TL;DR - we're taking a shortcut in our use of Nix when running the integration tests and this is a workaround for that shortcut to work.

Longer version: `klldb.py` is installed by the Nix derivation that builds the LLVM backend; it has a path like `/nix/store/llvm-backend/.../lldb/klldb.py` and is _read-only_ at that path because it's an output of a build (if you could write to it, you could break Nix's guarantees about reproducibility and hashing).

When building a Nix derivation, you need to explicitly declare what its sources are; the Nix builder then takes care of setting permissions on those source inputs so that your build process can modify them inside a sandboxed build environment. However, when we copy this file we aren't copying it from our own sources, but rather straight out of the Nix store. It therefore retains its read-only permissions, and we need to force the copy to overwrite it.

Baltoli (approved):

PR #1047, opened 2024-05-08
Fix `llvm-kompile pythonast`

Baltoli:
This PR addresses an issue experienced by @gtrepta and @theo25 where the Python AST bindings for the LLVM backend incorrectly reference runtime symbols like `FIRST_INJ_TAG` when loaded.

The solution is pretty straightforward; rather than trying to rely on the linker implicitly stripping out undefined symbols when the library is built, we just split the components into two parts (runtime and non-runtime) and selectively include them.

The test case added in https://github.com/runtimeverification/llvm-backend/pull/1047/commits/0eb1d7f63b360bd7e07a036a803e073ab0babcef fails before the script is amended.


gtrepta (approved):
Have you tested the pyk integration tests with this change?

Baltoli:
Yep, they're all good.

PR #1048, opened 2024-05-08
Remove injections from output files of proof hint tests

dwightguth:
Since the Pi2 team that consumes the proof hint is discarding the injections, and we have a minor issue in our upcoming serialization PR that does not quite deserialize injections correctly, I figured the expedient thing to do would be to test whether or not the updated serialization format provides the same output modulo the injections in the provided terms. We implement this by simply recursively visiting each term as it is printed by `kore-proof-trace` and stripping out the injections prior to printing it.

This will be used to show that the deserialized output does not change modulo the injections in a follow-up PR.


lib/ast/AST.cpp
```diff
@@ -1122,6 +1122,19 @@ sptr<kore_pattern> kore_composite_pattern::unflatten_and_or() {
   return result;
 }
 
+sptr<kore_pattern> kore_composite_pattern::strip_injections() {
+  if (constructor_->get_name() == "inj" && arguments_.size() == 1) {
+    return arguments_[0]->strip_injections();
+  }
```
Robertorosmaninho, line 1128: Shouldn't we still check if the `constructor_` has the `inj`  name before constructing the result object?

Baltoli: This implementation is recursive; if we go past this `if` block then we know all injections have been stripped.

Baltoli (approved):

PR #1050, opened 2024-05-08
[UnitTests] Adding `subsortMap` compiler unit tests

Robertorosmaninho:
This PR is part of investigating the `get_subsorts` function to ensure it's working as expected. 

This PR adds two small tests, the first just checking insertion, contains, emptiness, and size of `subsortMap`.  The second test builds a `kore_defintion` as expected by `get_subsorts`, then calls the function and checks if the `subsort2` is part of the `subsortSet` of `subsort1`.

### DIFF
--- a/unittests/compiler/subsortmap.cpp
+++ b/unittests/compiler/subsortmap.cpp
@@ -1,0 +1,50 @@
+#include <boost/test/unit_test.hpp>
+#include <kllvm/ast/AST.h>
+
+using namespace kllvm;
+
+BOOST_AUTO_TEST_SUITE(SubsortMapTest)
+
+BOOST_AUTO_TEST_CASE(insert_contains) {
+  SortSet sortSet;
+  SymbolSet symbolSet;
+  auto subsortMap = SubsortMap{};
+
+  auto sort = kore_sort_variable::create("Sort");
+  auto subsort = kore_sort_variable::create("Subsort 1");
+  auto subsort2 = kore_sort_variable::create("Subsort 2");
+
+  subsortMap[sort.get()].insert(subsort.get());
+  subsortMap[sort.get()].insert(subsort2.get());
+
+
+  BOOST_CHECK(subsortMap[sort.get()].contains(subsort.get()));
+  BOOST_CHECK(subsortMap[sort.get()].contains(subsort2.get()));
+  BOOST_CHECK(subsortMap[sort.get()].size() == 2);
+  BOOST_CHECK(subsortMap.size() == 1);
+  BOOST_CHECK(!subsortMap[sort.get()].contains(sort.get()));
+  BOOST_CHECK(!subsortMap[sort.get()].empty());
+}
+
+BOOST_AUTO_TEST_CASE(subsorts_supersorts) {
+  auto subsort1 = kore_sort_variable::create("bar");
+  auto subsort2 = kore_sort_variable::create("baz");
+
+  auto pat = sptr<kore_composite_pattern>(kore_composite_pattern::create("subsort"));
+  pat->get_constructor()->add_formal_argument(subsort1);
+  pat->get_constructor()->add_formal_argument(subsort2);
+
+  sptr<kore_axiom_declaration> decl = kore_axiom_declaration::create();
+  decl->attributes().add(pat);
+
+  auto mod = sptr<kore_module>(kore_module::create("FooModule"));
+  mod->add_declaration(decl);
+
+  auto def = kore_definition::create();
+  def->add_module(mod);
+
+  auto subsorts_ = def->get_subsorts();
+  BOOST_CHECK(subsorts_[subsort1.get()].contains(subsort2.get()));
+}
+
+BOOST_AUTO_TEST_SUITE_END()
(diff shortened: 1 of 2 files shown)


unittests/compiler/subsortmap.cpp
```diff
@@ -0,0 +1,50 @@
...
+  subsortMap[sort.get()].insert(subsort2.get());
+
+
+  BOOST_CHECK(subsortMap[sort.get()].contains(subsort.get()));
+  BOOST_CHECK(subsortMap[sort.get()].contains(subsort2.get()));
+  BOOST_CHECK(subsortMap[sort.get()].size() == 2);
+  BOOST_CHECK(subsortMap.size() == 1);
+  BOOST_CHECK(!subsortMap[sort.get()].contains(sort.get()));
+  BOOST_CHECK(!subsortMap[sort.get()].empty());
+}
```
Baltoli, line 27: I'm not sure what this test is really for; it boils down to checking that `std::unordered_map` works as documented. It's not wrong per se, but I don't think that the test convinces us of anything we didn't already understand.

unittests/compiler/subsortmap.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+BOOST_AUTO_TEST_CASE(subsorts_supersorts) {
+  auto subsort1 = kore_sort_variable::create("bar");
+  auto subsort2 = kore_sort_variable::create("baz");
```
Baltoli, line 31: It would be good to check that this works for concrete sorts; the underlying mechanism is the same and so it should work similarly, but asserting over concrete sorts will be closer to what's actually happening in the compiler when we deal with real KORE.

unittests/compiler/subsortmap.cpp
```diff
@@ -0,0 +1,50 @@
...
+  decl->attributes().add(pat);
+
+  auto mod = sptr<kore_module>(kore_module::create("FooModule"));
+  mod->add_declaration(decl);
+
+  auto def = kore_definition::create();
+  def->add_module(mod);
+
+  auto subsorts_ = def->get_subsorts();
+  BOOST_CHECK(subsorts_[subsort1.get()].contains(subsort2.get()));
```
Baltoli, line 47: Consider adding some more test cases; this currently only asserts that `bar <: baz`. It would be good to check:
* Transitive cases for larger subsorting partial orders.
* Negative cases; we want to know that `not (baz <: bar)`, `not (bar <: bar)` etc. as well.

Baltoli (changes requested):
It would be good to cherry-pick in your implementation of the inverse "supersort" relation from your other PR, and test it similarly to the existing code here. That way any bugs or differences will be clear before you return to working on the code generator.

unittests/compiler/subsortmap.cpp
Robertorosmaninho: It was meant to test the hashing stuff we suspected to be an error. The idea was to have a very straightforward test indeed. But I can remove it if we decide it doesn't aggregate much as a regression test.

unittests/compiler/subsortmap.cpp
Robertorosmaninho: How can I do that? Could you point me to the right function, please?

unittests/compiler/subsortmap.cpp
Robertorosmaninho: Great idea! Thanks!

unittests/compiler/subsortmap.cpp
Baltoli: The hashing problem we discussed is that if you run:
```
auto s_1 = kore_sort_variable::create("X");
auto s_2 = kore_sort_variable::create("X");
```
then `s_1.get()` and `s_2.get()` are different pointers, so will map to different things.

The solution if this becomes an issue (we don't currently implement this, not sure why) is to have a symbol table mapping sort names to canonical instances.

unittests/compiler/subsortmap.cpp
Baltoli: Just create a `kore_composite_sort` rather than a `kore_sort_variable`

include/kllvm/ast/AST.h
```diff
@@ -963,7 +963,17 @@ class kore_definition {
    *
    *   S |-> {T . S is a subsort of T}
    */
-  [[nodiscard]] SubsortMap get_subsorts() const;
+  SubsortMap get_subsorts();
+
+  /*
+   * Build this definition's inverted subsort relation from axioms that have the
```
Baltoli, line 969:
```suggestion
   * Build this definition's supersort relation from axioms that have the
```

Baltoli (approved):

unittests/compiler/subsortmap.cpp
Robertorosmaninho: It seems that this is not the case! Please check the latest commit of this PR!

unittests/compiler/subsortmap.cpp
Robertorosmaninho: Done!

PR #1051, opened 2024-05-09
Implement serialization/deserialization of new algorithm

dwightguth:
This PR implements a new version of the proof hint, version 9, which uses the newly created serialization and deserialization functions to serialize terms during the proof hint generation. Not only does this new serialization format take up less memory, it also is roughly 3x faster to serialize.

Note: there is one known issue with this serializer/deserializer: it incorrectly specifies the second sort parameter of the `inj` symbol when deserializing. We choose to ignore this known issue due to the fact that this deserializer is being consumed solely by code which transforms the resulting terms into an order-sorted representation by removing the injections entirely.


include/kllvm/binary/ProofTraceParser.h
```diff
@@ -337,30 +338,14 @@ class proof_trace_parser {
...
-    if (!read_uint64(ptr, end, pattern_len)) {
-      return nullptr;
-    }
-
-    if (std::distance(ptr, end) < pattern_len) {
-      return nullptr;
-    }
-    if (pattern_len > 0 && std::distance(ptr, end) > pattern_len) {
-      end = std::next(ptr, pattern_len);
-    }
```
Robertorosmaninho, line 361: Are these verifications really not necessary anymore? Or do they not fit the new version?

include/runtime/header.h
```diff
@@ -442,4 +461,16 @@ void sfprintf(writer *file, char const *fmt, Args &&...args) {
   }
 }
 
+template <typename... Args>
+void sfwrite(void const *ptr, size_t size, size_t nmemb, writer *file) {
```
Robertorosmaninho, line 465: What `nmemb` stands for?

Robertorosmaninho: Number of Memory Bytes?

lib/ast/definition.cpp
```diff
@@ -77,11 +77,16 @@ std::string get_raw_symbol_name(sort_category cat) {
...
+  auto k_decl = kore_symbol_declaration::create("rawKTerm", true);
   auto kitem = kore_composite_sort::create("SortKItem");
+  auto k = kore_composite_sort::create("SortK");
 
   decl->get_symbol()->add_sort(kitem);
   decl->get_symbol()->add_argument(kitem);
+  k_decl->get_symbol()->add_sort(kitem);
+  k_decl->get_symbol()->add_argument(k);
   mod->add_declaration(std::move(decl));
+  mod->add_declaration(std::move(k_decl));
```
Robertorosmaninho, line 89: Why are we replicating the logic of `decl` with `k_decl` here? 
Can we have some comments explaining/documenting this in the code if this is intended?

lib/binary/deserializer.cpp
```diff
@@ -37,4 +37,85 @@ sptr<kore_pattern> deserialize_pattern(std::string const &filename) {
   return deserialize_pattern(data.begin(), data.end());
 }
 
+// NOLINTNEXTLINE(*-cognitive-complexity)
+kore_header::kore_header(FILE *in) {
+  // NOLINTNEXTLINE(misc-redundant-expression)
+  if (fgetc(in) != 0x7f || fgetc(in) != 'K' || fgetc(in) != 'R'
+      || fgetc(in) != '2') {
+    throw std::runtime_error("invalid magic");
```
Robertorosmaninho, line 45: I didn't understand this error message. Can you explain it to me?

tools/kore-proof-trace/main.cpp
```diff
@@ -26,7 +31,10 @@ int main(int argc, char **argv) {
   cl::HideUnrelatedOptions({&kore_proof_trace_cat});
   cl::ParseCommandLineOptions(argc, argv);
 
-  proof_trace_parser parser(verbose_output, expand_terms_in_output);
+  FILE *in = fopen(header_path.getValue().c_str(), "r");
```
Robertorosmaninho, line 34: Shouldn't we need to close the file with `fclose `?

Robertorosmaninho (changes requested):
Overall, it looks good to me. I asked some questions and asked for some comments to understand the code better now and in the future. A small change is needed on the `kore-proof-trace`. I would also like some `// TODO:` pointing to the known issue of deserializing `inj` with the wrong second argument.

include/kllvm/binary/ProofTraceParser.h
dwightguth: I suppose we could probably still verify the 4-byte magic number, come to think of it, but the rest are relevant to things that are no longer part of the serialization format. I will make a small change here.

include/runtime/header.h
dwightguth: Number of members. As you can see, if `file` is a `FILE *`, it simply calls `fwrite`, which expects this signature.

Robertorosmaninho (approved):
I still would like to see the `TODOs` pointing to the known issue with deserializing `inj`, but overall, it LGTM!

theo25 (approved):
This looks good to me

PR #1053, opened 2024-05-12
Introducing `Ordinal` KORE Attribute

Robertorosmaninho:
Closes https://github.com/Pi-Squared-Inc/pi2/issues/1432
This PR introduces a new attribute specific to the LLVM Backend. It includes the `ordinal` attribute to the KORE definition when `preprocess` is called after parsing a definition. The content of the attribute is a `string_sort` with the value of the computed `axiom_ordinal`. 

A new C++ tool is planned to consume this feature instead of reading a definition with a bash script.
The main intention of this feature is to expose this information to the Python bindings to be used in Proof Hints integration tests.


dwightguth (approved):
this looks okay but I would want to make sure we've tested that the resulting attribute can be successfully grabbed by the python code that is consuming this before we merge this.

Robertorosmaninho:
@dwightguth, I added some Python tests that use the new ordinal attribute.
- Somethings to consider about this approach with Ordinal Attributes:
  - A downside of this approach is to perform a string comparison.
  - The upside is that we don't have to modify any K definitions to make that work.

- The other approach doesn't actually need the ordinal attribute; instead, we would need to create a new map `<rule_label, axiom>` and a function `get_axiom_from_rule_label`.
  - The downsides of this approach are that we must carry a new Map in every kore_definition, modify K definitions, and enforce that all rules we want to test must have a rule label.
  - The upside is that we can compare the ordinals instead of comparing strings.

include/kllvm/codegen/ProofEvent.h
```diff
@@ -104,7 +104,7 @@ class proof_event {
       llvm::BasicBlock *current_block);
 
   [[nodiscard]] llvm::BasicBlock *rewrite_event_pre(
-      sptr<kore_axiom_declaration> axiom, uint64_t arity,
+      sptr<kore_axiom_declaration> &axiom, uint64_t arity,
```
Baltoli, line 107: Why are these changed to be non-const references? In general I'm suspicious of code that needs a mutable reference to a smart pointer.

include/kllvm/ast/AST.h
```diff
@@ -1023,9 +1023,9 @@ class kore_definition {
   [[nodiscard]] std::list<kore_axiom_declaration *> const &get_axioms() const {
     return axioms_;
   }
-  [[nodiscard]] kore_axiom_declaration *
+  [[nodiscard]] sptr<kore_axiom_declaration>
   get_axiom_by_ordinal(size_t ordinal) const {
-    return ordinals_.at(ordinal);
+    return std::make_shared<kore_axiom_declaration>(*ordinals_.at(ordinal));
```
Baltoli, line 1028: I think we can have this return a const ref to the underlying ordinal, and use `return_value_policy::reference` (https://pybind11.readthedocs.io/en/stable/advanced/functions.html)

test/python/test_proof_trace_slow.py
```diff
@@ -37,14 +53,20 @@ def test_file(self):
 
            # check that the first event is the rewrite a() => b()
            self.assertTrue(trace.trace[0].is_step_event())
-           self.assertEqual(trace.trace[0].step_event.rule_ordinal, 95)
+           rule_ordinal = trace.trace[0].step_event.rule_ordinal
+           axiom = definition.get_axiom_by_ordinal(rule_ordinal).__repr__()
```
Baltoli, line 57:
```suggestion
           axiom = repr(definition.get_axiom_by_ordinal(rule_ordinal))
```
and similarly elsewhere.

Baltoli (changes requested):

include/kllvm/codegen/ProofEvent.h
Robertorosmaninho: I was fixing a clang-tidy issue, but I forgot to make them `const`, sorry!

include/kllvm/ast/AST.h
Robertorosmaninho: I think so, but using a smart pointer here isn't safer and a better design option?

Baltoli: A smart pointer would be safer and a better design option when compared to a _regular pointer_, but a `const&` reference is better than either.

Baltoli:
I think the ordinal attribute approach is fine; this is ultimately testing infrastructure and so the performance of a string comparison isn't critical. Avoiding modifications of the K source code is nice.

include/kllvm/ast/AST.h
Robertorosmaninho: Perfect then. Thanks for pointing that out!

include/kllvm/ast/AST.h
```diff
@@ -1023,9 +1023,9 @@ class kore_definition {
   [[nodiscard]] std::list<kore_axiom_declaration *> const &get_axioms() const {
     return axioms_;
   }
-  [[nodiscard]] kore_axiom_declaration *
+  [[nodiscard]] kore_axiom_declaration
```
Baltoli, line 1026: We don't want to copy this out, but rather return a const reference to the object that's stored.

Baltoli (approved):
One small change; otherwise good - feel free to merge once you've addressed it.

PR #1056, opened 2024-05-13
Introducing llvm-kompile `compute-ordinal` and `compute-loc` tools

Robertorosmaninho:
These tools are meant to replace the old `llvm-kompile-compute-ordinal` and `llvm-kompile-compute-loc` bash scripts with an appropriate C++ infrastructure. 

The new tools can work in two ways:
- Parsing the definition and looking for/using the K file location to compute the ordinal and location info.
- Using the `definition.kore` lines to compute the ordinal and returning the line in the definition when computing the location.

The interface of the new tools is defined as follows:
```bash
llvm-kompile-compute-loc [options] <kompiled-dir> <ordinal> [is-k-line]
```
and
```bash
llvm-kompile-compute-ordinal [options] <kompiled-dir> <line> [is-k-line]
```

in both cases, `is-k-line` is false by default to have the same behavior as the old scripts.


tools/llvm-kompile-compute-loc/main.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+  int64_t line_num = 0;
+  int64_t ordinal_num = 0;
+
+  std::ifstream file(definition);
+  std::string line;
+
+  while (std::getline(file, line)) {
+    line_num++;
+    if (line.starts_with("  axiom") || line.starts_with("    axiom")) {
```
theo25, line 66: I would like a more robust way to find axiom lines. Something that would ignore any number of whitespace and check that the first word is `"axiom"` maybe?

tools/llvm-kompile-compute-ordinal/main.cpp
```diff
@@ -0,0 +1,99 @@
...
+      location_att->get_arguments()[0].get());
+  std::string location = str_pattern->get_contents();
+
+  size_t l_paren = location.find_first_of('(');
+  size_t first_comma = location.find_first_of(',');
+  size_t length = first_comma - l_paren - 1;
+  return std::stoi(location.substr(l_paren + 1, length));
+}
+
+int64_t get_k_location(std::string &definition) {
```
theo25, line 42: This function returns an ordinal (not a location) so it should be renamed appropriately.

tools/llvm-kompile-compute-ordinal/main.cpp
```diff
@@ -0,0 +1,99 @@
...
+      if (start_line == stoi(line)) {
+        return axiom->get_ordinal();
+      }
+    }
+  }
+
+  return -1;
+}
+
+int64_t get_kore_location(std::string &definition) {
```
theo25, line 61: Similarly, this function returns an ordinal (not a location) so it should be renamed appropriately.

tools/llvm-kompile-compute-ordinal/main.cpp
```diff
@@ -0,0 +1,99 @@
...
+    cl::cat(ordinal_cat));
+
+cl::opt<std::string> line(
+    cl::Positional, cl::desc("<line>"), cl::Required, cl::cat(ordinal_cat));
+
+cl::opt<bool> is_k_line(
+    cl::Positional, cl::desc("[is-k-line]"), cl::init(false),
+    cl::cat(ordinal_cat));
+
+int64_t get_location(kore_axiom_declaration *axiom) {
```
theo25, line 27: This function is a duplicate of the function with the same name in the `llvm-kompile-compute-loc` tool. And more generally, there seems to exist a lot of duplicate code between the two tools. I am wondering if it is worth it to refactor the code in a way that avoids duplication. Maybe add some functions with the common functionality in `lib/codegen/Util.cpp`?

theo25 (dismissed):
This looks good in general but I left some comments about the code structure.

tools/llvm-kompile-compute-ordinal/main.cpp
Robertorosmaninho: I'm not sure if putting something only used by tools in the `codegen` directory is the appropriate strategy here. I would like to export it somewhere to reuse, but I am just not sure if this is the right place.

theo25: maybe @Baltoli and/or @dwightguth could offer a better suggestion

tools/llvm-kompile-compute-loc/main.cpp
Robertorosmaninho: I could address this by adding a new function in both tools code

tools/llvm-kompile-compute-loc/main.cpp
```diff
@@ -0,0 +1,102 @@
...
+      if (ordinal_num == std::stoi(ordinal)) {
+        file.close();
+        return line_num;
+      }
+      ordinal_num++;
+    }
+  }
+
+  file.close();
+  return -1;
```
Baltoli, line 84: Prefer `std::optional` if there's a failure mode; obviously -1 is _not_ a valid line number but it's cleaner to signal to callers that not every input will produce a valid output.

tools/llvm-kompile-compute-loc/main.cpp
```diff
@@ -0,0 +1,102 @@
...
+  int64_t ordinal_num = 0;
+
+  std::ifstream file(definition);
+  std::string line;
+
+  while (std::getline(file, line)) {
+    line_num++;
+    trim(line);
+    if (line.starts_with("axiom")) {
+      if (ordinal_num == std::stoi(ordinal)) {
```
Baltoli, line 75: We do have the ordinal global in scope, but this function would be cleaner taking it as a parameter. It's usually best to limit the usage of CLI globals outside of the top-level entrypoint, I find

tools/llvm-kompile-compute-loc/main.cpp
```diff
@@ -0,0 +1,102 @@
...
+        file.close();
+        return line_num;
+      }
+      ordinal_num++;
+    }
+  }
+
+  file.close();
```
Baltoli, line 83: This is handled automatically by `std::ifstream`'s destructor.

tools/llvm-kompile-compute-ordinal/main.cpp
Baltoli: I think this shared code could probably just live in a new module of the AST library?

tools/llvm-kompile-compute-loc/main.cpp
```diff
@@ -0,0 +1,102 @@
...
+using namespace llvm;
+using namespace kllvm;
+
+cl::OptionCategory loc_cat("llvm-kompile-compute-loc options");
+
+cl::opt<std::string> kompiled_dir(
+    cl::Positional, cl::desc("<kompiled-dir>"), cl::Required, cl::cat(loc_cat));
+
+cl::opt<std::string> ordinal(
+    cl::Positional, cl::desc("<ordinal>"), cl::Required, cl::cat(loc_cat));
```
Baltoli, line 20: Ordinals are integers; you can avoid a parsing step here by declaring a `cl::opt<int>` directly

Baltoli (changes requested):

tools/llvm-kompile-compute-loc/main.cpp
```diff
@@ -0,0 +1,102 @@
...
+      location_att->get_arguments()[0].get());
+  std::string location = str_pattern->get_contents();
+
+  size_t l_paren = location.find_first_of('(');
+  size_t first_comma = location.find_first_of(',');
+  size_t length = first_comma - l_paren - 1;
+  return std::stoi(location.substr(l_paren + 1, length));
+}
+
+int64_t get_k_location(std::string &definition) {
```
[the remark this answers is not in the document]
Robertorosmaninho: Not really, we can make it a const reference!

tools/llvm-kompile-compute-loc/main.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+cl::opt<std::string> ordinal(
+    cl::Positional, cl::desc("<ordinal>"), cl::Required, cl::cat(loc_cat));
+
+cl::opt<bool> is_k_line(
+    cl::Positional, cl::desc("[is-k-line]"), cl::init(false), cl::cat(loc_cat));
+
+int64_t get_location(kore_axiom_declaration *axiom) {
+  auto *location_att
+      = axiom->attributes().get(attribute_set::key::Location).get();
```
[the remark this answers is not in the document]
Robertorosmaninho: Thanks!

tools/llvm-kompile-compute-loc/main.cpp
```diff
@@ -0,0 +1,102 @@
...
+cl::opt<std::string> kompiled_dir(
+    cl::Positional, cl::desc("<kompiled-dir>"), cl::Required, cl::cat(loc_cat));
+
+cl::opt<std::string> ordinal(
+    cl::Positional, cl::desc("<ordinal>"), cl::Required, cl::cat(loc_cat));
+
+cl::opt<bool> is_k_line(
+    cl::Positional, cl::desc("[is-k-line]"), cl::init(false), cl::cat(loc_cat));
+
+int64_t get_location(kore_axiom_declaration *axiom) {
```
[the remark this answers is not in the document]
Robertorosmaninho: Thanks!

tools/llvm-kompile-compute-loc/main.cpp
```diff
@@ -0,0 +1,102 @@
...
+  size_t first_comma = location.find_first_of(',');
+  size_t length = first_comma - l_paren - 1;
+  return std::stoi(location.substr(l_paren + 1, length));
+}
+
+int64_t get_k_location(std::string &definition) {
+  // Parse the definition.kore to get the AST.
+  kllvm::parser::kore_parser parser(definition);
+  auto kore_ast = parser.definition();
+  kore_ast.get()->preprocess();
```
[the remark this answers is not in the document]
Robertorosmaninho: Thanks for the reminder!

tools/llvm-kompile-compute-loc/main.cpp
```diff
@@ -0,0 +1,102 @@
...
+  if (axiom) {
+    auto start_line = get_location(axiom);
+    return start_line;
+  }
+
+  return -1;
+}
+
+// trim the string from the start
+inline void trim(std::string &s) {
```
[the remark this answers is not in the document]
Robertorosmaninho: All right then! .Just because it is a small and self-contained function, that is only a function because it would make the code cleaner. But I can remove the attribute!

tools/llvm-kompile-compute-ordinal/main.cpp
Robertorosmaninho: Sure, but why not in a new header module in `kllvm/util`?

Robertorosmaninho: `trim` is also shared between the two tools. Should I put in this new module as well?

Baltoli: Yes, anything you've shared between these tools should be moved to the new module.

Robertorosmaninho: It should be in `kllvm/util` or `kllvm/ast`?

Baltoli: I think `ast` is fine here; these are all functions that operate on or are related to the AST data structures.

Robertorosmaninho: Perfect! Thanks!

include/kllvm/ast/util.h
```diff
@@ -0,0 +1,28 @@
...
+  return std::stoi(location.substr(l_paren + 1, length));
+}
+
+// trim the string from the start
+[[nodiscard]] std::string trim(std::string s) {
+  s.erase(s.begin(), std::find_if(s.begin(), s.end(), [](int c) {
+            return !std::isspace(c);
+          }));
+  return s;
+}
```
Baltoli, line 28: These need to live in an implementation file rather than a header; otherwise you'll get multiple definition errors. The header with declarations also needs include guards.

Baltoli (changes requested):

Baltoli (approved):

Baltoli:
https://github.com/runtimeverification/devops/issues/214 FWIW, the automerge noise above is because there was an un-dismissed review, which the automerger doesn't handle gracefully.

PR #1057, opened 2024-05-14
[DNM] set the linkage of some symbols to internal

dwightguth:
Each semantics exports a large number of semantics-specific global symbols. If we want to create applications which link in libraries for multiple semantics of the llvm backend, this is a problem, because symbols will collide with one another.

This PR is the first step of attempting to resolve this problem. Some symbols that are currently externally visible can simply be replaced with internal linkage in llvm, thus allowing them to be automatically renamed by the linker when they occur in multiple semantics.

Note that this does not fully resolve the problem. There are still two other classes of symbols that are an issue:

1. symbols that are generated to interface the code generator with the runtime, which the runtime needs to be able to call, but which need to be made somehow local to the semantics.
2. symbols which are intended to be part of the public API of the semantics, and thus need to be callable by applications which link against these libraries.

These will require a more complex solution that will come in a follow-up pull request.

Don't merge this yet. I am experimenting with another solution right now.

### DIFF
--- a/debug/kgdb.py
+++ b/debug/kgdb.py
@@ -375,13 +375,13 @@ class termPrinter:
         self.kompiled_dir = getKompiledDir()
 
     def getSymbolNameForTag(self, tag):
-        return gdb.lookup_global_symbol("table_getSymbolNameForTag").value()[tag]
+        return gdb.lookup_static_symbol("table_getSymbolNameForTag").value()[tag]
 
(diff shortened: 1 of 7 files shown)


Baltoli (approved):

Baltoli:
The changes here seem locally fine so I've approved in the event that your experimental alternative doesn't work out.

PR #1059, opened 2024-05-15
Using `map` instead of `unordered_map` to ensure reproducibility of kore definition generation

Robertorosmaninho:
We have seen that the `definition.kore` has been produced in a different order depending on the OS, which creates a flaky error for some tools like `llvm-komplie-compute-ordinal` and `llvm-kompile-compute-loc`, an error on #1053 MacOS tests, and some frontend errors as well. 

This PR aims to fix that issue by replacing the data structure used to store the attributes.


Baltoli (approved):

Baltoli:
Will need formatting to fix CI but otherwise looks good!

PR #1060, opened 2024-05-15
--hidden-visibility flag to llvm-kompile

dwightguth:
This flag is used to assist with the process of building an application that links against more than one llvm backend semantics.

It can be used in roughly the following way:

1. Compile the llvm backend with the `clang++` flag `-fvisibility=hidden`.
2. Run `llvm-kompile definition.kore dt library --hidden-visibility -- -shared entry.cpp -o definition.so` in the kompiled directory.
3. Link `definition.so` into your application. It will have all the symbols of `entry.cpp` and `entry.cpp` will be able to access any symbol in the semantics, but those internal symbols will not be visible to the application.

The application can do this with multiple semantics and there will be no linker issue. I have tested some basic examples that show that each semantics can separately apply its own rewrite rules. There is no test in this PR because the test harness does not allow us to rebuild the llvm backend with a particular set of build flags just for one test, and it didn't seem worth it to put all that effort in just for one test. When we create downstream projects that use this feature, we will be sure to include tests for this feature in those repositories.

### DIFF
--- a/bin/llvm-kompile-clang
+++ b/bin/llvm-kompile-clang
@@ -151,6 +151,12 @@ if ! $save_temps; then
   trap 'rm -rf "$tmpdir"' INT TERM EXIT
 fi
 
+if [[ "$OSTYPE" == "darwin"* ]]; then
+  set_visibility_hidden="$LIBDIR/libSetVisibilityHidden.dylib"
+else
+  set_visibility_hidden="$LIBDIR/libSetVisibilityHidden.so"
+fi
+
 # When building the Python AST module, there's no runtime and no main file, so
 # we skip this entire step. The library code is just C++, so we can skip
 # straight to invoking the C++ compiler.
@@ -167,7 +173,7 @@ if [ "$main" != "python_ast" ]; then
       fi
       if $visibility_hidden; then
         modhidden="$tmpdir/hidden.bc"
-        run @OPT@ "$modopt" -load-pass-plugin "$LIBDIR/libSetVisibilityHidden.so" -set-visibility-hidden -o "$modhidden"
+        run @OPT@ "$modopt" -load-pass-plugin "$set_visibility_hidden" -set-visibility-hidden -o "$modhidden"
         modopt="$modhidden"
       fi
       run @LLC@ \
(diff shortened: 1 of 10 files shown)


dwightguth:
Something seems broken on ARM. I may have to investigate this.

lib/set-visibility-hidden/SetVisibilityHidden.cpp
```diff
@@ -0,0 +1,62 @@
...
+      }};
+}
+
+#ifndef LLVM_BYE_LINK_INTO_TOOLS
+extern "C" __attribute__((visibility("default")))
+LLVM_ATTRIBUTE_WEAK ::llvm::PassPluginLibraryInfo
+llvmGetPassPluginInfo() {
+  return get_set_visibility_hidden_plugin_info();
+}
+#endif
```
Robertorosmaninho, line 62: It has been a while since the last time I wrote an LLVM Pass, so I would like to ask if it's really necessary to register both in the Legacy and in the New PassManager? Or is this a restriction from our source code?

Baltoli: For the moment we need to use both because of the range of versions we support; I _think_ once we are able to fully upgrade to Ubuntu Noble then we'll be able to drop the requirement.

bin/llvm-kompile-clang
```diff
@@ -160,6 +165,11 @@ if [ "$main" != "python_ast" ]; then
       if ! $link; then
         modasm="$output_file"
       fi
+      if $visibility_hidden; then
+        modhidden="$tmpdir/hidden.bc"
+        run @OPT@ "$modopt" -load-pass-plugin "$LIBDIR/libSetVisibilityHidden.so" -set-visibility-hidden -o "$modhidden"
```
Baltoli, line 170: I saw you had some issues on macOS that are now fixed; I wanted to confirm that the compiled pass library actually gets this name (rather than `.dylib`) on macOS.

Baltoli (approved):
One question but if that's resolved feel free to merge.

lib/set-visibility-hidden/SetVisibilityHidden.cpp
Robertorosmaninho: That's good to know! Thanks!

theo25 (approved):
Looks good to me

PR #1061, opened 2024-05-16
add axiom filtering property to python bindings

dwightguth:
This adds a boolean property to kore_axiom_declaration objects which will be `True` if the axiom is used by the LLVM backend and `False` if the LLVM backend filters it out.


Robertorosmaninho:
Can we have some unit tests like (or included in) [here](https://github.com/runtimeverification/k/blob/master/pyk/src/tests/integration/kllvm/test_utils.py)?

Robertorosmaninho (approved):
I would like some tests here or in the PyK test suite. Otherwise, LGTM!

Robertorosmaninho (approved):
LGTM!

theo25 (approved):
Looks good to me

PR #1066, opened 2024-05-22
Drop intel macOS support in CI

Baltoli:
Part of: https://github.com/runtimeverification/k/issues/4368

We no longer want to support these machines, and so we should remove them from all the CI matrices that we're using across the organisation.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,8 +17,6 @@ jobs:
         include:
           - runner: [self-hosted, linux, normal]
             os: ubuntu-20.04
-          - runner: macos-12
-            os: macos-12
           - runner: MacM1
             os: self-macos-12
     runs-on: ${{ matrix.runner }}


Baltoli:
Worth noting that this will bring CI times for the LLVM backend down to <15 mins, which is a nice improvement.

Robertorosmaninho (approved):
LGTM! Great improvement!

Robertorosmaninho:
Will we update our host m1 machine to have a newer macOS version? We're already on version 14. It should be safe to drop 12 and only keep supporting and testing 13 and 14!

Baltoli:
> Will we update our host m1 machine to have a newer macOS version? We're already on version 14. It should be safe to drop 12 and only keep supporting and testing 13 and 14! 

Yes, but doing so requires some work on freeman's part so we'll do that upgrade when he has the bandwidth to do so.

Issue #1055, opened 2024-05-13
Intermittent maven dependency failures on CI


dwightguth:
Right now, each time we invoke llvm-kompile-testing, we are invoking maven. A consequence of this is that if maven has a transient failure during that invocation due to its communication with the internet, the test fails.

We don't really need to be invoking maven each time. We ought to be able to use the maven appassembler plugin to create a script that we can invoke from the llvm-kompile-testing script which will invoke the jvm directly. If we do this, then the intermittent failures will be limited only to if the maven internet fails during the build process itself, which is much more rare.

Doing this ought to reduce the frequency of flaky tests on CI, which have been happening somewhat frequently recently.

PR #1067, opened 2024-05-22
Don't invoke Maven every time we run a test

Baltoli:
Previously, the testing script would invoke Maven every time it compiled a pattern-matching tree; this was slow and made the CI process prone to intermittent network failures. This PR addresses the issue by instead directly invoking the compiled JAR file that we already built with Maven.

Fixes https://github.com/runtimeverification/llvm-backend/issues/1055

### DIFF
--- a/bin/llvm-kompile-testing
+++ b/bin/llvm-kompile-testing
@@ -11,7 +11,17 @@ trap 'rm -rf "$dt_dir"' INT TERM EXIT
 definition=$(realpath "$1")
 mode="$2"
 shift; shift
-( cd "@PROJECT_SOURCE_DIR@"/matching && mvn exec:java -Dexec.args="$definition qbaL $dt_dir 1" -q )
+
+matching_dir="@PROJECT_SOURCE_DIR@/matching"
+installed_jar="$matching_dir/target/llvm-backend-matching-1.0-SNAPSHOT-jar-with-dependencies.jar"
+
+if [ ! -f "$installed_jar" ]; then
+  echo "Pattern matching compiler isn't installed!"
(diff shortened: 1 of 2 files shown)


bin/llvm-kompile-testing
```diff
@@ -11,7 +11,17 @@ trap 'rm -rf "$dt_dir"' INT TERM EXIT
...
 shift; shift
-( cd "@PROJECT_SOURCE_DIR@"/matching && mvn exec:java -Dexec.args="$definition qbaL $dt_dir 1" -q )
+
+matching_dir="@PROJECT_SOURCE_DIR@/matching"
+installed_jar="$matching_dir/target/llvm-backend-matching-1.0-SNAPSHOT-jar-with-dependencies.jar"
+
+if [ ! -f "$installed_jar" ]; then
+  echo "Pattern matching compiler isn't installed!"
+  echo "  looking for a jar file at $installed_jar"
+  exit 1
```
Robertorosmaninho, line 21: Why are we exiting here?

Baltoli: Because the pattern matching compiler jar isn't installed; how would you propose we continue execution?

Robertorosmaninho: The message above makes me feel that the script will look for the jar and not finish its execution...
I think I would create an else branch to build it as we were before, no?

Robertorosmaninho: Alright! Makes sense!

dwightguth (approved):

gtrepta (approved):

PR #1068, opened 2024-05-22
Apply backwards-compatible changes for LLVM 18

Baltoli:
This PR makes a series of small housekeeping changes to the backend that allow us to build against LLVM 18, which we will need soon to support Ubuntu 24.04.

The changes required are all self-contained and can be reviewed commit-by commit:
- Build system update to detect LLVM 18 and consistently use fmt's header-only distribution
- Fully adopt the opaque pointer API everywhere it can be; the remaining places we were using typed pointers are removed in LLVM 18.
- Conditional compilation to support the various different APIs LLVM has had for creating `malloc` calls; we'll be able to simplify this soon.
- Conditional compilation for changed enum visibilities across LLVM versions.

Note that we're not yet running CI against LLVM 18; that will require a Nixpkgs update which I'll handle in a separate PR. This PR is just backwards compatible changes that _allow_ us to adopt 18 with no extra work.


lib/codegen/CreateTerm.cpp
```diff
@@ -265,20 +265,11 @@ llvm::Value *allocate_term(
...
-    set_debug_loc(&block->back());
-  }
-
-#if LLVM_VERSION_MAJOR < 16
-  malloc->insertAfter(&block->back());
-#else
-  malloc->insertInto(block, block->end());
-#endif
-
+  set_debug_loc(malloc);
```
Robertorosmaninho, line 272: I'm not sure if we keep the same behavior here. Do we?

Robertorosmaninho (approved):
A small comment, but overall LGTM

lib/codegen/CreateTerm.cpp
Baltoli: Slightly different potentially, depending on which version of LLVM you're on. The difference is that the debug location is set on reaching the _second_ of two instructions that will always appear together in sequence, rather than the first, so I don't think it's important, and helps keep the code cleaner.

gtrepta (approved):

lib/codegen/Decision.cpp
```diff
@@ -1379,11 +1373,13 @@ void make_match_reason_function(
   llvm::BasicBlock *fail
       = llvm::BasicBlock::Create(module->getContext(), "fail", match_func);
   llvm::PHINode *fail_subject = llvm::PHINode::Create(
-      llvm::Type::getInt8PtrTy(module->getContext()), 0, "subject", fail);
+      llvm::PointerType::getUnqual(module->getContext()), 0, "subject", fail);
   llvm::PHINode *fail_pattern = llvm::PHINode::Create(
-      llvm::Type::getInt8PtrTy(module->getContext()), 0, "pattern", fail);
+      llvm::PointerType::getUnqual(llvm::Type::getInt8Ty(module->getContext())),
```
dwightguth, line 1378: why does this call site pass an argument type?

dwightguth (approved):
looks good but it looks like there might be one minor fix to be applied.

lib/codegen/Decision.cpp
```diff
@@ -1379,11 +1379,13 @@ void make_match_reason_function(
...
+      0, "pattern", fail);
   llvm::PHINode *fail_sort = llvm::PHINode::Create(
-      llvm::Type::getInt8PtrTy(module->getContext()), 0, "sort", fail);
+      llvm::PointerType::getUnqual(llvm::Type::getInt8Ty(module->getContext())),
```
theo25, line 1381: I have the same question as @dwightguth for this call site too.

theo25 (approved):

Baltoli:
I have added a comment explaining why the explicitly set element types are necessary.

PR #1069, opened 2024-05-22
Write streaming event parser for proof hints

dwightguth:
This is a pretty straightforward diff. We can reuse the code for parsing events and the proof trace header, so we just need to create an entry point to the streaming parser that invokes the relevant parsing logic and returns events one at a time, and then write code to test it. We test it by modifying lit tests that test the proof hints so that they test both parsers. This means we get the advantage of the entire coverage set of tests that we imported from the math proof repo.


theo25 (approved):
Looks good to me

Robertorosmaninho (approved):
LGTM

PR #1072, opened 2024-05-24
Update to build on Ubuntu Noble

Baltoli:
This PR sets up a parallel build for Ubuntu 24.04 (Noble) alongside our current 22.04 (Jammy) build. Everywhere where we previously built only on Jammy, we now build on both:
- Testing build-from-source on PRs
- Testing Ubuntu `.deb` package creation on PRs
- Releasing `.deb` packages on release

The changes here are largely procedural; there are only a couple of places we need to do things that aren't just pattern-matching or version parameterisation:
- Use the modern `pipx` tool to install `lit`
- Bump the version of `pybind` we use to be compatible with the one used in Noble (the one we had previously was backwards- but not forwards-compatible).

On Noble, we use LLVM 17 for the time being because of a [known bug affecting LLVM 18](https://bugs.launchpad.net/ubuntu/+source/llvm-toolchain-18/+bug/2064187) in the package repositories that prevents us from using it to package `.deb`s. It does not affect source builds in the same way; there's no real downside to doing this other than not having a totally up-to-date LLVM dependency.


Baltoli:
Note that this PR will look like checks are missing; this is because of the matrix renaming. I'll update the branch protection rules when the PR is ready to merge.

.github/workflows/release.yml
```diff
@@ -24,40 +24,47 @@ jobs:
             --target ${{ github.sha }}
 
 
-  build-jammy-package:
-    name: 'Build Ubuntu Jammy Package'
+  build-ubuntu-packages:
+    name: 'Build Ubuntu Packages'
```
tothtamas28, line 28: For consistency with `test.yml`:

```suggestion
  build-ubuntu-package:
    name: 'Build Ubuntu package'
```

Consistency can also be ensured by doing the renaming in the other workflow. I prefer the singular form:
* `Build Ubuntu package (jammy, 15)`
* `Build Ubuntu package (noble, 17)`

.github/workflows/ci-tests.sh
```diff
@@ -1,16 +1,24 @@
...
 mkdir build
 
 pushd build
-cmake -DCMAKE_EXPORT_COMPILE_COMMANDS=On -DBUILD_TESTS=On -DCMAKE_INSTALL_PREFIX=install ..
+cmake ..                                                    \
+  -DCMAKE_C_COMPILER="$(which clang-${llvm_version})"       \
+  -DCMAKE_CXX_COMPILER="$(which clang++-${llvm_version})"   \
+  -DCMAKE_EXPORT_COMPILE_COMMANDS=On                        \
+  -DBUILD_TESTS=On                                          \
+  -DCMAKE_INSTALL_PREFIX=install
```
tothtamas28, line 15:
```suggestion
cmake ..                                                  \
  -DCMAKE_C_COMPILER="$(which clang-${llvm_version})"     \
  -DCMAKE_CXX_COMPILER="$(which clang++-${llvm_version})" \
  -DCMAKE_EXPORT_COMPILE_COMMANDS=On                      \
  -DBUILD_TESTS=On                                        \
  -DCMAKE_INSTALL_PREFIX=install
```

tothtamas28 (approved):

package/debian/rules.noble
Robertorosmaninho, line 1: Can we delete the commands commented in this file, or is this still WIP?

Robertorosmaninho (approved):
Besides my question and Tamás suggestions, which I agree! It LGTM!

PR #1075, opened 2024-05-31
Add python bindings for proof hint streaming parser

dwightguth:
Previously we did not have any python bindings for the streaming parser I created.

This PR adds the relevant pybind code as well as adding a python unit test for that code. A slight change is made to the ownership semantics of proof_trace_buffer and its derived classes in order to allow the python bindings to work as expected.


tools/kore-proof-trace/main.cpp
```diff
@@ -42,8 +42,9 @@ int main(int argc, char **argv) {
 
   if (use_streaming_parser) {
     std::ifstream file(input_filename, std::ios_base::binary);
-    proof_trace_file_buffer buffer(file);
-    llvm_rewrite_trace_iterator it(buffer, header);
+    auto *buffer = new proof_trace_file_buffer(std::move(file));
+    llvm_rewrite_trace_iterator it(
+        std::unique_ptr<proof_trace_file_buffer>(buffer), header);
```
theo25, line 47: nitpick: same as above

```suggestion
    auto buffer = std::make_unique<proof_trace_file_buffer>(std::move(file));
    llvm_rewrite_trace_iterator it(buffer, header);
```

bindings/python/ast.cpp
```diff
@@ -457,6 +457,32 @@ void bind_proof_trace(py::module_ &m) {
...
+      llvm_rewrite_trace_iterator,
+      std::shared_ptr<llvm_rewrite_trace_iterator>>(
+      proof_trace, "llvm_rewrite_trace_iterator")
+      .def_static(
+          "from_file",
+          [](std::string const &filename, kore_header const &header) {
+            std::ifstream file(filename, std::ios_base::binary);
+            auto *buffer = new proof_trace_file_buffer(std::move(file));
+            return llvm_rewrite_trace_iterator(
+                std::unique_ptr<proof_trace_file_buffer>(buffer), header);
```
theo25, line 480: nitpick: I think the recommended practice here, is to call `std::make_unique` instead of a hardcoded use of `new`.

```suggestion
            auto buffer = std::make_unique<proof_trace_file_buffer>(std::move(file));
            return llvm_rewrite_trace_iterator(buffer, header);
```

theo25 (approved):
Looks good to me. I left two minor comments

theo25 (approved):

Robertorosmaninho (approved):
LGTM. I would like to see a follow-up PR adding these new bindings in [`prooftrace.py`](https://github.com/runtimeverification/k/blob/master/pyk/src/pyk/kllvm/hints/prooftrace.py) with appropriate in-line documentation following our current pattern to expose the hints bindings to the Math Proof Generation Team!

PR #1076, opened 2024-05-31
Remove sentinel for end of KORE terms in the proof hint format

theo25:
This PR removes the sentinel at the end of a KORE terms as it would appear in the proof hint trace. We do not need this sentinel to properly parse the trace because the KORE term binary format includes length information.

The hints version has been bumped to 11 and the relevant tests have been updated.

### DIFF
--- a/docs/proof-trace.md
+++ b/docs/proof-trace.md
@@ -54,13 +54,13 @@ hook              ::= WORD(0xAA) name symbol_name location arg* WORD(0xBB) kore_
 ordinal           ::= uint64
 arity             ::= uint64
 boolean_result    ::= uint8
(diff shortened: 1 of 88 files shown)


dwightguth (approved):
This code looks just fine for what it's doing, so I'm approving, but my understanding of the discussed proposed change was the plan to also remove the `0xFFFFFFFFFFFFFFFF` word as well. Did that just get miscommunicated or did you try it and discover a reason why it wouldn't actually work to remove?

dwightguth:
As Theo mentioned on slack, we are keeping the `0xFFFFFFFFFFFFFFFF` sentinel because it is actually used by the parser to determine how parsing should proceed.

PR #1081, opened 2024-06-03
add libunwind as dependency

dwightguth:

This PR adds the libunwind library as a runtime dependency to the llvm backend. Libunwind is going to be used by the garbage collector to walk the stack and find the GC roots for each stack frame. The code to actually generate the stack maps and walk the stack will be added in a future PR.


theo25 (approved):
This looks good to me, but maybe a person more familiar with all the different build options should review this because I cannot tell that these are all the places where the new library needs to be added.

dwightguth:
I'm going to mark this as a draft again. I thought I had caught every place where I needed to add this dependency but it looks like I missed some and I'm not going to get to fixing it until tomorrow.

dwightguth:
I mean, probably the reason it's passing CI is because the library isn't actually used yet, so in the places where I forgot to pass it, nothing bad happens. I don't believe that adding more tests will address that.

PR #1082, opened 2024-06-03
add flag to control whether gc should be enabled or not

dwightguth:

This PR adds a global variable that controls whether or not garbage collection is enabled. This boolean flag is disabled during certain call frames that it would be difficult to walk the stack of. In particular, we disable the flag during:

* allocations that occur inside immer
* calls to hooks
* allocations that occur inside GMP and MPFR
* construction of the initial configuration

The boolean flag is currently unused. It will be used in a future pull request to determine whether the allocator is allowed to trigger garbage collection at the current time.


theo25 (approved):
Looks good to me

PR #1083, opened 2024-06-03
simplify gc code for alwaysgc roots

dwightguth:
When we were trying to fix the GC for `kompile -O3` on EVM, we added some logic to deal with the case where a map/list/set was a garbage collection root. Over the course of several iterations, that code evolved, and eventually became rather complex. However, it came to my attention analyzing the code that some of the complexity that was added early on was made redundant by later complexity. So we are actually able to simplify somewhat.

What this PR does is remove the `is_block` boolean parameter everywhere. This was being done because we were passing a `block *` for map/list/set pointers that had been promoted from the alwaysgcspace to the youngspace. However, the new derived pointer logic we added can handle derived pointers to embedded map/list/set objects now, so this is no longer necessary.

I may make further simplifications to this logic in the near future; I am considering how it may interact with changes that would make the garbage collector walk the stack. However, this is an intermediate step that I was able to write and test to ensure it didn't break anything, and since it is overall a deletion of code, I wanted to get it in first.


runtime/collect/collect.cpp
```diff
@@ -380,14 +374,13 @@ bool store_map_for_gc(void **roots, map *ptr) {
   *offset = 16;
   auto *child = (map *)(hdr + 2);
   *child = std::move(*ptr);
-  *roots = mem;
-  return true;
+  *roots = child;
```
theo25, line 377: Could you explain the reason we store the `child` pointer here instead of the `mem` pointer as it was before? There is a similar change in the corresponding functions for sets, lists, and rangemaps.

dwightguth: okay, yes, I can explain. So, the garbage collection roots in this case are collections that were stored on the alwaysgcspace. In order to allow them to survive garbage collection, which alwaysgcspace objects cannot normally do, we are promoting them in this function to objects on the youngspace. This requires putting them as the child of a regular symbol (which is in this case a special symbol created for this specific purpose), and giving that regular symbol a block header. The `mem` pointer points to that block header. Previously, `is_block` was set to true in this case, which caused the garbage collector to know that the pointer pointed to a symbol and not a collection, and caused the `load_for_gc` function to know that it had to offset the pointer to get back to the collection.

However, logic already exists in the garbage collector (due to the most recent set of bug fixes) which allows the garbage collector to handle gc roots of sort map/list/set. It does this by using a special offset word embedded in blocks to be able to find the root of the symbol that the map/list/set is a derived pointer of, and then migrate the entire block to the tospace. This is the logic that is now being used in the case that was previously using `is_block`. In essence, instead of telling the gc directly "no, this is not actually a collection, it's a block containing a collection", we treat the root like any other derived pointer, allowing the collector to use the offset embedded in the block to find the root pointer of the allocation.

However, this means that this function needs to return a derived pointer to the collection, rather than the base pointer of the allocation. That is the change being made in this method; instead of returning the base pointer of the object, we are returning the derived pointer to the child of the object that will then be used by the gc algorithm to reconstruct the base pointer.

theo25: Makes sense. I am approving since everything looks good.

theo25 (approved):

PR #1089, opened 2024-06-05
add kore_clear function that clears alwaysgcspace on every step

dwightguth:

This PR separates koreCollect's logic for collecting the gc'd heap from its logic for resetting the alwaysgcspace arena. We separate the logic for resetting the alwaysgcspace arena into a new method, `kore_clear`. This function is no longer called by the garbage collector, but it is called after every rewrite step. This is a slight increase in overhead since we are clearing it more frequently, but the flip side is that it should have better locality which may improve the processor's dcache, since we are clearing it much more frequently. The clear itself is a very fast operation, only involving a few simple pointer operations.


theo25 (approved):
Looks good to me.

PR #1090, opened 2024-06-06
Update name of `EventType` and expose `__repr__` from `llvm_rewrite_trace_iterator`

Robertorosmaninho:
This small PR only updates the `LLVMEventType` name to `EventType` to preserve the `LLVM` prefix to be used by the wrapper classes created on PyK that export these bindings. 

This PR also exposes the `__repr__` function of `llvm_rewrite_trace_iterator`. The print function was already implemented but has yet to be exposed; we decided to expose it as the binding as it can be helpful for debugging.


Robertorosmaninho:
Follow up PR: https://github.com/runtimeverification/k/pull/4426

theo25 (approved):
Looks good to me.

PR #1094, opened 2024-06-17
update llvm-kompile-compute-loc and llvm-kompile-compute-ordinal

dwightguth:
These binaries were not quite feature-complete with their original intended purpose, probably primarily because at the time at which they were reimplemented, the past version of them was somewhat broken.

This PR adds the following features:

1. Outputting the k location is the default in llvm-kompile-compute-loc
2. If a k location does not exist, it falls back to the kore location rather than crashing.
3. The output is formatted with both a filename and a line number.
4. In llvm-kompile-compute-ordinal, you can specify a filename and it will search for the ordinal of a k rule on that line of that file in the input definition.

### DIFF
--- a/tools/llvm-kompile-compute-loc/main.cpp
+++ b/tools/llvm-kompile-compute-loc/main.cpp
@@ -49,20 +49,20 @@ get_kore_location(std::string &definition, int const &ordinal) {
 }
 
 std::optional<std::pair<std::string, int64_t>>
-get_k_location(std::string &definition, int const &ordinal) {
+get_k_or_kore_location(std::string &definition, int const &ordinal) {
   // Parse the definition.kore to get the AST.
   kllvm::parser::kore_parser parser(definition);
   auto kore_ast = parser.definition();
   kore_ast->preprocess();
 
   auto axiom = kore_ast->get_axiom_by_ordinal(ordinal);
-  auto result = get_start_line_location(axiom);
-  if (result) {
-    return result;
+  auto k_loc = get_start_line_location(axiom);
+  if (k_loc) {
+    return k_loc;
   }
-  auto result2 = get_kore_location(definition, ordinal);
-  if (result2) {
-    return std::make_pair<std::string, uint64_t>("", *result2);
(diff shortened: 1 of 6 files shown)


tools/llvm-kompile-compute-loc/main.cpp
```diff
@@ -60,17 +48,44 @@ get_kore_location(std::string &definition, int const &ordinal) {
...
-                            : get_kore_location(definition, ordinal);
+  std::optional<std::pair<std::string, uint64_t>> location;
+  if (is_kore_line) {
+    auto line = get_kore_location(definition, ordinal);
+    location = std::make_pair(definition, *line);
+  } else {
+    location = get_k_location(definition, ordinal);
+    if (location->first.empty()) {
+      location->first = definition;
+    }
```
Robertorosmaninho, line 84: Why we're testing this and doing this assignment instead of directly returning the definition in `get_k_location`?

tools/llvm-kompile-compute-loc/main.cpp
```diff
@@ -60,17 +48,44 @@ get_kore_location(std::string &definition, int const &ordinal) {
...
+  auto axiom = kore_ast->get_axiom_by_ordinal(ordinal);
+  auto result = get_start_line_location(axiom);
+  if (result) {
+    return result;
+  }
+  auto result2 = get_kore_location(definition, ordinal);
+  if (result2) {
+    return std::make_pair<std::string, uint64_t>("", *result2);
+  }
+  return std::nullopt;
```
Robertorosmaninho, line 67: I think the function's implementation changed enough to justify a new name, perhaps something like `get_k_or_kore_location` or just `get_location`.

To be more explicit, I would also rename the result variable names to `k_loc` and `kore_loc`.

Robertorosmaninho (changes requested):
The overall implementation LGTM, however, I believe we can improve it a bit! Thanks for this upgrade, Dwight!

Robertorosmaninho (approved):
LGTM! Thanks for addressing the comments!

PR #1097, opened 2024-06-26
fix llvm-kompile-codegen so it optimizes code properly

dwightguth:
When we migrated the llvm backend optimization pipeline from being applied by `opt` to being applied by `llvm-kompile-codegen`, we made an error. The code generator was not, in fact, applying the `opt` transformation pipeline correctly. This means that `kompile -O2` will not, in fact, apply `llc -O2` to the bitcode generated by the llvm backend, leading to a significant performance regression.

While this PR does not entirely fix the issue (an upstream change to the K frontend is also required), it fixes the issue within this repository. We see a roughly 1.5x speedup in runtime when `-O2` is passed to kompile. We do lose something in compilation time, however. This is normal; compiling with optimizations is more expensive, we just weren't doing it correctly before.

The main changes in this pull request are threefold:

1. Convert optimizer code to new pass manager.
2. Make sure to run the middle-end optimizer.
3. Convert `TailCallElimination` and `Mem2Reg` from required to optional passes on -O0. In order to make the tail call optimization still work correctly without `TailCallElimination` manually marking tail calls as `tail`, we instead explicitly mark them as `musttail` in the IR generated by the code generator.

### DIFF
--- a/.github/workflows/ci-tests.sh
+++ b/.github/workflows/ci-tests.sh
@@ -17,7 +17,7 @@ make -j$(nproc) install
 popd
 
 pushd matching
-mvn package
+mvn -B package
 popd
 
 export PATH="$(realpath ./build/install/bin):$(realpath ./build/bin):/usr/lib/llvm-${llvm_version}/bin:$PATH"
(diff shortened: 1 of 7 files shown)


Robertorosmaninho (approved):
LGTM

PR #1098, opened 2024-07-02
Allow `kore-proof-trace` to display `label` and `location` info for improve debugging if a kore definition file is given

Robertorosmaninho:
Fix: https://github.com/Pi-Squared-Inc/pi2/issues/1562

The Math Proof Team from Pi2 required that rule hints contain more high-level information, such as the `label` attribute, if present, and the k file `location` of a rule, instead of only printing the `ordinal` and `arity`, to help them debug their code.

With that in mind, we propose that `kore-proof-trace` could take an extra/optional argument of a kore definition file. When this file is passed in the CLI, the tool will parse and preprocess it so we can extract these high-level info from the definition and print it in the hints human-readable output! If the definition isn't passed to the tool, it keeps this old behavior.

The new output looks like this:
- If the rule doesn’t have a label or location attr
```
rule: 101 0 [] []
```
- If the rule has both attributes
```
rule: 97 2 [TEST-PROOF-TRACE.a-to-b] [/home/robertorosmaninho/rv/temp/test-proof-trace.k:7]
  Var'Unds'DotVar0 = kore[15]
  Var'Unds'DotVar1 = kore[4]
```
- If the rule doesn’t have a label but has a location attr
```
rule: 98 2 [] [/home/robertorosmaninho/rv/temp/test-proof-trace.k:8]
  Var'Unds'DotVar0 = kore[15]
  Var'Unds'DotVar1 = kore[4]
 ```
 
 We decided to create only one test for this feature, as this is only for debugging, and duplicating every single file test would cause massive and tedious extra work to review not only this simple PR but any other PR that bumps the Hints version.

### CODE REVIEW

**Review by dwightguth — APPROVED** [review]

**Comment by Robertorosmaninho** [comment]
Example of use:
```Bash
kore-proof-trace add-rewrite.header input.add-rewrite.hints definition.kore --expand-terms --verbose
```

PR #1100, opened 2024-07-04
.gitmodules: Change pybind11 URL from relative to absolute

0cjs:
All the other URLs are absolute, so logically this one should match. Additionally, the relative URL will break if someone's main remote is not on GitHub.


0cjs:
If you need to contact me about this, the best way to do so is in the Pi Squared slack, or at curt.sampson@pi2.network.

PR #1105, opened 2024-07-09
 Make some code easier to inline (rebased)

dwightguth:
The llvm backend was not being particularly intelligent about the ability to inline certain function calls. Here we move all the `.ll` files in the `runtime` directory into the module used to initialize code generation. This allows these calls to live in one translation unit and be better optimized, while also simplifying llvm-kompile-clang.


lib/codegen/Decision.cpp
```diff
@@ -791,6 +791,7 @@ void make_eval_or_anywhere_function(
       = llvm::FunctionType::get(return_type, args, false);
   std::string name = fmt::format("eval_{}", ast_to_string(*function, 0, false));
   llvm::Function *match_func = get_or_insert_function(module, name, func_type);
+  match_func->deleteBody();
```
Robertorosmaninho, line 794: Why are we setting this here?

lib/codegen/CreateTerm.cpp
```diff
@@ -27,94 +27,9 @@ using namespace fmt::literals;
...
-; %layoutN = type { %blockheader, [0 x i64 *], %map, %mpz *, %block * }
-
-; Interface to the configuration parser
-declare ptr @parse_configuration(ptr)
-declare void @print_configuration(ptr, ptr)
-)LLVM";
-  return target_dependent + rest;
+  return
+#include "config/llvm_header.h"
+      ;
```
Robertorosmaninho, line 32: I didn't know that this was possible! Awesome!

Robertorosmaninho (approved):
LGTM. Just interested in the answer for the the question in `Decision.cpp`. Feel free to merge after double-checking it!

PR #1107, opened 2024-07-10
Fixes for MInts

dwightguth:
There existed several bugs leading to crashes when the LLVM backend compiles certain definitions containing the MInt sort. This PR contains a regression test and the fixes needed to make that test pass, namely:

* Don't emit a case for `evaluate_function_symbol` if the function has a variable return sort (only happens with functions that have a sort variable in the return sort that does not appear in the argument list)
* Fix a case where a string literal did not get its initializer set in all cases.


Robertorosmaninho (approved):
LGTM

PR #1108, opened 2024-07-12
Shared memory ringbuffer based parser for  proof hints

theo25:
This PR adds a shared memory ringbuffer based subclass for the `proof_trace_buffer` abstract class that is used by the proof trace parser to abstract away the mechanism in which the proof trace is being read. The new subclass reads the proof trace from a ringbuffer located in shared memory. In the future, we plan to add a corresponding writer for the hint generator that will output the proof trace in the shared memory ringbuffer. We do this in order to reduce the overhead of the hint generation.

Some important notes/disclaimers for this PR:
- We use a typical producer/consumer synchronization scheme for accessing the ringbuffer: we use two POSIX semaphores to singal data available and space available. We assume only two processes, one reader and one writer, and therefore we do not need to lock the accesses to the buffer data.
- In order to test the new parser, we have added a new tool, namely `kore-proof-trace-shm-writer`, that reads a proof trace file (the only kind of output currently supported by the hint generator) and writes its contents to the shared memory ringbuffer. In the future, we will add support for the hint generator to opt to write the proof trace straight into the ringbuffer, and at that point the `kore-proof-trace-shm-writer` can be removed.
- The setup of the shared memory object and the initialization of the semaphores happens in `kore-proof-trace` when it is passed a new flag, namely `--shared-memory`. If the flag is passed then the `<input filename>` positional argument serves as the name of the shared memory object. The same name must be passed to `kore-proof-trace-shm-writer` in order for the IPC to work. Moreover, the `kore-proof-trace-shm-writer` invocation needs to allow some time for `kore-proof-trace` to finish the setup. That is why we use `sleep 1` in CI between the two invocations.
- Currently, the reader and writer processes synchronize on reads/writes of 1 byte. This is not very efficient and in a future PR we plan to add support for buffered reads/writes of larger sizes. Note that the ringbuffer API already can handle larger reads/writes.

The structure of the PR content is as follows:
- 2 new files, `include/kllvm/binary/ringbuffer.h` and `lib/binary/ringbuffer.cpp`, contain the API and implementation for the ringbuffer data structure.
- a new subclass of `proof_trace_buffer`, namely `proof_trace_ringbuffer` is added to `include/kllvm/binary/deserializer.h`. This subclass implements the abstract API for various types of reads and peeks, all based on two basic read and peek functions that contain the synchronization logic on the reader's side.
- a new mode is added to the `tools/kore-proof-trace/main.cpp` program, that does the setup of the shared memory object and the initialization of the semaphores and invokes the shared memory/ringbuffer based hint parser.
- a new tool is added in `tool/kore-proof-trace-shm-writer` that reads a binary hint file and writes its contents to a provided shared memory object. The tool contains the synchronization logic on the writer's side.
- additional CI logic is added to `test/lit.cg.py` to allow parsing the hint file with the new shared memory/ringbuffer based parser for all tests that involve hint generation.


Robertorosmaninho (approved):
The PR, so far LGTM! I couldn’t spot why it’s failing in macOS, though, so I think this is the last thing needed to merge this PR.

I would like to see if it’s possible to have a more C++ style for the ringbuffer definition in the future. We should also have some Python bindings to expose this to MPG. The code was very well inline-documented, and the PR description helped a lot during the review.


Thanks for your hard work on this!

gtrepta (approved):
LGTM, and works on my machine.

include/kllvm/binary/ringbuffer.h
```diff
@@ -0,0 +1,62 @@
...
+// function does not point to a contiguous memory chunk of the corresponding
+// size.
+void ringbuffer_put(
+    shm_ringbuffer_t &buf, uint8_t const *data, size_t count = 1);
+
+// Get and remove data from the ringbuffer. The behavior is undefined if more
+// data is requested than it is currently available in the ringbuffer. The
+// behavior is also undefined if the data pointer passed to this function does
+// not point to a contiguous memory chunk of the corresponding size.
+void ringbuffer_get(shm_ringbuffer_t &buf, uint8_t *data, size_t count = 1);
```
Baltoli, line 58: ~~I don't totally follow why this code has been written like this: it's not (and can't be) C-linkage as it uses namespaces and references, but doesn't take advantage of any other C++ features?~~

Ah, I see that it's so you can directly mmap the entire thing into memory - I think what you could do here is make this a C++ class with constructors, methods etc., then use [placement new](https://en.cppreference.com/w/cpp/language/new) to construct a ringbuffer object into your shared memory block.

test/lit.cfg.py
```diff
@@ -223,13 +245,15 @@ def one_line(s):
     ('%test-dir-in', os.path.join('%input-dir', '%test-basename')),
     ('%test-proof-diff-out', os.path.join('%output-dir', '%test-basename.proof.out.diff')),
     ('%test-proof-debug-diff-out', os.path.join('%output-dir', '%test-basename.proof.debug.out.diff')),
+    ('%test-shm-buffer', os.path.join('/', '%test-basename.shm.buffer')),
```
Baltoli, line 248: Why is this path defined to start with `/`?

include/kllvm/binary/ringbuffer.h
Baltoli: It would also be nice to have some unit tests for the internal buffer data structure.

Baltoli (commented):
I left a couple of suggestions for the internal data structure that would let you improve the code quality a little, but feel free to stick with your current implementation if you'd rather.

test/lit.cfg.py
theo25: It is a POSIX requirement that names of shared memory objects start with `/`.

theo25:
I have managed to fix the issues with MacOS. There were 2 main issues:
- As far as I can tell, MacOS does not support combining `O_TRUNC` with `O_CREAT | O_EXCL` in `shm_open`.
- More importantly, the unnamed POSIX semaphores are not supported in MacOS. For that reason, I am using named semaphores. Another complication is that the names used for named semaphores in MacOS cannot exceed 32 characters in length. To accommodate that, and still have different names for semaphores of different tests (required since tests can run in parallel), I had to make some brevity changes in the naming convection for these semaphores.

test/lit.cfg.py
Baltoli: Interesting, didn't know that!

Robertorosmaninho:
If you're confident that this PR is ready to merge, go ahead and merge it. We can deal with the C++ style later, just create an issue for that. That's low priority atm.

Robertorosmaninho:
@theo25, if you're confident that this PR is ready to merge, go ahead and merge it. We can deal with the C++ style later, just create an issue for that. That's a low priority atm.

include/kllvm/binary/ringbuffer.h
theo25: I have made the change from struct to C++ class and added a couple of unit tests for the ringbuffer. I will request another review as soon as the CI passes.

Robertorosmaninho (approved):
Thanks for addressing our comments. LGTM!

include/kllvm/binary/deserializer.h
```diff
@@ -203,6 +205,146 @@ class proof_trace_file_buffer : public proof_trace_buffer {
...
+
+    return true;
+  }
+
+public:
+  proof_trace_ringbuffer(
+      shm_ringbuffer *buffer, sem_t *data_avail, sem_t *space_avail)
+      : shm_buffer_(buffer)
+      , data_avail_(data_avail)
+      , space_avail_(space_avail) { }
```
Baltoli, line 283: One more minor nitpick: the destructor of the `shm_ringbuffer` object should be called explicitly somewhere. Currently this isn't an obvious issue because there are no resources being managed by the buffer (just plain member variables), but if this changes in the future you'll have a subtle issue from the missing destructor call.

The way I'd suggest organising this is to have `proof_trace_ringbuffer` take a `void *` argument and perform the placement new in its constructor, then call the destructor of the created buffer in its destructor. An example of what that might look like: https://godbolt.org/z/dfb5cTYns

Otherwise looks good; feel free to merge once this and my other minor nits are resolved.

include/kllvm/binary/deserializer.h
```diff
@@ -203,6 +205,146 @@ class proof_trace_file_buffer : public proof_trace_buffer {
...
+
+  bool read(void *ptr, size_t len) override {
+    auto *data = (uint8_t *)ptr;
```
Baltoli, line 286:
```suggestion
    auto *data = static_cast<uint8_t *>ptr;
```

include/kllvm/binary/deserializer.h
```diff
@@ -203,6 +205,146 @@ class proof_trace_file_buffer : public proof_trace_buffer {
...
+    uint8_t c = 0;
+    if (read(&c)) {
+      return c;
+    }
+    return -1;
+  }
+
+  bool has_word() override {
+    uint64_t word = 0;
+    auto *data = (uint8_t *)&word;
```
Baltoli, line 300:
```suggestion
    auto *data = reinterpret_cast<uint8_t *>(&word);
```

include/kllvm/binary/deserializer.h
```diff
@@ -203,6 +205,146 @@ class proof_trace_file_buffer : public proof_trace_buffer {
...
+    uint8_t c = 0;
+    if (peek(&c)) {
+      return c;
+    }
+    return -1;
+  }
+
+  uint64_t peek_word() override {
+    uint64_t word = 0;
+    auto *data = (uint8_t *)&word;
```
Baltoli, line 316:
```suggestion
    auto *data = reinterpret_cast<uint8_t *>(&word);
```

tools/kore-proof-trace-shm-writer/main.cpp
```diff
@@ -0,0 +1,92 @@
...
+  // Copy the trace file into the shared memory object
+  std::ifstream file(input_filename, std::ios_base::binary);
+  while (true) {
+    char c = 0;
+    file.read(&c, 1);
+    if (file.eof()) {
+      break;
+    }
+    sem_wait(space_avail);
+    shm_buffer->put((uint8_t *)&c);
```
Baltoli, line 77:
```suggestion
    shm_buffer->put(reinterpret_cast<uint8_t *>(&c));
```

Baltoli (approved):

include/kllvm/binary/deserializer.h
theo25: I have addressed this and merged all your suggestions. I will merge the PR when CI passes.

PR #1109, opened 2024-07-16
Drop flat namespace option

Baltoli:
This PR fixes two subtle, related issues that are blocking updates from going through downstream in the Kasmer project. At a high level, the issues are:
- Flat namespace linking on macOS produces incorrect symbol lookups in dynamic libraries.
- https://github.com/runtimeverification/llvm-backend/pull/1097 misses a subtle edge case related to tail-call optimisation.

The actual code changes required are small, but warrant some detailed explanation.

## Flat Namespaces

For a long time, macOS has implemented a system known as _two-level_ namespaces, whereby undefined symbol names in a dynamic library are prefixed with the name of the library in which the loader expects to be able to find them at run-time. This is a conservative behaviour; even if a symbol with the same name exists in a different library, it won't be selected. For example, the dynamic libraries built by `llvm-kompile` in `c` mode link against `libgmp`. Two-level namespaces produce dynamic symbol tables that look like:
```console
$ dyld_info test/c/Output/flat-namespace.kore.tmp.dir/libtest.so -symbolic_fixups | grep gmpz_clear
           +0x2B28      bind pointer   libgmp.10.dylib/___gmpz_clear
```

This behaviour is different to Linux, which does not have a notion of two-level namespaces. For legacy compatibility purposes, Apple supply a linker flag `-flat_namespace` that behaves more similarly to Linux behaviour. Its use is discouraged in new code, but we had enabled it to work around an issue in the Python bindings (https://github.com/python/cpython/issues/97524) that should be fixed in a future CPython / macOS combination.[^1] When enabled, the symbol table looks something like this for the same example:
```console
$ dyld_info test/c/Output/flat-namespace.kore.tmp.dir/libtest.so -symbolic_fixups | grep gmpz_clear
           +0x2EE8      bind pointer   flat-namespace/___gmpz_clear
```

As a consequence of this, if the symbol `___gmpz_clear` exists in multiple dynamic libraries loaded by the same process, then the order in which they will be selected by the dynamic loader is not clearly well-defined,[^2] and when it's referenced we could end up loading either the correct or the incorrect symbol. This caused the initial bug observed as follows:[^3]
- The Haskell backend statically links the `kore-rpc-booster` executable against `libgmp`, meaning that some GMP symbols appear in that binary.
- The backend compiles shared libraries that dynamically link against `libgmp`.
- `kore-rpc-booster` dynamically loads one of these libraries, and when resolving symbols to load, the flat namespace environment selects the static version for some and the dynamic version for others.
- A call to `__gmpz_clear` from a backend hook ends up referencing the statically linked symbol, rather than the dynamically linked version. Generally, I think this situation is harmless - GMP is very stable and it's plausible that doing this for most symbols is not observable.
- However, the dynamically-linked GMP library has been set up to use the KORE memory management functions. When the static version is called, it tries to `free()` a pointer allocated by the backend's GC, and crashes.

The fix for this issue is to drop our usage of `-flat_namespace` for C shared libraries compiled by the backend. This breaks a few places we were relying on the old (incorrect) behaviour in the presence of C++ RTTI; having multiple instances of identically-named typeinfo symbols in a process is known to be broken there:
- `libunwind` is actually implicitly linked via the macOS system library; if we explicitly link it as well, then code that handles exceptions will break.
- The `k-rule-apply` tool linked two copies of the KORE AST library, causing `dynamic_cast` to break. #1110 addresses this.

## Tail-Call Optimisation

In #1097, we made some changes that explicitly mark K functions as `musttail` when we know they're tail recursive. In doing so, we removed the need to use the `-tailcallopt` flag in most cases. However, the change in that PR missed that as well as IR-level transformations, `-tailcallopt` sets a lower-level flag in the backend[^4] code generator that guarantees tail-call code generation. For large programs, this meant I could observe stack overflows when traversing large terms.

The fix is just to enforce that this internal option gets set properly; doing so is just a restoration of the behaviour we got from `-tailcallopt` before.

[^1]: But isn't yet fixed, unfortunately - the underlying bug is still present on my system. Should be revisited in the future, ideally!
[^2]: It might be defined somewhere, but the initial manifestation of this bug appeared in an apparently unrelated commit, so I think we were just getting lucky previously. The fix in this PR is morally correct whether or not things worked accidentally beforehand.
[^3]: I intend to write this up fully later in a separate issue.
[^4]: As in the X86 or arm backend of LLVM itself.

### DIFF
--- a/bin/llvm-kompile-clang
+++ b/bin/llvm-kompile-clang
@@ -136,7 +136,6 @@ if [[ "$OSTYPE" == "darwin"* ]]; then
   flags=(
     "-L@BREW_PREFIX@/opt/libffi/lib"
     "-L@BREW_PREFIX@/lib"
-    "-L@LLVM_LIBRARY_DIR@"
     "-Wl,-u,_table_getArgumentSortsForTag"
     "-I" "@BREW_PREFIX@/include"
   )
@@ -156,6 +155,19 @@ if [[ "$OSTYPE" == "darwin"* ]]; then
   set_visibility_hidden="$LIBDIR/libSetVisibilityHidden.dylib"
 else
   set_visibility_hidden="$LIBDIR/libSetVisibilityHidden.so"
+fi
+
+# On macOS, we get libunwind supplied as part of the developer tools in the OS,
+# and so don't need to link it directly. If we instead try to explictly link
+# against the libunwind that's part of Homebrew-supplied LLVM, it's easy to end
+# up in a situation where exceptions thrown in a shared library are not
+# compatible with the unwinding machinery in the main application binary. This
+# then manifests as BAD_ACCESS errors (or similar) when an exception is thrown,
+# _even if the exception should in principle be caught_.
+if [[ "$OSTYPE" == "darwin"* ]]; then
+  libunwind=""
+else
+  libunwind="-lunwind"
 fi
 
 # When building the Python AST module, there's no runtime and no main file, so
@@ -218,7 +230,7 @@ if [ "$main" = "static" ]; then
 elif [[ "$main" =~ "python" ]]; then
   # Don't link jemalloc when building a python library; it clashes with the
   # pymalloc implementation that Python expects you to use.
-  all_libraries=("${libraries[@]}" "-lgmp" "-lgmpxx" "-lmpfr" "-lpthread" "-ldl" "-lffi" "-lunwind")
+  all_libraries=("${libraries[@]}" "-lgmp" "-lgmpxx" "-lmpfr" "-lpthread" "-ldl" "-lffi" "$libunwind")
   flags+=("-fPIC" "-shared" "-I${INCDIR}" "-fvisibility=hidden")
 
   read -r -a python_include_flags <<< "$("${python_cmd}" -m pybind11 --includes)"
@@ -240,11 +252,11 @@ elif [ "$main" = "c" ]; then
 
   # Avoid jemalloc for similar reasons as Python; we don't know who is loading
   # this library so don't want to impose it.
-  all_libraries=("${libraries[@]}" "-lgmp" "-lgmpxx" "-lmpfr" "-lpthread" "-ldl" "-lffi" "-lunwind")
+  all_libraries=("${libraries[@]}" "-lgmp" "-lgmpxx" "-lmpfr" "-lpthread" "-ldl" "-lffi" "$libunwind")
   flags+=("-fPIC" "-shared" "$start_whole_archive" "$LIBDIR/libkllvmcruntime.a" "$end_whole_archive")
   clangpp_args+=("-o" "${output_file}")
 else
-  all_libraries=("${libraries[@]}" "-lgmp" "-lgmpxx" "-lmpfr" "-lpthread" "-ldl" "-lffi" "-ljemalloc" "-lunwind")
+  all_libraries=("${libraries[@]}" "-lgmp" "-lgmpxx" "-lmpfr" "-lpthread" "-ldl" "-lffi" "-ljemalloc" "$libunwind")
 fi
 
 if $link; then
--- a/bin/llvm-kompile-clang
+++ b/bin/llvm-kompile-clang
@@ -233,6 +233,10 @@ elif [[ "$main" =~ "python" ]]; then
   all_libraries=("${libraries[@]}" "-lgmp" "-lgmpxx" "-lmpfr" "-lpthread" "-ldl" "-lffi" "$libunwind")
   flags+=("-fPIC" "-shared" "-I${INCDIR}" "-fvisibility=hidden")
 
+  if [[ "$OSTYPE" == "darwin"* ]]; then
+    flags+=("-Wl,-flat_namespace")
+  fi
+
   read -r -a python_include_flags <<< "$("${python_cmd}" -m pybind11 --includes)"
   flags+=("${python_include_flags[@]}")
 
--- a/bin/llvm-kompile-clang
+++ b/bin/llvm-kompile-clang
@@ -207,7 +207,7 @@ if [[ "$OSTYPE" == "darwin"* ]]; then
   start_whole_archive="-force_load"
   end_whole_archive=""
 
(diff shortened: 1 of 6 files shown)


Baltoli:
https://github.com/python/cpython/issues/97524#issuecomment-1458855301

bin/llvm-kompile-clang
```diff
@@ -218,9 +230,13 @@ if [ "$main" = "static" ]; then
...
   # Don't link jemalloc when building a python library; it clashes with the
   # pymalloc implementation that Python expects you to use.
-  all_libraries=("${libraries[@]}" "-lgmp" "-lgmpxx" "-lmpfr" "-lpthread" "-ldl" "-lffi" "-lunwind")
+  all_libraries=("${libraries[@]}" "-lgmp" "-lgmpxx" "-lmpfr" "-lpthread" "-ldl" "-lffi" "$libunwind")
   flags+=("-fPIC" "-shared" "-I${INCDIR}" "-fvisibility=hidden")
 
+  if [[ "$OSTYPE" == "darwin"* ]]; then
+    flags+=("-Wl,-flat_namespace")
+  fi
+
```
Robertorosmaninho, line 239: I'm not sure if I correctly understood what you are doing here:
> The fix for this issue is to drop our usage of `-flat_namespace` for C shared libraries compiled by the backend.

Aren't we doing the opposite here?

lib/codegen/ApplyPasses.cpp
```diff
@@ -123,6 +123,7 @@ void generate_object_file(llvm::Module &mod, llvm::raw_ostream &os) {
 
   auto features_string = features.getString();
   auto options = TargetOptions{};
+  options.GuaranteedTailCallOpt = true;
```
Robertorosmaninho, line 126: Isn't there a way to extend Dwight's PR instead or reenable this optimization?

nix/llvm-backend.nix
```diff
@@ -40,7 +40,7 @@ stdenv.mkDerivation {
       --replace '"-liconv"' '"-L${libiconv}/lib" "-liconv"' \
       --replace '"-lncurses"' '"-L${ncurses}/lib" "-lncurses"' \
       --replace '"-ltinfo"' '"-L${ncurses}/lib" "-ltinfo"' \
-      --replace '"-lunwind"' '"-L${libunwind}/lib" "-lunwind"' \
+      --replace '"$libunwind"' '"-L${libunwind}/lib" "-lunwind"' \
```
Robertorosmaninho, line 43: How can nix get the value of `$libunwind` here?

test/c/Inputs/flat-namespace.c
Robertorosmaninho, line 1: I'm not sure if I understood what this testing is doing, can you please elaborate on this?

Robertorosmaninho (commented):
Thank you for the careful explanation of the PR, Bruce! I still have some questions about it to check if I understand it correctly. I’ll do a second review when I fully understand them, alright?

bin/llvm-kompile-clang
Baltoli: Previously, the `-flat_namespace` option was enabled everywhere on macOS (C, Python, interpreter binaries). This change pushes the option inwards so it's only enabled for Python libraries where it's legitimately still needed.

lib/codegen/ApplyPasses.cpp
Baltoli: I'm not sure I follow what you mean - this change is doing precisely that; it's extending Dwight's PR to re-enable TCO.

nix/llvm-backend.nix
Baltoli: It's not a variable reference here; it's the literal string `$libunwind`. Note the single quotes!

test/c/Inputs/flat-namespace.c
Baltoli: It's a minimal reproduction of the initial failure; if you compile and run before disabling the flat namespace (i.e. check out the commit that adds the test) on macOS, this test will fail.

It fails because the `Int2Bytes` hook calls (indirectly) `__gmpz_clear` from libgmp; if you get the flat namespace mixing up symbols, then that hook will call the function defined here in the test (and therefore aborting). I'll add a comment clarifying this.

bin/llvm-kompile-clang
```diff
@@ -207,7 +219,7 @@ if [[ "$OSTYPE" == "darwin"* ]]; then
   start_whole_archive="-force_load"
   end_whole_archive=""
 
-  flags+=("-Wl,-flat_namespace" "-Wl,-undefined" "-Wl,dynamic_lookup")
```
Baltoli, line 210: Globally-enabled `-flat_namespace` is removed here @Robertorosmaninho

lib/codegen/ApplyPasses.cpp
Baltoli: The point is that `-tailcallopt` isn't just an IR-level feature; it hooks into lower-level parts of the code generator. You can see this in the options for `opt`: `--tailcallopt` is a top-level option rather than a pass that gets run by a pass manager.

bin/llvm-kompile-clang
Robertorosmaninho: Perfect! Thank you for the detailed explanation, Bruce!

Robertorosmaninho (approved):
LGTM! However, I would like you to have at least a second review before merging this. Thanks!

gtrepta (approved):
I'm approving, trusting that this fixes the issue. But I have a question

> The Haskell backend statically links the kore-rpc-booster executable against libgmp, meaning that some GMP symbols appear in that binary.

Why does the haskell backend statically link libgmp? Would an alternative solution be to change this?

Baltoli:
Good catch Guy, I should have expanded further on this in the original writeup.

> Why does the haskell backend statically link libgmp? Would an alternative solution be to change this?

It's woven pretty deep into the implementation of big integers in GHC. @goodlyrottenapple made a valiant attempt to remove it but it's sadly not a viable change! I also realise a very old change I made a long time ago was an equivalent band-aid: https://github.com/runtimeverification/llvm-backend/pull/568

PR #1110, opened 2024-07-18
Refactor `k-rule-apply` tool

Baltoli:
This PR is a preparatory refactoring that will enable me to fix a problematic bug in the Kasmer project; I've factored out the changes here because they include other related cleanups not directly relevant to the bug I'm fixing in the subsequent PR.

Currently, the `k-rule-apply` tool is explicitly linked against the KORE AST library in order for it to be able to parse a KORE definition and search for an axiom with a particular label. This works fine, but on changing the way that the LLVM backend is linked on macOS, this AST linkage caused a subtle bug to manifest because of the way that `k-rule-apply` uses `dlopen` to inspect a shared library compiled by the backend.[^1]

The fix here is to remove the explicit dependency of this tool on the AST + Parser libraries, and instead push all the required behaviour of the tool into the C bindings so that it can be used fully with `dlopen`.

The changes in this PR are as follows:
- Reorganise the existing code in the `k-rule-apply` tool to be more hygienic in its usage of header files etc.
- Push calls to the KORE parser library and references to the KORE AST structures out in favour of C-linkage calls to the bindings module, which is available and compiled into the same shared library as we're already `dlopen`ing.
- Extract the one non-trivial usage of the AST library (mapping rule labels to matching function names via their axiom) into the core C bindings.
- Remove the dependency of this tool on the AST and Parser libraries, finally enabling the downstream bugfix that motivated this change in the first place.

[^1]: I have a full writeup of this on the way; in the context of this PR the specifics aren't important.

### DIFF
--- a/bindings/c/include/kllvm-c/kllvm-c.h
+++ b/bindings/c/include/kllvm-c/kllvm-c.h
@@ -162,6 +162,18 @@ void kllvm_free_all_memory(void);
 
 bool kllvm_mutable_bytes_enabled(void);
 
+/* Definitions */
+
+/**
+ * Parse the given KORE definition, then if any of its axioms have a `label`
+ * attribute that matches the supplied label, return the name of the function
+ * symbol that attempts matching a pattern against that axiom (and will
+ * therefore populate the backend's global matching log).
+ *
+ * If no such axiom exists, return `nullptr`.
+ */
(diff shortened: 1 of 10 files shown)


gtrepta (approved):

Robertorosmaninho (approved):
LGTM! Thanks for the clean up! The code look much better now!

Issue #1115, opened 2024-07-23
Suboptimal type for kore_definition_syntax?


0cjs:
In `binds/c/lib.cpp`, we have the following definitions:

    /*
     * These symbols may not have been compiled into the library (if
     * `--embed-kprint` was not passed), and so we need to give them a weak default
     * definition. If the embed flag was passed, the value of these symbols will be
     * given by the embedded data.
     */
     char kore_definition_syntax __attribute__((weak)) = -1;
     char kore_definition_macros __attribute__((weak)) = -1;

Note that there's no indication of whether these are signed or unsigned.

However, later in the file we have:

    if (kore_definition_macros_len == -1 || kore_definition_macros == -1
        || kore_definition_syntax_len == -1 || kore_definition_syntax == -1) {

These comparisons with -1 of course work only with _signed_ chars, but on Linux `aarch64` and Darwin `arm64` platforms the default type of `char` is unsigned for clang (and perhaps the ABI). This produces (fortunately) an
error due to the `-Wall`.

There appear to be no warnings we can turn on that will indicate that such non-portable code is being used. (We get a `tautological-constant-out-of-range-compare` only when compiling on `aarch64`/`arm64` platforms.)

It looks as if this type is one we define internally, so perhaps [this document][int] might provide some guidance on a better type to choose?

(BTW, this makes the program uncompilable on Apple Silicon CPUs for both MacOS and Linux.)

[int]: https://en.cppreference.com/w/cpp/types/integer

0cjs:
Appears to be fixed by #1116.

PR #1116, opened 2024-07-24
Fix tautological comparison error

Baltoli:
Using `-1` as the sentinel "not present" value for these weak symbols of type `char` is not correct, as the standard makes no guarantee whether `char` is signed or not; it is therefore possible for an implementation to supply us with a type such that comparing the sentinels with `-1` later is tautologically false.

### DIFF
--- a/bindings/c/lib.cpp
+++ b/bindings/c/lib.cpp
@@ -75,15 +75,15 @@ bool hook_BYTES_mutableBytesEnabled();
 extern "C" {
 
 /*
- * These symbols may not have been compiled into the library (if
- * `--embed-kprint` was not passed), and so we need to give them a weak default
+ * These symbols may not have been compiled into the library (if `--embed FILE`
+ * was not passed appropriately), and so we need to give them a weak default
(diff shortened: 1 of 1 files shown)


Baltoli:
cc @0cjs @stevenmeker

0cjs:
This is a fix for #1115, right?

Baltoli:
Yep, I believe it morally should be, though I'm still not sure how to reproduce the configuration that causes the error.

Robertorosmaninho (approved):
LGTM! Thanks for addressing this!

0cjs:
> ...though I'm still not sure how to reproduce the configuration that causes the error.

Ah, my bad! I should have given you that info in the issue. We reproduced it only on Debian 12 running on `aarch64` (a Docker container hosted on an Apple Silicon Mac). This _should_ have been the standard clang package from Debian, as I don't see any way another clang could have snuck in there.

We spent quite some time trying to figure out if we could find a compiler warning flag that would warn on systems with signed ints, but couldn't find anything.

PR #1118, opened 2024-07-25
Abstract proof hints writer

theo25:
This PR refactors the process of outputting proof hints to file so that it is handled by an abstract `proof_trace_writer`. We then implement a subclass of the `proof_trace_writer` class, namely `proof_trace_file_writer`, that outputs the proof hints to a binary file.

We do that so that we can easily add other methods of outputting proof hints in the future, such as a shared memory / ringbuffer based writer that will work in tandem with the shared memory proof hints parser that was recently merged.

The PR can be easily reviewed commit by commit.
- the first 3 commits do some preparatory refactoring to normalize naming conventions that will make the big refactoring more straightforward.
- the 4th commit replaces the `writer *` argument of proof trace output functions with a `FILE *` argument, since this is the only part of the `writer` class that is used when it comes to hint generation.
- the 5th commit refactors all `FILE *` pointers related to hint generation into pointers to a placeholder class which simply wraps said `FILE *`. It also does some function renaming to better reflect that these functions participate in the hint generation output process.
- the 6th commit replaces the placeholder class with the abstract `proof_trace_writer` class and adds a concrete subclass for writing into a file, namely `proof_trace_file_writer`.


theo25:
@Baltoli I don't know if you will have the time to review this one so I didn't ask for a review. If you do have the time, your comments are always welcome and appreciated!

gtrepta (approved):
LGTM

include/kllvm/binary/serializer.h
```diff
@@ -141,6 +141,34 @@ void serializer::emit(T val) {
 
 void emit_kore_rich_header(std::ostream &os, kore_definition *definition);
 
+class proof_trace_writer {
+public:
+  virtual ~proof_trace_writer() = default;
+  virtual void write(void const *ptr, size_t len) = 0;
+  virtual void write_string(char const *str, size_t len) = 0;
+  virtual void write_string(char const *str) = 0;
```
Baltoli, line 149: What's the intended difference between this overload and `write_c_string`?

include/runtime/header.h
```diff
@@ -336,19 +337,27 @@ void serialize_configurations(
...
+    kllvm::proof_trace_writer *proof_writer, block *subject, uint32_t sort);
+void serialize_configuration_to_proof_writer(
+    kllvm::proof_trace_writer *proof_writer, block *subject);
+void write_uint64_to_proof_trace(
+    kllvm::proof_trace_writer *proof_writer, uint64_t i);
+void write_bool_to_proof_trace(kllvm::proof_trace_writer *proof_writer, bool b);
+void write_string_to_proof_trace(
+    kllvm::proof_trace_writer *proof_writer, char const *str);
+void serialize_term_to_proof_trace(
+    kllvm::proof_trace_writer *proof_writer, void *subject, uint64_t, bool);
```
Baltoli, line 360: These functions are marked as `extern "C"` but now take a pointer to a C++ type as an argument. This is fine and won't break anything, but if you were ever to move the declarations to a context where they'd actually be compiled as C, it would break things.

My suggestion would be to make the writer arguments `void *` and have an explicit static cast back to the writer type internally to the C++ implementation. This keeps the C declaration / C++ implementation boundary a bit cleaner.

Baltoli (approved):
A couple of small comments but this looks really good to me!

include/kllvm/binary/serializer.h
theo25: `write_c_string` is supposed to add the `\0` at the end, while the `write_string` overloads do not add a `\0`. I understand that the naming is not ideal and I am open to suggestions.

Baltoli: `write_string` and `write_terminated_string` maybe? Also worth adding a comment, perhaps

include/runtime/header.h
theo25: I have addressed this in my latest commit.

Baltoli: Go ahead and merge in that case!

theo25: I want Dwight to review this before I merge because it touched a lot of his code. But thank you for taking the time and giving feedback!

include/kllvm/binary/serializer.h
theo25: I just pushed a commit where I have done the renaming and added comments as you recommended.

dwightguth (approved):
I didn't examine everything at a hugely high level of detail, but all the changes you make look reasonable at a high level and I didn't spot anything obviously wrong. I'm going to approve this, since others have already reviewed also.

PR #1121, opened 2024-07-29
Better inlining for alloc.cpp

dwightguth:
This PR improves the inlining for allocations considerably by ensuring that the bump allocator gets inlined into the code generator. We do this by means of moving `kore_arena_alloc` to the header file and making it an inline definition, as well as modifying the build process so that the bitcode file generated by alloc.cpp gets linked using LLVMs bitcode linker into the code generator module prior to optimization.


theo25 (approved):
Looks good to me

include/runtime/alloc_cpp.h
Robertorosmaninho, line 1: I would appreciate some inline documentation here with some explanation of why this is necessary.

lib/codegen/ApplyPasses.cpp
```diff
@@ -145,4 +149,18 @@ void generate_object_file(llvm::Module &mod, llvm::raw_ostream &os) {
   pm.run(mod);
 }
 
+void do_bitcode_linking(llvm::Module &mod) {
```
Robertorosmaninho, line 152: Same thing thing here. A small in-line comment explaining why we're doing that will be much appreciated!

Robertorosmaninho (approved):
LGTM. I've asked for a few more comments on the code, but that's not a blocking issue.

PR #1122, opened 2024-07-29
Shared memory writer for proof hints

theo25:
This PR adds a subclass of the abstract `proof_trace_writer` for outputting hints into a ringbuffer that lives in shared memory. This output method is intended to be used with the shared memory option of the `kore-proof-trace` parser to communicate hints through shared memory instead of through a file.

Since we now can test the shared memory based proof generation/consumption loop, we also do not include the helper `kore-proof-trace-shm-writer` tool in the installation.

### DIFF
--- a/runtime/util/finish_rewriting.cpp
+++ b/runtime/util/finish_rewriting.cpp
@@ -1,16 +1,21 @@
+#include <kllvm/binary/serializer.h>
+
 #include <runtime/header.h>
 
 #include <cstdint>
+#include <fcntl.h>
 #include <iostream>
 #include <memory>
+#include <sys/mman.h>
 
 extern "C" {
 
 FILE *output_file = nullptr;
-kllvm::proof_trace_writer *proof_writer = nullptr;
+void *proof_writer = nullptr;
 bool statistics = false;
 bool binary_output = false;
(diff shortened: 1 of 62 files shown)


runtime/util/finish_rewriting.cpp
```diff
@@ -1,24 +1,77 @@
...
+      shm_object, data_avail, space_avail);
+}
+
 void init_outputs(char const *output_filename) {
+  if (proof_output && use_shm) {
+    set_up_shm_writer(output_filename);
+    return;
+  }
+
   output_file = fopen(output_filename, "a");
```
Robertorosmaninho, line 75: Should we though about this append mode here? What are your thoughts @theo25 and @dwightguth?

Robertorosmaninho (approved):
LGTM, one minor question unrelated to the PR, but I would lake to take this chance to double check it.

dwightguth (approved):

runtime/util/finish_rewriting.cpp
theo25: @dwightguth Any comment on why this has to be an append flag? I kept it as an append because it was like so in the main.ll from where the file open was moved.

PR #1123, opened 2024-07-31
Refactoring remaining naming issues in `kgdb.py` and `klldb.py`

Robertorosmaninho:
This PR has a small fix similar to what we had for https://github.com/runtimeverification/llvm-backend/pull/1120.


gtrepta:
Does it need to be changed here as well? I can't tell if it's referencing the same thing:

https://github.com/runtimeverification/llvm-backend/blob/c840139d8bff300ce1d7535faa564da05ae0b984/debug/klldb.py#L173

Robertorosmaninho:
> Does it need to be changed here as well? I can't tell if it's referencing the same thing:
> 
> https://github.com/runtimeverification/llvm-backend/blob/c840139d8bff300ce1d7535faa564da05ae0b984/debug/klldb.py#L173

Yes, thanks for pointing that @gtrepta! The issue experienced by `QWYNGER#1188` on Discord is related to this PR, as it has the same naming issue.

gtrepta (approved):

PR #1125, opened 2024-08-01
merge allocations in same basic block up to a size bound

dwightguth:
Previously, when we were allocating a large term, we would issue a number of repeated calls to kore_alloc. This is inefficient because it has to check whether there is enough space and execute a branch instruction on each allocation. It is more efficient if we do one, larger call to kore_alloc for all the memory that we need.

In order to efficiently implement this, we modify the way we generate calls to kore_alloc in the code generator so that it will place the allocation function at the beginning of the basic block and then, if further allocations are required in the same basic block, it will modify the call to the allocation function so it requests a larger amount of memory. This only happens up to a certain bound, in order to prevent memory from being wasted due to requesting particularly large chunks of memory. In practice, this leads to most apply_rule functions only calling kore_alloc once.

### DIFF
--- a/include/kllvm/codegen/CreateTerm.h
+++ b/include/kllvm/codegen/CreateTerm.h
@@ -11,7 +11,7 @@
 #include "llvm/IR/Value.h"
 
 // size up to which calls to the same allocation function will be merged.
-enum { MAX_BLOCK_MERGE_SIZE = 4096 };
+enum { MaxBlockMergeSize = 4096 };
 
 namespace kllvm {
 
--- a/include/kllvm/codegen/CreateTerm.h
+++ b/include/kllvm/codegen/CreateTerm.h
@@ -1,12 +1,17 @@
 #ifndef CREATE_TERM_H
 #define CREATE_TERM_H
 
+#include <concepts>
+
 #include "kllvm/ast/AST.h"
 #include "kllvm/codegen/DecisionParser.h"
 
 #include "llvm/ADT/StringMap.h"
 #include "llvm/IR/Module.h"
 #include "llvm/IR/Value.h"
+
+// size up to which calls to the same allocation function will be merged.
+#define MAX_BLOCK_MERGE_SIZE 4096
 
 namespace kllvm {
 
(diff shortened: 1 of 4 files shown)


include/kllvm/codegen/CreateTerm.h
```diff
@@ -1,13 +1,18 @@
...
+
 #include "kllvm/ast/AST.h"
 #include "kllvm/codegen/DecisionParser.h"
 
 #include "llvm/ADT/StringMap.h"
 #include "llvm/IR/Module.h"
 #include "llvm/IR/Value.h"
 
+// size up to which calls to the same allocation function will be merged.
+enum { MaxBlockMergeSize = 4096 };
```
Robertorosmaninho, line 14: Perhaps it is a silly question, but is there any reason to use an enum instead of a global int var?

lib/codegen/CreateTerm.cpp
```diff
@@ -168,22 +168,56 @@ llvm::Value *get_block_header(
...
-
-llvm::Value *allocate_term(
-    llvm::Type *alloc_type, llvm::Value *len, llvm::BasicBlock *block,
+    llvm::Type *alloc_type, llvm::Value *len, T *insert_point,
     char const *alloc_fn) {
   auto *malloc = create_malloc(
-      block, len, kore_heap_alloc(alloc_fn, block->getModule()));
+      insert_point, len, kore_heap_alloc(alloc_fn, insert_point->getModule()));
 
   set_debug_loc(malloc);
```
Robertorosmaninho, line 180: Before, we were sure that we were always setting this debug location because the first function was always calling this one. However, that's not the current behavior, and in the new function, we don't have this `set_debug_loc` anywhere.

It would be good to have it here as well!

Robertorosmaninho (changes requested):
The PR LGTM! However, there's a missing debug info that we're not emitting in the new function that I would like to keep the old behavior for consistence.

include/kllvm/codegen/CreateTerm.h
dwightguth: Originally it was a macro, but clang tidy told me to use an enum, so I did. Not entirely sure the logic behind it. I probably could use a constexpr global too, if you think that's better. A regular global wouldn't necessarily get optimized as efficiently.

lib/codegen/CreateTerm.cpp
```diff
@@ -168,22 +168,58 @@ llvm::Value *get_block_header(
...
+            }
+          }
+        }
+      }
+    }
+    return allocate_term(
+        alloc_type, llvm::ConstantInt::get(ty, type_size), block, alloc_fn);
+  }
+  return allocate_term(
+      alloc_type, llvm::ConstantInt::get(ty, type_size), block, alloc_fn);
```
Robertorosmaninho, line 220: In these 2 cases is ok not to have debug info?

PR #1126, opened 2024-08-01
Updating `StringPattern` to return its contents as Python Bytes

Robertorosmaninho:


theo25 (approved):
Looks good to me

Robertorosmaninho:
@dwightguth I've tested it upstream with K. It breaks in some tests, which I already have a PR to fix (https://github.com/runtimeverification/k/pull/4567). With this fix I've tested KEVM as well, and everything is passing there!

PR #1127, opened 2024-08-05
Buffered reads/writes for hint generation with shared memory

theo25:
This PR adds implementation for buffered reads/writes from and to the shared memory ringbuffer during hint generation. The previous implementation would synchronize reads and writes at the granularity of one byte which is not ideal for performance. In this PR, we add support for synchronization across reads and writes of larger sizes. Both the ringbuffer-based proof trace reader and writer classes are updated to buffer incoming reads and writes: They collect data into an internal buffer and only interacting with the shared memory ringbuffer when that buffer is full in case of the writer or empty in case of the reader.

### DIFF
--- a/include/kllvm/binary/deserializer.h
+++ b/include/kllvm/binary/deserializer.h
@@ -210,68 +210,172 @@ private:
   shm_ringbuffer *shm_buffer_;
   sem_t *data_avail_;
   sem_t *space_avail_;
+
   std::deque<uint8_t> peek_data_;
+  std::array<uint8_t, shm_ringbuffer::buffered_access_sz> buffer_;
+  size_t buffer_data_size_{0};
+  size_t buffer_data_start_{0};
+
   bool peek_eof_{false};
+  bool buffered_eof_{false};
   bool read_eof_{false};
 
+  // Helper function that reads from the shared memory ringbuffer into the
+  // buffer_. It tries to read shm_ringbuffer::buffered_access_sz bytes of data.
+  // It assumes that buffer_ is empty.
+  void read_from_shared_memory() {
+    assert(buffer_data_size_ == 0);
+    assert(buffer_data_start_ == 0);
+
+    if (buffered_eof_) {
+      return;
+    }
+
+    sem_wait(data_avail_);
+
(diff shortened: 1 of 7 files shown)


include/kllvm/binary/deserializer.h
```diff
@@ -210,68 +210,172 @@ class proof_trace_ringbuffer : public proof_trace_buffer {
...
+    // Else, either EOF has not been written or EOF has been written but there
+    // are remaining full chunks to be read. In any case, we can read a full
+    // chunk.
+    shm_buffer_->get(buffer_.data());
+    sem_post(space_avail_);
+
+    buffer_data_size_ = shm_ringbuffer::buffered_access_sz;
+  }
+
+  bool read(uint8_t *ptr, size_t len = 1) {
```
Robertorosmaninho, line 261: Why we are still using len = 1 as default?

include/kllvm/binary/deserializer.h
```diff
@@ -210,68 +210,172 @@ class proof_trace_ringbuffer : public proof_trace_buffer {
...
       }
-      shm_buffer_->get(&ptr[i]);
-      sem_post(space_avail_);
     }
 
+    assert(len == 0);
     return true;
   }
 
   bool peek(uint8_t *ptr, size_t len = 1) {
```
Robertorosmaninho, line 320: Same question here

Robertorosmaninho (approved):
Besides my questions, it LGTM

include/kllvm/binary/deserializer.h
dwightguth: This is just the wrapper, not the actual read to/from shared memory.

dwightguth (approved):

include/kllvm/binary/deserializer.h
theo25: As Dwight said, these are not the functions that do the actual shared memory accesses. They have default values of 1 because they are commonly called with this value and that makes the code more readable.

include/kllvm/binary/deserializer.h
theo25: See reply on the other question.

PR #1128, opened 2024-08-05
Improvements to MINT hooks

dwightguth:
This PR contains a couple different small changes to improve support for MInt in the WASM semantics:

* Add implementation of MINT.round, MINT.sext, MINT.umin, MINT.umax, MINT.smin, and MINT.smax.
* Add support for equality between machine integers of size 32 and 64 bits.
* Add variant of hook_LIST_update which takes an MInt.


runtime/collections/kelemle.cpp
```diff
@@ -125,6 +125,25 @@ bool hook_KEQUAL_eq(block *arg1, block *arg2) {
           }
           break;
         }
+        case MINT_LAYOUT + 32: {
```
theo25, line 128: Could you remind me where is the code that constructs the MINT layouts following this convention about how to find the bit width?

theo25 (approved):
I have a question for my better understanding, but this looks good to me so feel free to merge.

runtime/collections/kelemle.cpp
dwightguth: I don't remember exactly where that code lives, but basically, when we have to encode a sort category as an integer, the bit width gets added to the enum value. I tested this and it works.

PR #1129, opened 2024-08-07
Update KElem names in kgdb

gtrepta:
https://github.com/runtimeverification/k/issues/4576

More fixes that were missed in #1123 and #1120


Robertorosmaninho (approved):
Thanks for taking care of this @gtrepta and sorry for missing them in previous PRs!

PR #1130, opened 2024-08-07
Disabling shared memory proof trace tests on macos

theo25:
This PR disables the shared memory proof trace tests on macOS until we resolve a flaky issue with lingering shared memory objects.

### DIFF
--- a/test/lit.cfg.py
+++ b/test/lit.cfg.py
@@ -1,5 +1,6 @@
 import glob
 import os
+import platform
 import subprocess
 
 from lit import formats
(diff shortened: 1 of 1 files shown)


test/lit.cfg.py
```diff
@@ -75,6 +76,12 @@ def one_line(s):
         .replace('then ;', 'then') \
         .replace("' ; '", r"'\\n'") \
 
+# We want to be able to disable some checks on macOS, this function allows us to
```
Robertorosmaninho, line 79: Small typo: "to do"

Robertorosmaninho (approved):
Fancy solution. LGTM.

test/lit.cfg.py
theo25: thanks! fixed in a new commit

gtrepta (approved):

PR #1131, opened 2024-08-08
Adding Time Hooks for measuring semantics performance

Robertorosmaninho:
This PR introduces `hook_timer_start` and `hook_timer_stop`, which can be used to measure the execution time of an instruction or an entire semantics by getting the current timestamp when starting the timer and then when stopping it, we get the current timestamp again, and save the difference into a default file `hooks_time.txt` in the current directory.


theo25 (approved):
this looks good but given that I have contributed to this code, don't merge until @dwightguth approves as well.

dwightguth (approved):

PR #1134, opened 2024-08-13
Introducing new `llvm_pattern_matching_failure_event` to Proof Events

Robertorosmaninho:
Fixes: https://github.com/Pi-Squared-Inc/pi2/issues/1722
This PR introduces a new function event called `llvm_pattern_matching_failure_event`, which is identified by the `pattern_matching_failure_sentinel`. This new event flags to the user that an attempt to apply a rule failed in pattern matching its constructor; therefore, no further simplification can be done at this point.


Robertorosmaninho:
We still need to bump the version from 11 to 12, but this will be terrible to review after modifying all the tests, so I'll bump it as soon as I have your approval in the main implementation.

dwightguth (approved):

docs/proof-trace.md
```diff
@@ -49,6 +49,9 @@ name              ::= string
 location          ::= string
 function          ::= WORD(0xDD) name location arg* WORD(0x11) // the arg list is ommited in normal mode
 
+function_name     ::= string
+pattern_matching_failure ::= WORD(0x44) function_name
```
theo25, line 53: This needs to appear in the right hand side of some production, most likely the `argument` and `event` productions.

theo25 (changes requested):
Looks good, but please take care of the one comment that I left.

theo25 (approved):

PR #1135, opened 2024-08-16
disable tail call on return values of MInt of >192 bits

dwightguth:
On X86 architecture, LLVM can only return integers of size up to 192 bits in registers. Return values of greater size are subject to "sret demotion", which returns the value in memory. Due to the sret implied by this, tail calls cannot be performed on functions that return these large integer types. Thus, we need to delete the MustTail attribute or code generation will crash.


theo25 (approved):
I am approving but mayI ask that you add the k file for the new test you added?

PR #1136, opened 2024-08-16
Abstract proof trace writer at event level

theo25:
This PR refactors the proof trace writer to be abstract at the event generation level, rather than the byte stream level. We do this in order to be able to use alternative ways of communicating the proof hints that do not require writing the trace to a file, such as triggering of callback invocations.

The PR can be reviewed commit by commit:
- the 1st commit removes an unused flag from main.ll
- the 2nd commit adds the new abstract `proof_trace_writer` and a subclass that writes in a file as before.
- the 3rd commit removes the old `proof_trace_writer` that operated at the level of writing byte streams.


dwightguth (commented):
I'm having trouble tracking all of these changes, but the high level signature of proof_trace_writer looks good, the tests pass, and you seem happy with it as a stepping stone towards the callbacks, so I don't have any comments. I will let Roberto approve, though. I want another set of eyes on it.

runtime/util/ConfigurationSerializer.cpp
```diff
@@ -665,6 +659,77 @@ void write_string_to_proof_trace(void *proof_writer, char const *str) {
       ->write_null_terminated_string(str);
 }
 
+void serialize_configuration_to_proof_trace(
+    FILE *file, block *subject, uint32_t sort) {
```
Robertorosmaninho, line 638: Is this `FILE *file` right?

Robertorosmaninho: Why not `void *proof_writer`?

runtime/util/finish_rewriting.cpp
```diff
@@ -25,7 +24,7 @@ int32_t get_exit_code(block *);
 void init_outputs(char const *output_filename) {
   output_file = fopen(output_filename, "a");
   if (proof_output) {
-    proof_writer = new kllvm::proof_trace_file_writer(output_file);
+    proof_writer = new proof_trace_file_writer(output_file);
```
Robertorosmaninho, line 27: Is there a reason for not using `kllvm` namespace anymore?

runtime/util/util.cpp
```diff
@@ -34,9 +35,8 @@ block *construct_raw_term(void *subject, char const *sort, bool raw_value) {
   return static_cast<block *>(construct_composite_pattern(tag, args));
 }
 
-void print_proof_hint_header(kllvm::proof_trace_writer *writer) {
+void print_proof_hint_header(proof_trace_writer *writer) {
   uint32_t version = 11;
```
Robertorosmaninho, line 39: You'll probably need to update this version if rebasing doesn't do it automatically for you.

Robertorosmaninho (commented):
I would like some clarifications before approving it. But overall, it looks good to me, and if all tests area passing we're good as well.

include/runtime/proof_trace_writer.h
```diff
@@ -0,0 +1,142 @@
...
+  // The passed string should be 0 terminated.
+  void write_string(char const *str) { fputs(str, file_); }
+
+  // Note: this method will write a 0 at the end of the string.
+  // The passed string should be 0 terminated.
+  void write_null_terminated_string(char const *str) {
+    write_string(str);
+    char n = 0;
+    write(&n, 1);
+  }
```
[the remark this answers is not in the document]
stevenmeker: It would be clearer to rename write() to something more informative; say write_bytes(). Otherwise it looks to much like the system call write().

include/kllvm/codegen/ProofEvent.h
```diff
@@ -44,52 +44,96 @@ class proof_event {
...
-   * term must be known.
+   * Emit an instruction that has no effect and will be removed by optimization
+   * passes.
+   *
+   * We need this workaround because some callsites will try to use
+   * llvm::Instruction::insertAfter on the back of the MergeBlock after a proof
+   * branch is created. If the MergeBlock has no instructions, this has resulted
+   * in a segfault when printing the IR. Adding an effective no-op prevents this.
+   */
+  llvm::BinaryOperator *emit_no_op(llvm::BasicBlock *insert_at_end);
```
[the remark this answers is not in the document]
theo25: So there is an issue with the way github presents the diff for this PR. If you notice this is not my code, it has been just deleted from a different place in the file and moved here verbatim. As such I have no comment on what should be done with it and whatever it is that we may decide, it shouldn't be part of this PR anyway.

include/runtime/proof_trace_writer.h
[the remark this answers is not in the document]
theo25: I think I will go with @stevenmeker suggestion here. Calling `write_string` which is documented to not add a null at the end with the null string does not seem very clear to me. I will use `\0` though instead of `0` as @Robertorosmaninho suggested.

runtime/util/ConfigurationSerializer.cpp
theo25: This method now just writes to a file. Before it needed a pointer to the abstract class but now the abstraction is not at the level where this function will be called. When this is called we know we want to write to a file so that is why I have changed back to accept a file pointer as it was before I introduced the lower level abstract writer.

runtime/util/finish_rewriting.cpp
theo25: The `proof_trace_file_writer` class is now part of the runtime library that does not use any namespaces so I followed that convention. The reason it is part of the runtime library and not the binary library as it was before is that its methods need to call other functions that are part of the runtime library and therefore there would be a cyclical dependency if it remained in the binary library.

runtime/util/util.cpp
theo25: Good catch! will do

include/runtime/proof_trace_writer.h
```diff
@@ -0,0 +1,142 @@
...
+      = 0;
+  virtual void
+  rewrite_event_post(void *config, uint64_t block_header, bool indirect)
+      = 0;
+  virtual void function_event_pre(char const *name, char const *location_stack)
+      = 0;
+  virtual void function_event_post() = 0;
+  virtual void side_condition_event_pre(uint64_t ordinal, uint64_t arity) = 0;
+  virtual void
+  side_condition_event_post(uint64_t ordinal, bool side_cond_result)
```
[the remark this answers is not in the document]
theo25: Done

include/kllvm/codegen/ProofEvent.h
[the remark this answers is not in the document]
Robertorosmaninho: That's pretty awkward indeed.

include/runtime/proof_trace_writer.h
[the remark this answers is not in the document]
Robertorosmaninho: Sure! I'm happy with it too!

runtime/util/ConfigurationSerializer.cpp
Robertorosmaninho: Perfect! Thanks for the explanation!

runtime/util/finish_rewriting.cpp
Robertorosmaninho: Make sense! Thanks!

Robertorosmaninho (approved):
LGTM! Thanks for addressing my comments and explaining to me your modifications!

PR #1138, opened 2024-08-22
Equality function on mint

dwightguth:
This is used by the pattern matcher when nonlinear matching occurs.


Robertorosmaninho (approved):
LGTM

theo25 (approved):

PR #1140, opened 2024-08-28
Un-nest proof hint events for arguments of functions/hooks

theo25:
The events that correspond to evaluation of arguments of functions and hooks are currently nested inside the corresponding function/hook event. There is no need for that and it creates various complications in the processing of the proof hint trace. For this reason, this PR un-nests those events so that they appear before their corresponding function/hook event.


Robertorosmaninho (approved):
LGTM.
Should we update the [proof-trace.md](https://github.com/runtimeverification/llvm-backend/blob/master/docs/proof-trace.md) with new info?

theo25:
@Robertorosmaninho I added a commit with the update and a simplification to the parser logic. Thanks for pointing that out!

Robertorosmaninho:
> @Robertorosmaninho I added a commit with the update and a simplification to the parser logic. Thanks for pointing that out!

Wow! That was a huge simplification! Thanks!

lib/codegen/CreateTerm.cpp
```diff
@@ -735,16 +819,16 @@ llvm::Value *create_term::create_function_call(
...
+    current_block_
+        = e.function_event_pre(current_block_, pattern, location_stack);
+    size_t i = 0;
     for (auto const &p : pattern->get_arguments()) {
       auto *sort = dynamic_cast<kore_composite_sort *>(p->get_sort().get());
-      proof_event e(definition_, module_);
-      current_block_ = e.argument(args[i], sort, true, current_block_);
-      i++;
+      current_block_ = e.argument(args[i], sort, false, current_block_);
     }
```
gtrepta, line 830: `i` isn't being incremented here.

gtrepta (changes requested):
LGTM, but I think I found a bug. Please fix it and then I can approve.

lib/codegen/CreateTerm.cpp
theo25: Good catch! Fixed

gtrepta (approved):

PR #1141, opened 2024-09-03
Introducing new arena and function that allows an object to be ignored by GC

Robertorosmaninho:


Robertorosmaninho:
A c++ unittest is under development, but this was already tested using a K definition for a future K PR.

Robertorosmaninho:
Closing as the performance gain isn't meaningful.

PR #1143, opened 2024-09-09
Add pattern matching support for list random access patterns

dwightguth:
This is not rigorously tested, nor is it likely to be in the near future because the code it was written to support ended up not being worth merging. However, it is broadly useful code that may find value in the future, so I would like to get it merged.

It also includes a few fixes to python scripts in the repo that had gone slightly stale.


Robertorosmaninho (approved):
I couldn’t spot anything obvious that would prevent me from approving this PR, but I’m not very familiar with this code either. I’ll let @theo25 give the final approval on that. 

I would like to see, if not some tests, at least some instructions on how we could execute it ourselves. It may be a helpful starting point for us in the future if we want to fully support it in K.

theo25 (approved):
Looks good to me!

PR #1144, opened 2024-09-11
make llvm backend do better dce and dae

dwightguth:
Previously, the llvm optimization passes weren't terribly effective at removing dead code and dead function arguments. We add some custom llvm passes that make this work a bit more effectively.

There are two passes added:

* RemoveDeadKFunctions deletes function call instructions which call side-effecting functions that are "effectively" pure, i.e., they exist only to construct pure terms and return them.
* MustTailDeadArgElimination is a minorly modified version of the deadargelim LLVM pass which fixes a bug preventing it from applying dead argument elimination to musttail tail calls.


lib/passes/PluginInfo.cpp
```diff
@@ -0,0 +1,42 @@
+#include <kllvm/codegen/MustTailDeadArgElimination.h>
+#include <kllvm/codegen/RemoveDeadKFunctions.h>
+#include <kllvm/codegen/SetVisibilityHidden.h>
+
+using namespace kllvm;
+
+/* New PM Registration */
+llvm::PassPluginLibraryInfo get_kllvm_plugin_info() {
+  return {
+      LLVM_PLUGIN_API_VERSION, "SetVisibilityHidden", LLVM_VERSION_STRING,
```
theo25, line 10: This likely needs to be changed to something more general than `"SetVisibilityHidden"`.

lib/passes/PluginInfo.cpp
```diff
@@ -0,0 +1,42 @@
...
+            });
+        pb.registerPipelineParsingCallback(
+            [](StringRef name, llvm::FunctionPassManager &pm,
+               ArrayRef<llvm::PassBuilder::PipelineElement>) {
+              if (name == "remove-dead-k-functions") {
+                pm.addPass(remove_dead_k_functions());
+                return true;
+              }
+              return false;
+            });
```
theo25, line 33: Could you explain the way these passes are being registered here? Also why is this different than what we do in `codegen/ApplyPasses.cpp`?

lib/passes/MustTailDeadArgElimination.cpp
```diff
@@ -0,0 +1,1137 @@
+//===- DeadArgumentElimination.cpp - Eliminate dead arguments -------------===//
```
theo25, line 1: This is a huge file that came from the LLVM code base. Could you give me some pointers to the code that you added so that I can focus my review on that?

theo25 (commented):
This looks fine but I have a few questions that would help me review some parts of the PR.

lib/passes/PluginInfo.cpp
dwightguth: This is how the passes get registered so they run when you ask for them using the --passes flag of opt. In ApplyPasses.cpp, they are being registered as part of the default pipeline. The latter is what happens during default compilation, the former is for debugging.

lib/passes/PluginInfo.cpp
dwightguth: Good catch, I will fix this first thing tomorrow.

lib/passes/MustTailDeadArgElimination.cpp
dwightguth: My advice is just to check out the file locally and do a diff with the version in llvm 16.0.6. I don't have another way of showing you the exact difference between these two versions. The actual operational difference is that tailcc musttail calls do not inhibit the optimization.

theo25: OK I did what you suggested and the changes seem reasonable to me.

theo25 (approved):

Robertorosmaninho (approved):
LGTM

PR #1145, opened 2024-09-11
Verify ir during CI

dwightguth:
I noticed to my dismay that a recent PR had introduced code that passed tests, but was generating LLVM IR that did not pass the verification process. This PR does two things:

1. Fixes the bug in the code generator that was leading to code that did not verify.
2. Introduces code to the regression test suite to test that every definition we compile generates IR that is valid.


Robertorosmaninho (approved):
LGTM

PR #1147, opened 2024-09-17
Split proof hint trace into multiple files

theo25:
This PR adds support for generating multiple hint trace files containing parts of the total trace. To this end, we add an optional `--proof-chunk-size N` flag to the `interpreter` binary. When the flag is passed and `N` is greater than `0`, then the trace is separated into multiple files as follows:
- an `<output filename>.pre_trace` file with the pre-trace events and the initial configuration.
- a number of `<output filename>.<idx>` files that each contain `N` top-level rewrite rule events along with their corresponding side condition events and simplification events. If the `interpreter` binary has been generated with the slow option, then each of these files also contains configuration events after each rewrite event. Note that the last file may contain less than `N` events.

By default `N` equals to `0` and in this case, the trace is generated exactly as before.

This PR also adds support to the `kore_proof_trace` tool to be able to parse partial trace files. To achieve this, we introduce two more special headers (one for the pre-trace file and one for the chunk files) that allow the parser to recognize what kind of file it should expect.


include/kllvm/codegen/ProofEvent.h
```diff
@@ -60,6 +70,19 @@ class proof_event {
...
   llvm::LoadInst *emit_get_proof_trace_writer(llvm::BasicBlock *insert_at_end);
 
+  /*
+   * Emit instructions to get the current value of the steps global variable,
+   * which counts the number of rewrite steps taken.
+   */
+  llvm::LoadInst *emit_get_steps(llvm::BasicBlock *insert_at_end);
+
+  /*
+   * Emit instructions to get the current value of the hproof_chunk_size global
```
Robertorosmaninho, line 80: Is this `h` before `proof_chunk_size` a typo?

Robertorosmaninho (approved):
Looks good to me, but I wonder if we can and should expose this to Python bindings as well! what do you think?

include/kllvm/codegen/ProofEvent.h
theo25: yes, fixed in latest commit

PR #1148, opened 2024-09-18
Exposing `llvm_pattern_matching_failure_event` to Python Bindings

Robertorosmaninho:
We missed this feature while implementing the new step event; we now need to add it so the Pi2 MPG Team can handle it on the Python side. A subsequent K PR will also be necessary to correctly expose this feature.


Robertorosmaninho:
Subsequent PR in K: https://github.com/runtimeverification/k/pull/4643

theo25 (approved):

PR #1149, opened 2024-09-18
Fix `--statistics` flag by calling `get_steps()` instead of trying to access the `steps` global variable

Robertorosmaninho:
The previous way of accessing the `steps` global variable was broken, generating a segmentation fault. By making this simple modification, the correct number of steps taken by the program execution can be returned as expected. A regression test to check the correct behavior of this feature was also added.


runtime/util/finish_rewriting.cpp
```diff
@@ -15,7 +15,7 @@ bool statistics = false;
 bool binary_output = false;
 bool proof_output = false;
 
-extern int64_t steps;
+int64_t get_steps();
```
theo25, line 18: Not sure why this was a signed integer before, but I think it should an unsigned integer, ie `uint64_t`

runtime/util/finish_rewriting.cpp
```diff
@@ -51,6 +51,7 @@ void init_outputs(char const *output_filename) {
   }
 
   if (statistics) {
+    int steps = get_steps();
```
theo25, line 54: this should definitely not be an `int` because that can easily cause overflow issues. I would go with `uin64_t` here as well

test/output/imp.out.diff
```diff
@@ -0,0 +1 @@
+Lbl'-LT-'generatedTop'-GT-'{}(Lbl'-LT-'T'-GT-'{}(Lbl'-LT-'k'-GT-'{}(dotk{}()),Lbl'-LT-'state'-GT-'{}(\left-assoc{}(Lbl'Unds'Map'Unds'{}(Lbl'UndsPipe'-'-GT-Unds'{}(inj{SortId{}, SortKItem{}}(\dv{SortId{}}("s")),inj{SortInt{}, SortKItem{}}(\dv{SortInt{}}("66"))),Lbl'UndsPipe'-'-GT-Unds'{}(inj{SortId{}, SortKItem{}}(\dv{SortId{}}("q")),inj{SortInt{}, SortKItem{}}(\dv{SortInt{}}("1"))),Lbl'UndsPipe'-'-GT-Unds'{}(inj{SortId{}, SortKItem{}}(\dv{SortId{}}("m")),inj{SortInt{}, SortKItem{}}(\dv{SortInt{}}("2"))),Lbl'UndsPipe'-'-GT-Unds'{}(inj{SortId{}, SortKItem{}}(\dv{SortId{}}("r")),inj{SortInt{}, SortKItem{}}(\dv{SortInt{}}("3"))),Lbl'UndsPipe'-'-GT-Unds'{}(inj{SortId{}, SortKItem{}}(\dv{SortId{}}("n")),inj{SortInt{}, SortKItem{}}(\dv{SortInt{}}("1"))))))),Lbl'-LT-'generatedCounter'-GT-'{}(\dv{SortInt{}}("0")))
```
theo25, line 1: do we need this file?

theo25 (approved):

test/output/imp.out.diff
Robertorosmaninho: Not really, just deleted it! Thanks!

dwightguth (approved):

PR #1150, opened 2024-09-20
No proof hint events for unit/elememt collection hooks

theo25:
The unit/element hooks for collections are not useful to the math proof team so we modify our instrumentation to not emit them at all.


dwightguth (approved):
Looks fine. Did you confirm that the events for simplifying the arguments of elements still get emitted?

Robertorosmaninho (approved):
LGTM. I would like us to update the documentation just to be sure that this modification is highlighted there as a feature, not as as bug.

theo25:
@dwightguth yes the events related to argument evaluation should be generated as always.

PR #1153, opened 2024-09-30
disable proof hint generation during call to get_exit_code

dwightguth:
Previously, the get_exit_code call would emit events into the proof trace. This is not actually part of the proof of execution though, so we should omit these events.


theo25 (approved):

Robertorosmaninho (commented):
LGTM

Robertorosmaninho (approved):

PR #1162, opened 2024-10-23
fix ordinals not being applied to return types of symbols

dwightguth:
This fixes a bug where the sort ordinals for formal parameters mentioned only in the return type of the symbol were not correct, leading to incorrect serialization of terms containing injections.


lib/ast/definition.cpp
```diff
@@ -286,40 +325,11 @@ void kore_definition::preprocess() {
...
-              all_sorts_.push_back(sort_to_process);
-            }
-
-            sort_to_process->set_ordinal(sorts[*sort_to_process]);
-          }
+          process_sort_ordinal(sort.get(), sorts, all_sorts_, next_sort);
+        }
+        if (symbol->get_sort()->is_concrete()) {
+          process_sort_ordinal(
+              symbol->get_sort().get(), sorts, all_sorts_, next_sort);
```
theo25, line 332: I noticed that I missed processing the return sort in my implementation of the work list. However checking back at the original code, I don't think this was there either. So could you explain why this is needed? Is it the case that it missing was a bug even before we added the work list implementation?

theo25 (approved):
I approved, but left one question for my better understanding.

Robertorosmaninho (approved):
LGTM. I've tested and it passes the corner cases it was failing before!

lib/ast/definition.cpp
dwightguth: It was a bug that was there before your changes were made, yes. The issue is that formal parameters that appear only in the return sort of the symbol (e.g. the second parameter of `inj`) were not getting their ordinal set and thus were getting serialized incorrectly.

PR #1163, opened 2024-11-14
try to delete some probably unneeded bitcasts

dwightguth:
When we updated to the version of the llvm backend with opaque pointers, all pointer types become the same type in llvm bitcode. As such, we no longer need bitcast instructions to cast from one pointer type to another. These now redundant instructions can be removed.


Robertorosmaninho (approved):
LGTM. is there any other cleaning issues we can do regarding these versions?

PR #1166, opened 2024-11-22
New allocation algorithm

stevenmeker:
Replaced algorithm in class arena with one that uses mmap()'d address spaces and demand paging for each semispace and initializes the current semispace of each arena at start time to avoid and the second semispace of each arena at swap time to avoid relying on Undefined Behavior or having an extra test in kore_arena_alloc().

### DIFF
--- a/include/runtime/arena.h
+++ b/include/runtime/arena.h
@@ -61,7 +61,7 @@ public:
   // by the blocks of that arena. This difference will include blocks containing
   // sentinel bytes. Undefined behavior will result if the pointers belong to
   // different arenas.
-  static ssize_t ptr_diff(char *ptr1, char *ptr2);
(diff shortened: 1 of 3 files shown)


dwightguth (approved):

theo25 (approved):
This looks good to me. 

Just to make sure I understand correctly, could you confirm if the following is accurate: In this new allocation algorithm, we have only one memory block header at the start of the hyperblock, as opposed to the previous algorithm that had one header per smaller block. This is the reason for the change in the calculation for arena size. Also, we still compute the equivalent of the number of small blocks in the new algorithm to inform GC triggering decisions.

stevenmeker:
That is correct. I tried to simulate the old GC triggering algorithm, however note that allocations > 1 MB are now accommodated in the GC'd arena.

PR #1168, opened 2024-11-26
do not garbage collect empty map/list/set

dwightguth:
We move the empty map/list/set allocations to be allocated using `new` and never deallocated or relocated in order to assist in the thread safety of the llvm backend. When we make each allocation arena thread-local, this change will be required, otherwise garbage collection in one thread will corrupt collections in other threads.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -86,8 +86,8 @@ jobs:
         include:
           - runner: [self-hosted, linux, normal]
             os: ubuntu-24.04
-          - runner: MacM1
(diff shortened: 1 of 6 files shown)


theo25 (approved):
Looks good to me.

Robertorosmaninho (approved):
LGTM. The deallocation will happen in each individual thread in the future?

dwightguth:
@Robertorosmaninho the idea is that because there can be only one of this term across all threads, we allocate it somewhere where it never gets deallocated.

PR #1170, opened 2024-11-27
Minor optimization to search

dwightguth:
In the final version of the ULM, we want to enforce that semantics are deterministic. We do this using the search feature and the `--execute-to-branch` option which treats all states with more than one successor as final states and does not rewrite them further.

I tested this by running the ethereum test suite with this option and all the tests passed (with a few trivial changes to the semantics). A minor change resulted from this effort: if only a single state is active at a given time, we do not need to do the expensive equality check performed by `erase`. It is sufficient to simply clear the hash set.

### DIFF
--- a/runtime/util/search.cpp
+++ b/runtime/util/search.cpp
@@ -72,7 +72,11 @@ std::unordered_set<block *, hash_block, k_eq> take_search_steps(
   while (!states.empty() && depth != 0) {
     state = states.front();
     states.pop_front();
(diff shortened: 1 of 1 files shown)


Robertorosmaninho (approved):
LGTM.
Will you push these semantics modifications somewhere?

dwightguth:
@Robertorosmaninho there is a pr in the evm fork you can take a look at if you want.

PR #1173, opened 2024-12-04
Further allocator refactoring

stevenmeker:
Dead code eliminated:
  youngspace_size()
  arena_size()

union memory_block_header and associated functionality replaced. The semispace id is now stored in the last byte of the hyperblock which simplifies address computations.

### DIFF
--- a/include/runtime/alloc.h
+++ b/include/runtime/alloc.h
@@ -15,7 +15,7 @@ extern "C" {
 
 char youngspace_collection_id(void);
 char oldspace_collection_id(void);
-  //size_t youngspace_size(void);
+//size_t youngspace_size(void);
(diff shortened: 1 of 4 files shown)


include/runtime/alloc.h
```diff
@@ -15,7 +15,7 @@ extern "C" {
 
 char youngspace_collection_id(void);
 char oldspace_collection_id(void);
-size_t youngspace_size(void);
+//size_t youngspace_size(void);
```
dwightguth, line 18: let's go ahead and actually delete this; I prefer to avoid commented out code when possible.

theo25 (approved):
Looks good to me.

PR #1175, opened 2024-12-05
New GC strategy.

mariaKt:
This PR adds a new GCStrategy, that treats any pointer with address space other than 0 as a pointer to managed memory.

We extent class GCStrategy in llvm/IR/GCStrategy.h, and register the new GCStrategy.

### DIFF
--- a/bin/llvm-kompile
+++ b/bin/llvm-kompile
@@ -42,6 +42,7 @@ Options:
                                     (immutable) that are enabled by default.
   --hidden-visibility               Set the visibility of all global symbols in generated code to
                                     "hidden"
+  --use-gcstrategy                  Use GC strategy defined for the LLVM backend.
   --profile-matching                Instrument interpeter to emit a profile of time spent in
                                     top-level rule matching on stderr.
   --verify-ir                       Verify result of IR generation.
@@ -195,6 +196,11 @@ while [[ $# -gt 0 ]]; do
       codegen_flags+=("--hidden-visibility")
       codegen_verify_flags+=("--hidden-visibility")
       kompile_clang_flags+=("--hidden-visibility")
+      shift
+      ;;
+    --use-gcstrategy)
+      codegen_flags+=("--use-gcstrategy")
+      kompile_clang_flags+=("--use-gcstrategy")
       shift
       ;;
     --profile-matching)
--- a/bin/llvm-kompile-clang
+++ b/bin/llvm-kompile-clang
@@ -15,6 +15,7 @@ flags=()
 llc_flags=()
 llc_opt_flags="-O0"
 visibility_hidden=false
+use_gcstrategy=false
 link=true
 export verbose=false
 export profile=false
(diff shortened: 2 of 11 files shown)


mariaKt:
I tested that the strategy named `gcs-llvm-backend` exists in the generated plugin as follows:
Using the sample llvm program

```
define void @f() gc "gcs-llvm-backend" {
entry:
  ret void
}
```
I ran

```
llvm-as-16 < sample.ll | llc-16
```
which fails not finding the gc strategy
```
LLVM ERROR: unsupported GC: gcs-llvm-backend
```
and
```
llvm-as-16 < sample.ll | llc-16 -load build/install/lib/kllvm/libKLLVMPass.so
```
which succeeds.

mariaKt:
To test if the gc strategy is correctly loaded as needed, I then added the gc strategy to the function generated by [make_function](https://github.com/runtimeverification/llvm-backend/blob/dfcadbebbf578529644974460a4cf05268a8b63e/lib/codegen/CreateTerm.cpp#L1176), using
```
apply_rule->setGC("gcs-llvm-backend");
```
and edited [llvm-kompile-clang](https://github.com/runtimeverification/llvm-backend/blob/dfcadbebbf578529644974460a4cf05268a8b63e/bin/llvm-kompile-clang#L191) to load the pugin to llc
```
      run @LLC@ \
        "$modopt" -mtriple=@BACKEND_TARGET_TRIPLE@ \
        -load="$passes" \
        -filetype=obj "$llc_opt_flags" "${llc_flags[@]}" -o "$modasm"
```
I used test/defn/imp-sum.kore for testing.

### Testing with --emit-ir

```
llvm-kompile-testing test/defn/imp-sum.kore main --emit-ir -v -o imp-sum-interpreter
```
This worked only when I emit textual IR, not binary IR (I edited [llvm-compile](https://github.com/runtimeverification/llvm-backend/blob/dfcadbebbf578529644974460a4cf05268a8b63e/bin/llvm-kompile#L253) for that). Otherwise, the following error was generated.
```
/usr/lib/llvm-16/bin/llc: error: /usr/lib/llvm-16/bin/llc: tmp.tmZpzEAPip: error: Invalid record (Producer: 'LLVM16.0.6' Reader: 'LLVM 16.0.6')
```
This was the case in develop branch as well if `--emir-ir` is used. I am not sure if there is something wrong with my setup.

### Testing with --emit-ir
```
llvm-kompile-testing test/defn/imp-sum.kore main -v -o imp-sum-interpreter
```
llvm-kompile-codegen fails with
```
LLVM ERROR: unsupported GC: gcs-llvm-backend
```

dwightguth:
Okay so this works when loaded into LLC but it doesn't seem to find it when just running `llvm-kompile-codegen` normally. I believe in the latter case, it is linked via a static library, so it's possible that the TU containing the registration object is not getting linked in because none of its symbols are required. Try making some trivial modification to the main method of llvm-kompile-codegen to ensure that linking in that TU occurs.

dwightguth:
To do that, you can probably just construct an instance of the strategy and just not use it.

mariaKt:
Your suggestion solved the issue, the test I described above worked.

dwightguth (approved):

PR #1176, opened 2024-12-05
Add relevant info for MacOS building process

Robertorosmaninho:


INSTALL.md
```diff
@@ -91,7 +130,8 @@ To run the integration tests, run:
 ```shell
 lit test
 ```
-from the root source directory.
+from the root source directory. You can use `-v` which test is being executed
```
theo25, line 133: you are missing some words here, maybe 
```suggestion
from the root source directory. You can use `-v` to see which test is being executed
```

theo25 (commented):
I also had to install `libffi` with homebrew. Did you not have to do that in your fresh instance of macos?

PR #1178, opened 2024-12-06
use kore_alloc_always_gc instead of malloc when building config

dwightguth:
Making this change prevents a memory leak because `free` was never being called previously on this memory.


Robertorosmaninho (approved):
lgtm

PR #1179, opened 2024-12-09
Create proof trace event for tail call information

theo25:
This PR adds a new event that provides information on the way a function exits control, specifically whether it exits via a tail call or a conventional return statement. This event is added to assist in computing the call stack of the various simplifications from the proof trace hint.


lib/codegen/ProofEvent.cpp
```diff
@@ -372,13 +392,42 @@ std::pair<llvm::BasicBlock *, llvm::BasicBlock *> proof_event::proof_branch(
   return {true_block, merge_block};
 }
 
+std::pair<llvm::BasicBlock *, llvm::BasicBlock *> proof_event::proof_branch(
```
dwightguth, line 395: this feels pretty copy-pasted; is there a way we could separate out a more generic function? maybe using templates?

include/kllvm/codegen/ProofEvent.h
```diff
@@ -228,6 +239,10 @@ class proof_event {
   [[nodiscard]] llvm::BasicBlock *pattern_matching_failure(
       kore_composite_pattern const &pattern, llvm::BasicBlock *current_block);
 
+  [[nodiscard]] llvm::BasicBlock *tail_call_info(
+      std::string const &caller_name, bool is_tail,
+      llvm::Instruction *insert_before, llvm::BasicBlock *current_block);
```
dwightguth, line 244: you don't really need both insert_before and current_block, right? could we change this signature so that you pass one or the other?

dwightguth (commented):
I have a couple low-level comments that I highlighted in the diff, but also one higher level comment: tail call info feels like a weird name for the event when it is emitted even in functions with no tail call. What if we just call it a function exit event?

theo25:
@dwightguth I addressed your comments. Could you take another look?

lib/codegen/CreateTerm.cpp
```diff
@@ -1277,6 +1277,11 @@ bool make_function(
     call->setTailCallKind(llvm::CallInst::TCK_MustTail);
     retval = call;
   } else {
+    bool is_apply_rule = name.substr(0, 11) == "apply_rule_";
```
dwightguth, line 1280: I see that the function signature has a `bool apply`. Is it possible this is already what you require?

dwightguth (approved):
It looks good but please double check what I commented on and fix it if my suspicions are correct.

lib/codegen/CreateTerm.cpp
theo25: I have noticed that but I am not sure what this parameter is supposed to do. I checked all call sites for `make_function` they all pass `true` to this parameter, including one call site from `make_side_condition_function`: https://github.com/runtimeverification/llvm-backend/blob/3b3c19d752797e34fe4a8199bbe76d84f2cf8c41/lib/codegen/CreateTerm.cpp#L1429-L1431

Unless I am missing some call site, it seems that this parameter is dead code.

theo25: OK I was mixing things. You are correct that we can use the `apply` parameter. I have pushed a commit that does. The first of the boolean parameters, i.e. `tailcc`, is the one that is always passed `true`.

PR #1184, opened 2024-12-10
Third attempt at refactoring garbage collector

stevenmeker:
The condition for requesting collections is replaced by requesting a collection
if an allocation passes a tripwire. This tripwire is initially set at MIN_SPACE
and is recalculated after each garbage collation as max(MIN_SPACE, |live data|).

This avoids a potential issue with the old behavior where collections were
trigged by allocating in the last 1 MB of previously allocated semispace, even
if little garbage is being generated, or not collecting frequently enough if
a period of low garbage generation is followed by a period of high garbage
generation, affecting cache performance.

Furthermore, garbage collections are only trigged by allocations in youngspace.
Allocations in oldspace no longer trigger collections since oldspace collections
are handled during a youngspace collection.
Allocations in alwaysgcspace no longer trigger collections since collection
of this space is handled outside of the main garbage collector.

Some class arena member functions have the arena_ component removed as superfluous.

### DIFF
--- a/include/runtime/arena.h
+++ b/include/runtime/arena.h
@@ -30,12 +30,12 @@ public:
 
   // Returns the address of the first byte that belongs in the given arena.
   // Returns nullptr if nothing has been allocated ever in that arena.
-  char *arena_start_ptr() const { return current_addr_ptr; }
+  char *start_ptr() const { return current_addr_ptr; }
 
   // Returns a pointer to a location holding the address of last allocated
   // byte in the given arena plus 1.
   // This address is nullptr if nothing has been allocated ever in that arena.
-  char *arena_end_ptr() { return allocation_ptr; }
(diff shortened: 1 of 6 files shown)


dwightguth:
This code tries to make collections happen more frequently when the live memory shrinks in order to improve cache performance of the garbage collector. It would be helpful before we merge this to see data about gc throughput to determine whether we see any regression that might be caused by the data not bearing out that assumption.

stevenmeker:
Actually I got the formula wrong in the description - the tripwire is set to max(MIN_SPACE, EXPAND_FACTOR * |live data|). I set EXPAND_FACTOR = 2 but this could well be too low - I don't think you'd want to go any lower. An alternative would be to track the largest value of |live data| seen in collections to date and multiply this by EXPAND_FACTOR.

PR #1185, opened 2024-12-11
runtime/proof_trace_writer.h

dmitry-vlasov:
Callbacks in `proof_trace_callback_writer` class are made virtual so that they may be overriden in descendent classes.


theo25 (approved):
Looks good to me.

PR #1189, opened 2024-12-13
runtime/proof_trace_writer.h

dmitry-vlasov:
-  set of inner `proof_trace_callback_writer` structs are made `protected` instead of `private`
 - `kore_configuration_construction` struct stores `block*` instead of a raw pointer `void*`


theo25 (approved):
Looks good to me

PR #1191, opened 2024-12-16
make hyperblock size depend on physical memory

stevenmeker:
hyperblock_size is now a class static variable that is set to the size of physical memory rounded to the next power of 2.


stevenmeker:
It doesn't really fix a bug, but I think it can be regarded as defensive programming.

PR #1193, opened 2024-12-19
WIP: Gc addressspace 0

mariaKt:
This PR adds
- the explicit use/casting to address space 0 in CreateTerm.cpp.
- the infrastructure to define the functions that contain the `addrspacecast` instructions for the various address spaces (a separate file, linked after the regular passes, and the always inline pass executed afterwards).

Current status: The tests work as expected when the conditional compilation flag `use_gcstrategy` that enables the use of the LLVM backend strategy is disabled.

Enabling the flag, e.g. by adding the tests
```
// RUN: %gcs-interpreter
// RUN: %check-grep
```
in defn/imp.kore, results in the test failing (that interpreter fails with segfault).

This is WIP, and is merged in order to keep the code up to date until we revisit this, to
- address the segfault
- replace address space 0 with the actual address space needed in each place
- similarly edit the other files


tools/llvm-kompile-codegen/main.cpp
```diff
@@ -239,12 +242,14 @@ int main(int argc, char **argv) {
...
   }
 
-  do_bitcode_linking(*mod);
+  do_bitcode_linking(*mod, (char *)alloc_cpp_o_ll, alloc_cpp_o_ll_len);
 
   if (!no_optimize) {
     apply_kllvm_opt_passes(*mod, hidden_visibility);
   }
 
+  do_bitcode_linking(*mod, (char *)opaque_ll, opaque_ll_len);
```
dwightguth, line 251: looks like you forgot to call apply_inline_pass here

runtime/opaque/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+add_custom_command(OUTPUT ${CMAKE_BINARY_DIR}/runtime/opaque/opaque_cpp.cpp
```
dwightguth, line 1: this cmake file seems suspect. this is based on a C++ file, but you have an ll file here. I suspect a copy paste error. You should take a closer look.

dwightguth (commented):
I haven't fully reviewed this yet, but I have a couple comments. For the purposes of clarity, can you share the .ll file of IMP that this PR generates?

dwightguth (commented):
it's possible the segfault is due to doing an addrspace cast from an address space to itself, which the LLVM documentation tells you not to do. I'll see if I can spot any other possible causes when you share the llvm bitcode I asked you for.

tools/llvm-kompile-codegen/main.cpp
mariaKt: Right, thank you. Addressed.

mariaKt:
Here is the ll for IMP.
[imp.ll.txt](https://github.com/user-attachments/files/18213722/imp.ll.txt)

dwightguth:
It would help if you could do the following:

1. Determine whether or not the bug still occurs in the version of IMP that does not have the pattern matching optimization enabled.
2. Share the version of the code generated with --no-optimize.

dwightguth:
Something feels weird about the code, but I don't immediately know what it is and I could be seeing something that is actually normal.

There's a third thing you should try:

* compare the generated code before and after making this change and see if anything seems obviously suspect in the difference between the two on the same function.

mariaKt:
1. imp.kore also failed when disabling pattern matching optimization.
2. I attach the code generated with --nooptimize (pattern matching optimization enabled again).
[imp.gc.patmatchopt.noopt.ll.txt](https://github.com/user-attachments/files/18216024/imp.gc.patmatchopt.noopt.ll.txt)
3. I will compare the ll files for IMP from develop and this branch. I attach them as well.
[imp.develop.patmatchopt.opt.ll.txt](https://github.com/user-attachments/files/18216032/imp.develop.patmatchopt.opt.ll.txt)
[imp.gc.patmatchopt.opt.ll.txt](https://github.com/user-attachments/files/18216033/imp.gc.patmatchopt.opt.ll.txt)

dwightguth (approved):

PR #1194, opened 2025-01-09
Callbacks: Fix side-condition events when arity is non-null

alexoltean61:
This PR fixes a bug in calls to calls to `side_condition_event_callback`. Every time such a call should occur, the `rewrite_event_callback` is invoked instead.

This is due to a bug in the method that builds substitutions, and calls a callback once the substitution has the required size. Instead of conditionally calling either the rewrite callback or the sidecondition callback, it always calls the rewrite one.

This PR fixes the bug.

### DIFF
--- a/include/runtime/proof_trace_writer.h
+++ b/include/runtime/proof_trace_writer.h
@@ -285,6 +285,8 @@ private:
   std::optional<rewrite_event_construction> current_rewrite_event_{
       std::nullopt};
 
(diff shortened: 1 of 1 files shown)


theo25:
This looks as a legitimate bug fix to me. However I expect tests to fail with this fix because we would need to update their reference output. I will start the ci and let’s see what we get.

alexoltean61:
Thank you!

PR #1199, opened 2025-02-18
Made the size of the arena semispaces configurable with cmake

mariaKt:
This PR makes the computation of the arena semispaces to depend on a variable configurable with cmake.


Robertorosmaninho (approved):
LGTM

theo25 (approved):

PR #1200, opened 2025-03-04
add cases for MInt{8}, MInt{16}, and MInt{128}

dwightguth:
This is intended to fix crashes when the llvm backend manipulates machine integers of sizes 8, 16, and 128.

### DIFF
--- a/runtime/collections/hash.cpp
+++ b/runtime/collections/hash.cpp
@@ -19,6 +19,12 @@ __attribute__((always_inline)) void add_hash8(void *h, uint8_t data) {
   hash_length++;
 }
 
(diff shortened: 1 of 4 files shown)


runtime/collections/hash.cpp
Robertorosmaninho, line 1: Should we add
- add_hash128
- add_hash160
- add_hash256
here?

Robertorosmaninho (approved):
LGTM. But I'll like to wait @virgil-serbanuta tests.

virgil-serbanuta:
LGTM.

Thank you! I uncommented all tests that crashed because of the MInt issue and they seem to work on the mint3 branch.

PR #1203, opened 2025-04-10
Update `free_all_kore_mem` to avoid memory corruption

Robertorosmaninho:
Ideally, when we call this function, we want to clean everything, including all semispaces to free the memory for future usage. However, we found a bug com a few `ValidBlocks` tests from Ethereum BlockchainTests that throws an error when collecting the old generation when we had called this `free_all_kore_mem` before. This issue doesn't happen for the GeneralStateTests, that's why we didn't find it before.

This modification is safe and still keeps the benefits of freeing a lot of space, as the majority of allocated memory observed in the tests are stored in the young generation. 

We'll go back to investigate this GC bug soon.

### DIFF
--- a/runtime/collect/collect.cpp
+++ b/runtime/collect/collect.cpp
@@ -347,7 +347,6 @@ void free_all_kore_mem() {
   kore_collect(nullptr, 0, nullptr, true);
   kore_clear();
   youngspace.munmap_arena_and_reset();
-  oldspace.munmap_arena_and_reset();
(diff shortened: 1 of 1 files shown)


dwightguth (commented):
As I've said elsewhere, reth does not need to unmap the memory like this, so we should probably remove this entirely and put it into the geth codebase. This wasn't urgent, but if we're making a pr to touch this anyway, we might as well do it

dwightguth (approved):

PR #1206, opened 2025-05-05
Make thread_local variables constinit

dwightguth:
By declaring all thread_local variables that can easily be given static initialization constinit, and removing the destructor of `arena`, we can remove the call to `__tls_init` which is inserted into every single allocation, significantly speeding up the code. One consequence of this is that threads are responsible for calling munmap explicitly before exiting now; my understanding is they are pretty much always doing this currently via `free_all_kore_mem`, so this shouldn't break anything, but it needs testing to make sure.


Robertorosmaninho:
I believe you can undo the last commit. The issue seems to be on the `CXX_STANDARD: 11` version. We must pass it to 20. We'll have the same issue with the blockchain-k-plugin which uses C++ 17.

Robertorosmaninho (approved):
I could test it with our GEth and KEVM tests and everything seemed to work fine. From previous tests, I could also check the intended speed up.

PR #1209, opened 2025-06-28
Adding support for Machine Integer version of Bytes, List, and MInt functions

Robertorosmaninho:


runtime/collections/lists.cpp
```diff
@@ -111,6 +120,10 @@ size_t hook_LIST_size_long(SortList list) {
   return list->size();
 }
 
+uint64_t hook_LIST_size64(SortList list) {
```
dwightguth, line 123: hook_list_size_long already does this. Check for other similar duplicates that may exist.

dwightguth (commented):
I'd suggest considering the following two factors regarding the hooks added:

1. Is there a particularly compelling reason why both arguments of hooks with more than one mint arg need to have the same size?
2. If a mint arg or return value carresponds to a size in bytes or in elements of something in memory, you should be able to use a non parametric mint 64 for it.

None of this is absolutely necessary but if we're adding hooks to the frontend we ought to think about this.

config/llvm_header.inc
```diff
@@ -245,6 +245,15 @@ define void @take_search_step(ptr %subject) {
   ret void
 }
 
+define i256 @hook_BYTES_length256(ptr %b) {
+entry:
+  %hdr = load i64, ptr %b
+  %len = and i64 %hdr, 1099511627775 ; @LENGTH_MASK@
```
theo25, line 361: This is not ideal. Since LENGTH_MASK is a configurable value, having it hardcoded here can potentially create confusion. Maybe you can write a C++ function that is just wrapper for the `len` macro and use it here.

There are more instances of the value being hard coded. We will need to apply this to all of them.

mariaKt (changes requested):
Thank you @Robertorosmaninho. I added only two minor comments.

config/llvm_header.inc
```diff
@@ -81,6 +81,36 @@ declare void @print_configuration(ptr, ptr)
 ; finish_rewriting.ll
 
 declare i64 @__gmpz_get_ui(ptr)
+declare ptr @kore_alloc(i64)
+declare void @memcpy(ptr, ptr, i64)
+declare void @interger_overflow(i64)
```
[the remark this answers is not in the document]
Robertorosmaninho: Done

theo25 (approved):

PR #1213, opened 2025-08-28
bytes2hexstring hook

RaoulSchaffranek:
This PR adds a new hook implementation for converting byte arrays into hex strings.
This is needed by the kontrol-node to encode memory as JSON strings efficiently.

### DIFF
--- a/bin/llvm-krun
+++ b/bin/llvm-krun
@@ -9,7 +9,8 @@ expanded_input_file="$(mktemp tmp.in.XXXXXXXXXX)"
 parser_file="$(mktemp tmp.parse.XXXXXXXXXX)"
 temp_inputs=()
 
-cleanup () { # shellcheck disable=SC2329
+# shellcheck disable=SC2329
+cleanup () {
   # shellcheck disable=SC2317
   rm -rf "$input_file" "$expanded_input_file" "$output_file" "$parser_file" "${temp_inputs[@]}"
 }
--- a/bin/llvm-krun
+++ b/bin/llvm-krun
@@ -9,7 +9,7 @@ expanded_input_file="$(mktemp tmp.in.XXXXXXXXXX)"
 parser_file="$(mktemp tmp.parse.XXXXXXXXXX)"
 temp_inputs=()
 
-cleanup () {
+cleanup () { # shellcheck disable=SC2329
   # shellcheck disable=SC2317
   rm -rf "$input_file" "$expanded_input_file" "$output_file" "$parser_file" "${temp_inputs[@]}"
 }
--- a/runtime/strings/bytes.cpp
+++ b/runtime/strings/bytes.cpp
@@ -148,37 +148,34 @@ hook_BYTES_int2bytes(SortInt len, SortInt i, SortEndianness endianness_ptr) {
   return result;
 }
 
-string *bytes2string(string *b, size_t len) {
-  auto *result = static_cast<string *>(kore_alloc_token(sizeof(string) + len));
-  memcpy(result->data, b->data, len);
-  init_with_len(result, len);
-  return result;
-}
-
 SortString hook_BYTES_bytes2string(SortBytes b) {
-  return bytes2string(b, len(b));
+  auto len_b = len(b);
+  auto *result = static_cast<string *>(kore_alloc_token(sizeof(string) + len_b));
+  memcpy(result->data, b->data, len_b);
+  init_with_len(result, len_b);
+  return result;
 }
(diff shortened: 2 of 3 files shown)


runtime/strings/bytes.cpp
```diff
@@ -162,6 +163,24 @@ SortBytes hook_BYTES_string2bytes(SortString s) {
   return hook_BYTES_bytes2string(s);
 }
 
+string *bytes2hexstring(string *b, size_t len) {
+  static const std::array<char, 16> hexchars
+      = {'0', '1', '2', '3', '4', '5', '6', '7',
+         '8', '9', 'a', 'b', 'c', 'd', 'e', 'f'};
+  auto *result
+      = static_cast<string *>(kore_alloc_token(sizeof(string) + len * 2));
+  for (size_t i = 0; i < len; i++) {
```
tothtamas28: I see that this pattern is repeated in `bytes.cpp`, but if `len > len(b)`, this loop will produce an out of bounds error. So consider either checking this as a precondition, or obtaining the length in the function body.

runtime/strings/bytes.cpp
```diff
@@ -162,6 +163,24 @@ SortBytes hook_BYTES_string2bytes(SortString s) {
...
+    result->data[i * 2] = hexchars.at((b->data[i] >> 4) & 0xf);
+    result->data[i * 2 + 1] = hexchars.at(b->data[i] & 0xf);
+  }
+  init_with_len(result, len * 2);
+  return result;
+}
+
+SortString hook_BYTES_bytes2hexstring(SortBytes b) {
```
tothtamas28: Consider `bytes2hex` for a snappier name.

Also, consider adding a doc comment.

unittests/runtime-strings/bytestest.cpp
```diff
@@ -171,6 +172,19 @@ BOOST_AUTO_TEST_CASE(bytes2string) {
   BOOST_CHECK_EQUAL(0, memcmp(_1234->data, "1234", 4));
 }
 
+BOOST_AUTO_TEST_CASE(bytes2hexstring) {
+  auto empty = make_string("");
+  auto res = hook_BYTES_bytes2hexstring(empty);
+  BOOST_CHECK(res != empty);
+  BOOST_CHECK_EQUAL(len(res), 0);
+
+  auto _00ff = make_string("\x00\xff", 2);
```
tothtamas28: This is not a great test input, as it will mask the error where characters for a given byte are in the wrong order in the output. Maybe `"\x01\xef"`?

tothtamas28:
@jberthold, is there a spreadsheet or similar for documenting hook implementations?

unittests/runtime-strings/bytestest.cpp
tothtamas28: Also, consider adding a test case for other boundary values, in particular the empty string.

RaoulSchaffranek: Good catch! I improved the test case. The empty string is already covered.

runtime/strings/bytes.cpp
```diff
@@ -1,4 +1,5 @@
 #include <algorithm>
+#include <array> // Add this include for std::array
```
tothtamas28:
```suggestion
#include <array>
```

tothtamas28 (commented):
LGTM. I'll leave the final approval to @Robertorosmaninho.

bin/llvm-krun
```diff
@@ -9,6 +9,7 @@ expanded_input_file="$(mktemp tmp.in.XXXXXXXXXX)"
 parser_file="$(mktemp tmp.parse.XXXXXXXXXX)"
 temp_inputs=()
 
+# shellcheck disable=SC2329
```
Robertorosmaninho, line 12: Sorry, why is this needed here?

Robertorosmaninho (approved):
LGTM, just one question:

bin/llvm-krun
RaoulSchaffranek: The shell check CI workflow kept failing, see for example this action: https://github.com/runtimeverification/llvm-backend/actions/runs/17319617915/job/49169763898

If I understand correctly, it's a false positive because the function is used in the following line. So I marked it to skip the check.

It's not related to this PR, I think, but maybe some CI flakiness caused by a shell check update.

Robertorosmaninho: Got it! Thanks for clarifying it!

PR #1219, opened 2026-06-18
Allocate block-exceeding tokens directly in the old generation

Robertorosmaninho:
While running large MegaETH workloads under the LLVM backend we hit a crash in `kore_collect` / `arena::evacuate` while migrating a block with a corrupted header. The culprit is a `String` token larger than a block: `init_with_len` correctly flags it `NOT_YOUNG`, but `kore_alloc_token` was still allocating it in the young semispace. The collector never migrates a `NOT_YOUNG` object out of the young from-space, which is reclaimed on every collection, so the token dangled — a later allocation overwrote its bytes and the next collection walked a garbage header. It reproduces independently of the hyperblock size.

The fix routes tokens larger than `BLOCK_SIZE` to `oldspace` (matching what `hook_BUFFER_concat_raw` already does for large buffers) and sets `AGE_MASK` on the `NOT_YOUNG` header so `is_in_old_gen_hdr()` recognises the block as a live old-generation object. The oversized branch is kept in a cold, out-of-line helper (`__builtin_expect`) so the common small-token path of the `always_inline` `kore_alloc_token` isn't bloated at every call site.

I also added `test-gc-large-token`: it builds a single > 1 MiB token, keeps it live across forced collections that reuse the abandoned young region, then reads it back. The test crashes on the buggy version (the collector walks the dangling token) and passes with the fix; the rest of the GC and unit suites are unaffected.

This is safe for the hot path: small-token allocation is unchanged apart from one predicted-not-taken branch, and the oversized path is rare. Instruction counts on an allocation-heavy benchmark are flat versus before, and the full `lit` suite stays green.


runtime/lto/alloc.cpp
```diff
@@ -60,9 +60,28 @@ __attribute__((always_inline)) void *kore_alloc(size_t requested) {
...
+  }
+  // A token whose total size exceeds a block is flagged NOT_YOUNG by init_with_len (it cannot
+  // participate in the young semispace's block model). Such tokens must be allocated directly in
+  // the old generation: the young from-space is reclaimed on every collection, and the collector
+  // never migrates a NOT_YOUNG object out of it, so a large token left in the young space would
+  // dangle. This matches the existing large-buffer handling in hook_BUFFER_concat_raw.
+  if (__builtin_expect(size > BLOCK_SIZE, 0)) {
+    return kore_alloc_token_oldspace(size);
+  }
+  return youngspace.kore_arena_alloc(size);
```
gtrepta: Have you done any performance experiments around this? I'm curious if flipping the return statements around, or maybe declaring something like this:
```cpp
arena& a = __builtin_expect(size > BLOCK_SIZE, 0) ? oldspace : youngspace;
return a.kore_arena_alloc(size);
```
could have an appreciable difference. I think stylistically it looks nicer, but performance should take priority.

include/runtime/types.h
```diff
@@ -96,7 +96,15 @@ __attribute__((always_inline)) constexpr uint64_t len(T const *s) {
 
 template <typename T>
 __attribute__((always_inline)) constexpr void init_with_len(T *s, uint64_t l) {
-  s->h.hdr = l | (l > BLOCK_SIZE - sizeof(char *) ? NOT_YOUNG_OBJECT_BIT : 0);
+  // A token larger than a block cannot live in the young semispace's per-block model, so it is
+  // marked NOT_YOUNG and (see kore_alloc_token) allocated directly in the old generation. It
+  // must also carry the AGE bit so that is_in_old_gen_hdr() recognises it as a live heap block;
+  // otherwise the collector treats it as a non-heap/static block, never migrates it, and the
+  // young from-space it was (incorrectly) placed in gets reclaimed, dangling the reference.
```
gtrepta: This looks like comment slop. Can it just be a one-liner, like "Allocations bigger than BLOCK_SIZE go to oldspace"? That seems to be the original intent anyway, and no comment existed before.

runtime/lto/alloc.cpp
```diff
@@ -60,9 +60,28 @@ __attribute__((always_inline)) void *kore_alloc(size_t requested) {
...
   size_t size = (requested + 7) & ~7;
-  return youngspace.kore_arena_alloc(size < 16 ? 16 : size);
+  if (size < 16) {
+    size = 16;
+  }
+  // A token whose total size exceeds a block is flagged NOT_YOUNG by init_with_len (it cannot
+  // participate in the young semispace's block model). Such tokens must be allocated directly in
+  // the old generation: the young from-space is reclaimed on every collection, and the collector
+  // never migrates a NOT_YOUNG object out of it, so a large token left in the young space would
+  // dangle. This matches the existing large-buffer handling in hook_BUFFER_concat_raw.
```
gtrepta: More comment slop. Can be a one-liner.

gtrepta (approved):
LGTM

runtime/lto/alloc.cpp
Robertorosmaninho: Hey guy, here is the result of a small experiment with 8M small-token allocs:
  | median wall | cycles | IPC
-- | -- | -- | --
master | 1.010s | 5.655B | 3.72
current approach | 1.030s | 5.840B | 3.59
ternary (your suggestion) | 1.030s | 5.736B | 3.66

So the performance is basically the same, but with the ternary approach, we achieve a higher IPC and require fewer cycles.

Robertorosmaninho: So, I think it's worth using your idea here, let me push a new commit with it! Thanks for the careful review and suggestion!

runtime/lto/alloc.cpp
Robertorosmaninho: Solved.

include/runtime/types.h
Robertorosmaninho: Solved with a small one-line comment.

gtrepta (approved):

